PR #127, opened 2019-02-26
Switching tests to use Pose in Win32Tests

karanbirsingh:
This PR switches the tests in Win32Tests.csproj to use Pose instead of using Fakes. The project should no longer have any references to fakes.

I looked into creating a helper method to reduce redundancy but the shim did not seem to apply when the method under test (RS3 vs RS5) was passed in as a parameter.


karanbirsingh:
@neocln are you alright removing the 'do not merge' label?

PR #158, opened 2019-03-06
Add UIAutomation as a prefix in front of Name property to emphasize t…

neocln:
#### PR checklist

- [x] Does this address an existing issue? If yes, Issue# - 1469626
- [x] Includes UI changes?

#### Describe the change
In How-to-fix resources of Rules project, Add UIAutomation as a prefix in front of Name property to emphasize that the Name property is for UIAutomation, not Name of XAML.

### DIFF
--- a/src/AccessibilityInsights.Rules/Resources/HowToFix.Designer.cs
+++ b/src/AccessibilityInsights.Rules/Resources/HowToFix.Designer.cs
@@ -822,7 +822,7 @@ namespace AccessibilityInsights.Rules.Resources {
         }
         
         /// <summary>
-        ///   Looks up a localized string similar to The Name property for the given element is empty, but the element isn&apos;t focusable. Please consider whether or not the element should have a name..
+        ///   Looks up a localized string similar to The UIAutomation Name property for the given element is empty, but the element isn&apos;t focusable. Please consider whether or not the element should have a name..
         /// </summary>
         internal static string NameEmptyButElementNotKeyboardFocusable {
             get {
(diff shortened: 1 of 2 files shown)


src/AccessibilityInsights.Rules/Resources/HowToFix.Designer.cs
```diff
@@ -822,7 +822,7 @@ internal class HowToFix {
         }
         
         /// <summary>
-        ///   Looks up a localized string similar to The Name property for the given element is empty, but the element isn&apos;t focusable. Please consider whether or not the element should have a name..
+        ///   Looks up a localized string similar to The UIAutomation Name property for the given element is empty, but the element isn&apos;t focusable. Please consider whether or not the element should have a name..
```
babylonbee: Maybe it is minor, but could it be better if we use lower case for some words in the middle, to be more specific:
"...similar to The UIAutomation Name property..." => "...similar to the UIAutomation name property..."  Same comments for other places

src/AccessibilityInsights.Rules/Resources/HowToFix.Designer.cs
```diff
@@ -831,7 +831,7 @@ internal class HowToFix {
         }
         
         /// <summary>
-        ///   Looks up a localized string similar to Provide a Name property for the element that:
+        ///   Looks up a localized string similar to Provide a UI Automation Name property for the element that:
```
babylonbee: "Provide" => "provide"; "Name property" => "name property" ? Same comments for other places

PR #180, opened 2019-03-18
Add a job that builds without fakes

karanbirsingh:
This adds a job that simply builds the solution with FAKES_SUPPORTED disabled. It is a sanity check to ensure nothing is broken for users without fakes. The tests themselves run in existing jobs.

### DIFF
--- a/build/prbuild.yml
+++ b/build/prbuild.yml
@@ -99,7 +99,7 @@ jobs:
       configuration: debug
       rerunFailedTests: true
 
-- job: Build without fakes
+- job: build-without-fakes
   pool:
     vmImage: 'vs2017-win2016'
   steps:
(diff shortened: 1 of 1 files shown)


build/prbuild.yml
```diff
@@ -95,3 +98,20 @@ jobs:
...
+
+  - task: NuGetCommand@2
+    displayName: 'NuGet restore'
+
+  - task: VSBuild@1
+    displayName: 'Build Solution **\*.sln'
+    inputs:
+      vsVersion: 15.0
+      platform: '$(BuildPlatform)'
+      configuration: debug
```
DaveTryon: >debug [](start = 21, length = 5)

What would happen if both the debug and the build_without_fakes jobs were to run on the same agent instance? Is there a guarantee that the build folders will be isolated between the agents?

karanbirsingh: I believe so. If you go to any of the build pipelines, click on 'edit pipeline' (in the dropdown with '...' as its label on the top right), and then on 'get sources', you can see 'clean' is set to true.

PR #182, opened 2019-03-18
Excluding Fakes test methods based on a preprocessor symbol …

RobGallo:
#### PR checklist

- [ ] Includes UI changes?

#### Describe the change
This allows us to exclude only test methods which use Fakes, rather than all methods in a file with a method which uses Fakes. And that makes it possible for developers who use versions of Visual Studio which don't have Fakes support (like Community) to run more unit tests.

### DIFF
--- a/src/AccessibilityInsights.AutomationTests/AutomationSessionUnitTests.cs
+++ b/src/AccessibilityInsights.AutomationTests/AutomationSessionUnitTests.cs
@@ -1,11 +1,13 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using AccessibilityInsights.Automation;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
 using System;
 using System.Collections.Generic;
+#if FAKES_SUPPORTED
 using AccessibilityInsights.Actions.Fakes;
-using AccessibilityInsights.Automation;
 using Microsoft.QualityTools.Testing.Fakes;
-using Microsoft.VisualStudio.TestTools.UnitTesting;
+#endif
 
 namespace AccessibilityInsights.AutomationTests
 {
@@ -19,6 +21,7 @@ namespace AccessibilityInsights.AutomationTests
 
         private readonly CommandParameters TestParameters = new CommandParameters(new Dictionary<string, string>(), string.Empty);
 
+#if FAKES_SUPPORTED
         /// <summary>
         /// Set up all of the shims needed for the unit tests--we make no attempt to
         /// identify a subset of shims to configure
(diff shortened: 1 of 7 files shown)


DaveTryon:
>         private static A11yElement CreateA11yElement(List<A11yElement> children = null, List<ScanStatus> statuses = null)

This is only called by code that needs FAKES_SUPPORTED. Should it be in the #if?

---
Refers to: src/AccessibilityInsights.AutomationTests/SnapshotCommandUnitTests.cs:69 in f49f86e. [](commit_id = f49f86ecca9396d8e60e9fb1079dd377cf7d6b93, deletion_comment = False)

DaveTryon:
>         const string TestMessage = "I find his reaction most illogical";

Lines 23-96 are currently outside the #if FAKES_SUPPORTED block, but they're only used by code within the block. I'd suggest moving the #if block to immediately after line 22

---
Refers to: src/AccessibilityInsights.AutomationTests/SnapshotCommandUnitTests.cs:23 in f49f86e. [](commit_id = f49f86ecca9396d8e60e9fb1079dd377cf7d6b93, deletion_comment = False)

src/AccessibilityInsights.AutomationTests/AutomationSessionUnitTests.cs
```diff
@@ -57,22 +62,19 @@ public void NewInstance_NoInstanceExists_ExceptionInCtor_RethrowsSameException()
                 }
             }
         }
+#endif
 
         [TestMethod]
         [Timeout (2000)]
         public void NewInstance_ClearInstance_NotPowerShell_AssemblyResolverIsNull_DoesNotThrow()
```
DaveTryon: >NewInstance_ClearInstance_NotPowerShell_AssemblyResolverIsNull_DoesNotThrow [](start = 20, length = 75)

I want to double-check the internals to confirm that these shims aren't needed internally to exercise the proper code.

src/AccessibilityInsights.AutomationTests/AutomationSessionUnitTests.cs
```diff
@@ -57,22 +62,19 @@ public void NewInstance_NoInstanceExists_ExceptionInCtor_RethrowsSameException()
...
         {
-            using (ShimsContext.Create())
+            try
             {
-                InitializeShims();
```
DaveTryon: >InitializeShims [](start = 16, length = 15)

These should have been removed with commit 523568c0919c30fd2e19d6b3b83804be9ad88a0e. Thanks for removing them here

src/AccessibilityInsights.AutomationTests/CommandParametersUnitTests.cs
```diff
@@ -74,6 +76,7 @@ public void Ctor_ConfigFileDoesNotExist_ThrowsAutomationException_ErrorAutomatio
             }
         }
 
+        #if FAKES_SUPPORTED
```
karanbirsingh: It looks in general, the #if indentation is aligned with its method whereas the #endif line has no indentation. It might be nicer for both to be consistent

RobGallo: The indent is now uniform, although I can't really tell if it looks great.

PR #191, opened 2019-03-19
Refactor azure dev ops extension to implement new issue interface

jalkire:
#### PR checklist

- [ ] Includes UI changes?

#### Describe the change
- Refactored AzureDevOps extension to line up with issue filing interface changes
  - Moved most of issue filing logic and ux out of SharedUx into ADO extension
  - Ported over applicable existing unit tests

Todo:
- Fix FabricIcons and other ux components to which this extension has lost access


src/AccessibilityInsights.Extensions.AzureDevOps/AzureBoardsIssueReporting.cs
```diff
@@ -0,0 +1,81 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using AccessibilityInsights.Extensions.Interfaces.IssueReporting;
+using System;
+using System.Collections.Generic;
+using System.ComponentModel.Composition;
+using System.Threading.Tasks;
+using AccessibilityInsights.Extensions.AzureDevOps.Models;
```
RobGallo: Alpha sort

src/AccessibilityInsights.Extensions.AzureDevOps/AzureBoardsIssueReporting.cs
```diff
@@ -0,0 +1,81 @@
...
+using AccessibilityInsights.Extensions.AzureDevOps.FileIssue;
+using System.Windows;
+
+namespace AccessibilityInsights.Extensions.AzureDevOps
+{
+    [Export(typeof(IIssueReporting))]
+    public class AzureBoardsIssueReporting : IIssueReporting
+    {
+        private AzureDevOpsIntegration AzureDevOps = AzureDevOpsIntegration.GetCurrentInstance();
```
RobGallo: Are these fields meant to be readonly?

src/AccessibilityInsights.Extensions.AzureDevOps/AzureBoardsIssueReporting.cs
```diff
@@ -0,0 +1,81 @@
...
+
+        private ExtensionConfiguration Configuration => AzureDevOps.Configuration;
+
+        public bool IsConnected => AzureDevOps.ConnectedToAzureDevOps;
```
RobGallo: If this is inherited, can we make it a get-only property?

src/AccessibilityInsights.Extensions.AzureDevOps/AzureBoardsIssueReporting.cs
RobGallo: > Is this field meant to be readonly?

src/AccessibilityInsights.Extensions.AzureDevOps/AzureBoardsIssueReporting.cs
```diff
@@ -0,0 +1,81 @@
...
+                Action<int> updateZoom = (int x) => Configuration.ZoomLevel = x;
+                (int? issueId, string newIssueId) = FileIssueHelpers.FileNewIssue(issueInfo, Configuration.SavedConnection,
+                    Application.Current.MainWindow.Topmost, Configuration.ZoomLevel, updateZoom);
+
+                // Check whether issue was filed once dialog closed & process accordingly
+                if (issueId.HasValue)
+                {
+                    try
+                    {
+                        var success = FileIssueHelpers.AttachIssueData(issueInfo, newIssueId, issueId.Value).Result;
```
RobGallo: Consider removing the success variable and placing the function call inside the if statement.

src/AccessibilityInsights.Extensions.AzureDevOps/AzureBoardsIssueReporting.cs
```diff
@@ -0,0 +1,81 @@
...
+        public Task<IIssueResult> FileIssueAsync(IssueInformation issueInfo)
+        {
+            return Task.Run<IIssueResult>(() => {
+
+                Action<int> updateZoom = (int x) => Configuration.ZoomLevel = x;
+                (int? issueId, string newIssueId) = FileIssueHelpers.FileNewIssue(issueInfo, Configuration.SavedConnection,
+                    Application.Current.MainWindow.Topmost, Configuration.ZoomLevel, updateZoom);
+
+                // Check whether issue was filed once dialog closed & process accordingly
+                if (issueId.HasValue)
```
RobGallo: You can eliminate a level of nesting by checking for the negative and returning null.

src/AccessibilityInsights.Extensions.AzureDevOps/AzureDevOpsIntegration.cs
```diff
@@ -432,26 +435,47 @@ public Uri GetTeamProjectUrl(string projectName, string teamName)
             return builder.Uri;
         }
 
+
+        public async Task HandleLoginAsync(bool showDialog=false, Uri serverUri = null)
```
RobGallo: As a rule, it's not a good idea to pass bools into a function unless you are setting a bool value. The problem is that when looking at the function call, the info about what the bool represents may not be present. Even if you rely on intellisense, the comments could easily get out-of-date.

src/AccessibilityInsights.Extensions.AzureDevOps/ConfigurationControl.xaml
```diff
@@ -0,0 +1,176 @@
+﻿<!-- Copyright (c) Microsoft. All rights reserved.
```
RobGallo: How can I view this control in the UI to check its accessibility?

jalkire: This is just the production azure dev ops connection control moved over to the extension, so its accessibility is hopefully unchanged. That being said, you should be able to find it in the connections tab of settings if you pull this branch. There were some other changes made to the tab which are not a part of this PR.

src/AccessibilityInsights.Extensions.AzureDevOps/AzureDevOpsIntegration.cs
```diff
@@ -48,6 +49,8 @@ public partial class AzureDevOpsIntegration
         /// </summary>
         public bool ConnectedToAzureDevOps => _baseServerConnection?.HasAuthenticated == true;
 
+        public ExtensionConfiguration Configuration { get; private set; } = new ExtensionConfiguration();
```
DaveTryon: >private set; [](start = 59, length = 12)

The private setter is only required if this member gets changed after construction. From what I see in the code, that's not happening, so I think the private setter can go away.

src/AccessibilityInsights.Extensions.AzureDevOps/AzureDevOpsIntegration.cs
```diff
@@ -163,7 +166,7 @@ public void Disconnect()
         /// with the currently connected account
         /// </summary>
         /// <returns>Team projects associated with current account or null if disconnected</returns>
-        public IEnumerable<IProject> GetTeamProjects()
```
DaveTryon: >IProject [](start = 27, length = 8)

Is this still the public interface that gets called by the app? If so, then I'd argue that we should be using the generic types defined in the app-visible interface

src/AccessibilityInsights.Extensions.AzureDevOps/AzureDevOpsIntegration.cs
```diff
@@ -163,7 +166,7 @@ public void Disconnect()
...
+        public IEnumerable<Models.TeamProject> GetTeamProjects()
```
DaveTryon: >public [](start = 8, length = 6)

Now that these methods are no longer called by the app, can we also tighten the visibility of these methods (or maybe the whole class) from public to internal?

src/MSI/Product.wxs
```diff
@@ -142,12 +142,12 @@ Licensed under the MIT license. See LICENSE file in the project root for full li
 
                  <Directory Id="BugTemplates" Name="BugTemplates">
```
DaveTryon: >BugTemplates [](start = 32, length = 12)

Should the install folder also be renamed? It seems like the component name of BugTemplateComp is also a candidate for renaming here...

src/AccessibilityInsights.Extensions.AzureDevOpsTests/FileIssue/FileIssueHelpersTests.cs
```diff
@@ -0,0 +1,69 @@
...
+                var output = FileIssueHelpers.FileNewIssue(issueInfo,
+                    connInfo, false, 0, (_) => { });
+
+                Assert.IsNull(output.issueId);
+                Assert.IsNotNull(output.newIssueId);
+                Assert.IsTrue(string.IsNullOrEmpty(output.newIssueId));
+            }
+        }
+
+
```
DaveTryon: Can we remove one of the extra newlines between these 2 tests?

DaveTryon:
> ��<Fakes xmlns="http://schemas.microsoft.com/fakes/2011/">

We started with assembly-specific fakes, but ran into trouble--that was why the Fakes.Prebuild project was created. To prevent potential problems like we had before, I'd lobby that it would safer and more consistent to include this as part of the Fakes.Prebuild project.

---
Refers to: src/AccessibilityInsights.Extensions.AzureDevOpsTests/Fakes/AccessibilityInsights.Core.fakes:1 in 5ae390a. [](commit_id = 5ae390a22184ed392ca282d3061fb89d61fdaf22, deletion_comment = False)

src/AccessibilityInsights.Extensions.AzureDevOps/Models/Team.cs
```diff
@@ -9,20 +8,20 @@ namespace AccessibilityInsights.Extensions.AzureDevOps.Models
     /// <summary>
     /// Team
     /// </summary>
-    public class Team : AzureDevOpsEntity, ITeam
```
DaveTryon: >ITeam [](start = 43, length = 5)

Should this PR also remove the now-unused interfaces of ITeam, IProject, IConnectionInfo, etc?

src/AccessibilityInsights.Extensions.AzureDevOps/Models/IssueResult.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using AccessibilityInsights.Extensions.Interfaces.IssueReporting;
+using System;
+
+namespace AccessibilityInsights.Extensions.AzureDevOps.Models
+{
+    public class IssueResult : IIssueResult
```
DaveTryon: >IssueResult [](start = 17, length = 11)

Please add some /// comments to clarify the role of this class and its methods

src/AccessibilityInsights.Extensions.AzureDevOps/Models/IssueResult.cs
```diff
@@ -0,0 +1,14 @@
...
+    {
+        public string DisplayText { get; set; }
```
DaveTryon: >set [](start = 41, length = 3)

Should DisplayText and IssueLink be get-only? I can't think of any case where the app would need to change them. The implementing class can have setters even if the abstraction doesn't

src/AccessibilityInsights.Extensions.AzureDevOps/Models/ExtensionConfiguration.cs
```diff
@@ -0,0 +1,27 @@
...
+
+        public ConnectionInfo SavedConnection { get; set; }
+
+        public ConnectionCache CachedConnections { get; set; }
+
+        public string GetSerializedConfig() => JsonConvert.SerializeObject(this);
+
+        public void LoadFromSerializedString(string serializedConfig)
+        {
+            var config = JsonConvert.DeserializeObject<ExtensionConfiguration>(serializedConfig);
```
DaveTryon: >config [](start = 16, length = 6)

Probably worth some error handling here in case the serialized data is unparsable

src/AccessibilityInsights.Extensions.AzureDevOps/AzureBoardsIssueReporting.cs
```diff
@@ -0,0 +1,81 @@
...
+                    {
+                        var success = FileIssueHelpers.AttachIssueData(issueInfo, newIssueId, issueId.Value).Result;
+                        if (!success)
+                        {
+                            //MessageDialog.Show(Properties.Resources.HierarchyControl_FileIssue_There_was_an_error_identifying_the_created_issue_This_may_occur_if_the_ID_used_to_create_the_issue_is_removed_from_its_Azure_DevOps_description_Attachments_have_not_been_uploaded);
+                        }
+                        return new IssueResult() { DisplayText = newIssueId, IssueLink = null };
+                    }
+                    catch (Exception)
+                    {
```
flyingUnderTheRadar: Not for this PR and not specific to this one but we really should not be catching exceptions silently. We should be logging them somewhere is my take on this. This will help us debug any issue much faster. Event log on windows looks like a good candidate.

src/AccessibilityInsights.Extensions.AzureDevOps/ConfigurationControl.xaml.cs
```diff
@@ -0,0 +1,521 @@
...
+                    }
+                    else
+                    {
+                        ToggleLoading(false);
+                        ServerComboBox.Focus();
+                    }
+                }
+                else
+                {
+                    //MessageDialog.Show("URL format is not valid. Example URL: https://accountname.visualstudio.com");
```
flyingUnderTheRadar: Do we want to do something in here? If not we should get rid of the else.

src/AccessibilityInsights.Extensions.AzureDevOps/ConfigurationControl.xaml.cs
```diff
@@ -0,0 +1,521 @@
...
+                        ToggleLoading(true);
+                        projects.Clear();
+                        var newProjectList = await UpdateTeamProjects().ConfigureAwait(true); // need to come back to original UI thread. 
+                        newProjectList.ForEach(p => projects.Add(p));
+                        ToggleLoading(false);
+                    Dispatcher.Invoke(() => serverTreeview.ItemsSource = projects);
+                    }
+                    catch (Exception)
+                    {
+                        //MessageDialog.Show("Error when retrieving team projects");
```
flyingUnderTheRadar: It does look like we want to show this particular error to the user. This can also be done withing the extension UI.

src/AccessibilityInsights.Extensions.AzureDevOps/IssueTemplates/IssueColorContrast.html
```diff
@@ -9,15 +9,15 @@
 <b>To Reproduce and How to Fix guidance:</b> Please open the attached Accessibility Insights for Windows Test File (*.a11ytest) and use the color picker dropdown to find a color that meets the necessary color contrast ratio.
 <br/>
 <br/>
-This accessibility issue was found using <b>Accessibility Insights for Windows</b>, a tool that helps debug and find accessibility issues earlier. Get more information & download this tool at <a href="https://aka.ms/AccessibilityInsights">https://aka.ms/AccessibilityInsights</a>.<br/>
+This accessibility issue was found using <b>Accessibility Insights for Windows</b>, a tool that helps dedebug and find accessibility issues earlier. Get more information & download this tool at <a href="https://aka.ms/AccessibilityInsights">https://aka.ms/AccessibilityInsights</a>.<br/>
```
flyingUnderTheRadar: typo - dedebug - double negation so a bug? :-P

src/AccessibilityInsights.Fakes.Prebuild/Fakes.Prebuild.csproj
```diff
@@ -31,17 +31,53 @@
     <WarningLevel>4</WarningLevel>
   </PropertyGroup>
   <ItemGroup>
```
flyingUnderTheRadar: 🔕

src/AccessibilityInsights.Extensions.AzureDevOps/AzureBoardsIssueReporting.cs
jalkire: Sorted

src/AccessibilityInsights.Extensions.AzureDevOps/AzureBoardsIssueReporting.cs
jalkire: Yes--updated to be get only

src/AccessibilityInsights.Extensions.AzureDevOps/AzureBoardsIssueReporting.cs
jalkire: No more success in this function

src/AccessibilityInsights.Extensions.AzureDevOps/AzureBoardsIssueReporting.cs
jalkire: Nesting eliminated

src/AccessibilityInsights.Extensions.AzureDevOps/AzureDevOpsIntegration.cs
jalkire: Changed to use `CredentialPromptType`, which is what the bool eventually got converted into

src/AccessibilityInsights.Extensions.AzureDevOps/AzureDevOpsIntegration.cs
jalkire: Class and functions changed to be internal

src/MSI/Product.wxs
jalkire: Changed to Issue. I assume that's what you meant

src/AccessibilityInsights.Extensions.AzureDevOpsTests/FileIssue/FileIssueHelpersTests.cs
jalkire: Only for you, Dave

src/AccessibilityInsights.Extensions.AzureDevOps/Models/Team.cs
jalkire: No, they haven't been cleaned up fully on the AI-Win side of things :(

src/AccessibilityInsights.Extensions.AzureDevOps/ConfigurationControl.xaml.cs
jalkire: Yes, I still need to port over MessageDialog

src/AccessibilityInsights.Extensions.AzureDevOps/ConfigurationControl.xaml.cs
jalkire: Yes, I still need to port over MessageDialog

src/AccessibilityInsights.Extensions.AzureDevOps/Models/IssueResult.cs
jalkire: They are get only in the abstraction; this is the impementation.

src/AccessibilityInsights.Extensions.AzureDevOps/AzureDevOpsIntegration.cs
jalkire: It is gone

src/AccessibilityInsights.Extensions.AzureDevOps/Models/IssueResult.cs
jalkire: Commented

src/AccessibilityInsights.Extensions.AzureDevOps/Models/ExtensionConfiguration.cs
jalkire: Errors now caught--I don't think we want to do anything with the exception

src/AccessibilityInsights.Extensions.AzureDevOps/IssueTemplates/IssueColorContrast.html
jalkire: That's one way of looking at it!

src/AccessibilityInsights.Fakes.Prebuild/Fakes.Prebuild.csproj
jalkire: 🔔

PR #193, opened 2019-03-19
Fix empty message box issue

babylonbee:
#### PR checklist

- [ ] Includes UI changes?

#### Describe the change
The PR addresses the issue where it shows an empty message box when user is not logged in and trying to view an existing bug.

### DIFF
--- a/src/AccessibilityInsights.SharedUx/Controls/TestTabs/AutomatedChecksControl.xaml.cs
+++ b/src/AccessibilityInsights.SharedUx/Controls/TestTabs/AutomatedChecksControl.xaml.cs
@@ -972,8 +972,16 @@ namespace AccessibilityInsights.SharedUx.Controls.TestTabs
                 // Bug already filed, open it in a new window
                 try
                 {
-                    var bugUrl = BugReporter.GetExistingBugUriAsync(vm.BugId.Value).Result.ToString();
-                    Process.Start(bugUrl);
+                    var bugUrl = BugReporter.GetExistingBugUriAsync(vm.BugId.Value).Result?.ToString();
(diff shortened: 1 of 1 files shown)


src/AccessibilityInsights.SharedUx/Controls/TestTabs/AutomatedChecksControl.xaml.cs
```diff
@@ -972,8 +972,16 @@ private void btnFileBug_Click(object sender, RoutedEventArgs e)
                 // Bug already filed, open it in a new window
                 try
                 {
-                    var bugUrl = BugReporter.GetExistingBugUriAsync(vm.BugId.Value).Result.ToString();
-                    Process.Start(bugUrl);
+                    var bugUrl = BugReporter.GetExistingBugUriAsync(vm.BugId.Value).Result?.ToString();
+
+                    if (bugUrl == null)
+                    {
+                        MessageDialog.Show("Please sign into Azure DevOps to view the bug information");
```
karanbirsingh: Any strings need to be in resource (.resx) files, so we need a change here; we also cannot assume it will be Azure Devops. cc @jalkire for what we should expect from `GetExistingBugUriAsync` in cases like this so we can have an appropriate error message.

One related existing problem is that we query the bug ID using the current ADO connection, regardless of which account the bug was filed in. I don't think we can address that as part of this PR and it seems like that's not the issue here, but FYI.

babylonbee:
Discussed with @karanbirsingh and @jalkire offline, since this issue will be fixed with @jalkire refactoring code later, close this PR for now.

PR #194, opened 2019-03-19
Update logo when the state of the issue reporter changes

flyingUnderTheRadar:
#### PR checklist

- [ ] Includes UI changes?

#### Describe the change
image

Provides a way to update the logo and the tooltip after config restore (on login) and when we hit save on the connections page. We do not have the tooltip text and the default logo yet. Added the ladybug tentatively.

Inverts the order of the settings and the connection order.

Please ignore all files with a "TestIssue" prefix in the name (TestIssue*)

### DIFF
--- a/src/AccessibilityInsights/MainWindow.xaml
+++ b/src/AccessibilityInsights/MainWindow.xaml
@@ -182,7 +182,7 @@
                                 <Image x:Name="imgAvatar" Visibility="Collapsed" Source="{Binding Path=vmAvatar.ByteData, TargetNullValue={x:Null}}"/>
                                 <Grid x:Name="newAccountGrid" Visibility="Visible">
                                     <AccessText Text="_i" Opacity="0"/>
-                                    <fabric:FabricIconControl GlyphName="Contact" FontSize="24" VerticalAlignment="Center" Foreground="{DynamicResource ResourceKey=LightIconBrush}"  HorizontalAlignment="Center"/>
+                                    <fabric:FabricIconControl GlyphName="LadybugSolid" FontSize="24" VerticalAlignment="Center" Foreground="{DynamicResource ResourceKey=LightIconBrush}"  HorizontalAlignment="Center"/>
                                 </Grid>
                             </Grid>
                         </Button>
--- a/src/AccessibilityInsights/MainWindow.xaml
+++ b/src/AccessibilityInsights/MainWindow.xaml
@@ -146,25 +146,7 @@
                             <RowDefinition/>
                             <RowDefinition/>
                         </Grid.RowDefinitions>
-                        <Button Width="40" Height="40"
-                                    x:Name="btnConfig"
-                                    AutomationProperties.Name="{x:Static properties:Resources.btnConfigAutomationPropertiesName}"
-                                    AutomationProperties.HelpText="{x:Static properties:Resources.btnConfigAutomationPropertiesHelpText}"
-                                    Style="{StaticResource btnLeftNav}" Click="onbtnConfigClicked">
-                            <Button.ToolTip>
-                                <ToolTip>
-                                    <Run Text="{x:Static properties:Resources.btnConfigToolTip}" />
-                                </ToolTip>
-                            </Button.ToolTip>
-                            <i:Interaction.Behaviors>
-                                <behaviors:KeyboardToolTipButtonBehavior/>
-                            </i:Interaction.Behaviors>
-                            <Grid>
-                                <AccessText Text="_c" Opacity="0"/>
-                                <fabric:FabricIconControl GlyphName="Settings" FontSize="24" VerticalAlignment="Center" HorizontalAlignment="Center" Foreground="{DynamicResource ResourceKey=LightIconBrush}"/>
-                            </Grid>
-                        </Button>
-                        <Button Width="40" Height="40" Grid.Row="1"
+                        <Button Width="40" Height="40" Grid.Row="0"
                             Style="{StaticResource btnLeftNav}"
                             x:Name="btnAccountConfig"
                             Click="btnAccountConfig_Click"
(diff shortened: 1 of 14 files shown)


src/AccessibilityInsights/MainWindow.xaml.cs
```diff
@@ -704,10 +704,32 @@ private static void RestoreConfigurationAsync(/*Action callback = null*/)
...
 
+        /// <summary>
+        /// Update sign in logo and tooltip
+        /// </summary>
+        internal void UpdateMainWindowConnectionFields()
+        {
+            bool isConfigured = BugReporter.IssueReporting != null && BugReporter.IsConnected;
+            
+            // Main window UI changes
+            vmAvatar.ByteData = isConfigured ? BugReporter.Logo : null;
```
jalkire: Based on the figma, our logos are going to be fabric icons. We may want to expose a GlyphName string instead of an `IEnumerable<byte>` from the IssueReporter now and remove `vmAvatar`. Thoughts?

DaveTryon: I'd prefer to see an enum of supported values in the interface defined in the Extensions assembly. That lets us still remain loosely coupled, but with a well-defined mechanism for extending the model in the future. One note: Since it's possible to cast any int to an enum, the code that converts from the enum to the actual glyph would need have a default case to map nonsensical values to some predetermined default

---
In reply to: [267126925](https://github.com/Microsoft/accessibility-insights-windows/pull/194#discussion_r267126925) [](ancestors = 267126925)

flyingUnderTheRadar: Can fabric icons be converted to a IEnumerable<byte> form? If yes we should go with that or ask for these icons to be provided to us in a image format. We cannot opt for fabric icons as not all the extensions will have one in the fabric icons set.

DaveTryon: Fabric icons are currently implemented in the DesktopUI assembly, and the issue filing assembly is allowed to have direct dependencies **only** on the Extensions definition and the Win32 assemblies. That's part of having the loose coupling that is part of the extensions design.

---
In reply to: [267137502](https://github.com/Microsoft/accessibility-insights-windows/pull/194#discussion_r267137502) [](ancestors = 267137502)

jalkire: @flyingUnderTheRadar The only reason we are using `IEnumerable<byte>` is because that is what the ADO profile pic is given to us as. Since we are no longer using that, there is no reason to retain its format. 
> We cannot opt for fabric icons as not all the extensions will have one in the fabric icons set.
One other option would be to take a control and display it here. I would be fine with crossing this bridge when we get to it and using the GlyphName for now.

@DaveTryon I've noticed that we have a few contorls (FabricIcon and MessageDialog specifically) which are a. Useful in the extensions and b. not dependent on anything in AI-Win. Should we consider creating a new project which allows us to share decoupled controls like this accross extensions?

src/AccessibilityInsights/MainWindow.xaml
```diff
@@ -182,7 +182,7 @@
                                 <Image x:Name="imgAvatar" Visibility="Collapsed" Source="{Binding Path=vmAvatar.ByteData, TargetNullValue={x:Null}}"/>
                                 <Grid x:Name="newAccountGrid" Visibility="Visible">
                                     <AccessText Text="_i" Opacity="0"/>
-                                    <fabric:FabricIconControl GlyphName="Contact" FontSize="24" VerticalAlignment="Center" Foreground="{DynamicResource ResourceKey=LightIconBrush}"  HorizontalAlignment="Center"/>
+                                    <fabric:FabricIconControl GlyphName="LadybugSolid" FontSize="24" VerticalAlignment="Center" Foreground="{DynamicResource ResourceKey=LightIconBrush}"  HorizontalAlignment="Center"/>
                                 </Grid>
```
neocln: did we validate it with PM?

src/AccessibilityInsights/MainWindow.xaml.cs
```diff
@@ -704,10 +704,30 @@ private static void RestoreConfigurationAsync(/*Action callback = null*/)
                 var serializedConfigsDict = appConfig.IssueReporterSerializedConfigs;
                 Dictionary<Guid, string> configsDictionary = JsonConvert.DeserializeObject<Dictionary<Guid, string>>(serializedConfigsDict);
                 configsDictionary.TryGetValue(selectedIssueReporterGuid, out string serializedConfig);
-                BugReporter.RestoreConfigurationAsync(serializedConfig);
+                await BugReporter.RestoreConfigurationAsync(serializedConfig).ConfigureAwait(true);
+                UpdateMainWindowConnectionFields();
```
neocln: Is this part of code updating UI? if so, we need to call it with Dispatcher to run on UI thread.

src/AccessibilityInsights/MainWindow.xaml.cs
flyingUnderTheRadar: I would actually argue that other extensions (3rd party) might not necessarily have representative icons in office fabric so just accepting a generic image would have been great. We discussed this with Risa and she mentioned that we should go ahead with Office Fabric icons for now and that we will change when that case shows up.

Looking into how we can make this change.

src/AccessibilityInsights/MainWindow.xaml
flyingUnderTheRadar: We didn't it was tentative. We now have confirmation from PM. I have updated the screenshot with the latest logo.

src/AccessibilityInsights/MainWindow.xaml.cs
flyingUnderTheRadar: Discussed offline where we decided to go with a dispatcher call despit configureawait(true)

src/AccessibilityInsights/MainWindow.xaml.cs
flyingUnderTheRadar: Ended up following the enum approach.

src/AccessibilityInsights/MainWindow.xaml
```diff
@@ -179,13 +161,30 @@
                                 <behaviors:KeyboardToolTipButtonBehavior/>
                             </i:Interaction.Behaviors>
                             <Grid Width="40" Height="40" x:Name="addAccountGrid">
-                                <Image x:Name="imgAvatar" Visibility="Collapsed" Source="{Binding Path=vmAvatar.ByteData, TargetNullValue={x:Null}}"/>
                                 <Grid x:Name="newAccountGrid" Visibility="Visible">
                                     <AccessText Text="_i" Opacity="0"/>
-                                    <fabric:FabricIconControl GlyphName="Contact" FontSize="24" VerticalAlignment="Center" Foreground="{DynamicResource ResourceKey=LightIconBrush}"  HorizontalAlignment="Center"/>
+                                    <fabric:FabricIconControl GlyphName="{Binding Path=vmReporterLogo.FabricIconLogoName, TargetNullValue=PlugDisconnected}" FontSize="24" VerticalAlignment="Center" Foreground="{DynamicResource ResourceKey=LightIconBrush}"  HorizontalAlignment="Center"/>
                                 </Grid>
```
neocln: Does it mean that we don't show a user profile picture?

jalkire: Correct, it will only be visible in the extension's configurion in the connection settings tab.

PR #195, opened 2019-03-20
Improve integration with new issue filing interfaces

jalkire:
#### PR checklist

- [ ] Includes UI changes?

#### Describe the change
- Handle case where connection config is null in ConnectionControl.
- Re-selected saved issue reporter
- Fix serialization issues with Guid

### DIFF
--- a/src/AccessibilityInsights.SharedUx/Controls/SettingsTabs/ConnectionControl.xaml.cs
+++ b/src/AccessibilityInsights.SharedUx/Controls/SettingsTabs/ConnectionControl.xaml.cs
@@ -89,12 +89,12 @@ namespace AccessibilityInsights.SharedUx.Controls.SettingsTabs
             if (issueConfigurationControl.CanSave)
             {
                 configuration.SelectedIssueReporter = selectedIssueReporter.StableIdentifier;
-                string seralizedConfigs = configuration.IssueReporterSerializedConfigs;
+                string serializedConfigs = configuration.IssueReporterSerializedConfigs;
                 Dictionary<Guid, string> configs = new Dictionary<Guid, string>();
 
-                if (seralizedConfigs != null)
+                if (serializedConfigs != null)
                 {
-                    configs = JsonConvert.DeserializeObject<Dictionary<Guid, string>>(seralizedConfigs);
+                    configs = JsonConvert.DeserializeObject<Dictionary<Guid, string>>(serializedConfigs);
                 }
 
                 string newConfigs = issueConfigurationControl.OnSave();
--- a/src/AccessibilityInsights.SharedUx/Controls/SettingsTabs/ConnectionControl.xaml.cs
+++ b/src/AccessibilityInsights.SharedUx/Controls/SettingsTabs/ConnectionControl.xaml.cs
@@ -90,7 +90,12 @@ namespace AccessibilityInsights.SharedUx.Controls.SettingsTabs
             {
                 configuration.SelectedIssueReporter = selectedIssueReporter.StableIdentifier;
                 string seralizedConfigs = configuration.IssueReporterSerializedConfigs;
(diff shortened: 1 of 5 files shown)


src/AccessibilityInsights.SharedUx/Controls/SettingsTabs/ConnectionControl.xaml.cs
```diff
@@ -90,7 +90,12 @@ public void UpdateConfigFromSelections(ConfigurationModel configuration)
             {
                 configuration.SelectedIssueReporter = selectedIssueReporter.StableIdentifier;
                 string seralizedConfigs = configuration.IssueReporterSerializedConfigs;
-                Dictionary<Guid, string> configs = JsonConvert.DeserializeObject<Dictionary<Guid, string>>(seralizedConfigs);
+                Dictionary<Guid, string> configs = new Dictionary<Guid, string>();
+
+                if (seralizedConfigs != null)
```
DaveTryon: >seralizedConfigs [](start = 20, length = 16)

Typo (missing the 'i' in "serial")

src/AccessibilityInsights.SharedUx/Settings/ConfigurationModel.cs
```diff
@@ -126,8 +126,8 @@ public string AppVersion
 
         public Guid SelectedIssueReporter
         {
-            get => GetDataValue<Guid>(keyIssueReporting);
-            set => SetDataValue<Guid>(keyIssueReporting, value);
+            get => Guid.Parse(GetDataValue<string>(keyIssueReporting));
+            set => SetDataValue<string>(keyIssueReporting, value.ToString());
```
DaveTryon: >ToString() [](start = 65, length = 10)

Guid.ToString supports an optional formatting parameter (see [https://docs.microsoft.com/en-us/dotnet/api/system.guid.tostring?view=netframework-4.7.2](https://docs.microsoft.com/en-us/dotnet/api/system.guid.tostring?view=netframework-4.7.2)). Are we OK with the default format?

src/AccessibilityInsights.SharedUx/Settings/ConfigurationConstants.cs
```diff
@@ -71,7 +71,7 @@ public partial class ConfigurationModel
         private const string keyTestReportPath = "TestReportPath";
         private const string keyTreeViewMode = "TreeViewMode";
         private const string keyVersion = "Version";
-        private const string keyIssueReporting= "IssueReportinGuid";
+        private const string keyIssueReporting= "IssueReportingGuid";
```
DaveTryon: >IssueReportingGuid [](start = 49, length = 18)

There should be a unit test that ensures that this data is correct in the JSON, and that unit test should need to change as a result of this. I see no unit test change, so it seems that perhaps the unit test was omitted originally. Can we please add it now?

src/AccessibilityInsights.SharedUx/Settings/ConfigurationModel.cs
jalkire: Guid.Parse should work regardless of format, so I think we're fine with the default:
>This method can convert strings in any of the five formats produced by the ToString
- https://docs.microsoft.com/en-us/dotnet/api/system.guid.parse?view=netframework-4.7.2

src/AccessibilityInsights.SharedUx/Settings/ConfigurationConstants.cs
jalkire: @flyingUnderTheRadar is working on the necessary unit test updates for this generally. The ConfigurationModelTests have already been broken in the feature branch--I would rather have them fixed/updated all at once than start the process in this PR.

src/AccessibilityInsights.SharedUx/Controls/SettingsTabs/ConnectionControl.xaml.cs
jalkire: 👓

PR #209, opened 2019-03-21
Validate GitHub Repo Link Input Format

AhmedAbdoOrtiga:
#### PR checklist

- [ ] Includes UI changes?

#### Describe the change
Make Sure that the GitHub repo link is valid before saving it.

### DIFF
--- a/src/AccessibilityInsights.Extensions.GitHub/LinkValidator.cs
+++ b/src/AccessibilityInsights.Extensions.GitHub/LinkValidator.cs
@@ -16,6 +16,7 @@ namespace AccessibilityInsights.Extensions.GitHub
 
         public static bool IsValidGitHubRepoLink(string Link)
         {
+            Link = Link.Replace(@"\", "/").Trim(' ').TrimEnd('/');
             string UserNamePattern = string.Format(CultureInfo.InvariantCulture, Properties.Resources.UserNamePattern, AlphaNumricPattern);
             string RepoNamePattern = string.Format(CultureInfo.InvariantCulture, Properties.Resources.RepoNamePattern, AlphaNumricPattern);
             string LinkPatttern = string.Format(CultureInfo.InvariantCulture, Properties.Resources.LinkPatttern, GitHubLink, UserNamePattern, RepoNamePattern);
@@ -25,7 +26,8 @@ namespace AccessibilityInsights.Extensions.GitHub
                 return false;
             }
 
-            string[] parts = Regex.Replace(Link, GitHubLink, "").Split('/');
+            Link = Regex.Replace(Link, GitHubLink, "");
+            string[] parts = Regex.Split(Link, @"/+");
             if (parts.Length != 2)
             {
                 return false;
--- a/src/AccessibilityInsights.Extensions.GitHub/LinkValidator.cs
+++ b/src/AccessibilityInsights.Extensions.GitHub/LinkValidator.cs
@@ -6,6 +6,9 @@ using System.Text.RegularExpressions;
 
 namespace AccessibilityInsights.Extensions.GitHub
 {
+    /// <summary>
+    /// GitHub Repo Link Format Validator
+    /// </summary>
     public static class LinkValidator
     {
         private static readonly string GitHubLink = Properties.Resources.GitHubLink;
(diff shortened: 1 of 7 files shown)


src/AccessibilityInsights.Extensions.GitHub/Properties/Resources.resx
```diff
@@ -159,4 +159,22 @@ This accessibility issue was found using Accessibility Insights for Windows, a t
   <data name="tbURLPlaceHolder" xml:space="preserve">
     <value>Enter desired GitHub repo link</value>
   </data>
+  <data name="AlphaNumricPattern" xml:space="preserve">
+    <value>[a-zA-Z0-9]</value>
```
jalkire: I don't think these need to be in resources because they are essentially programmatic logic in string form, not user-facing strings. But I'm not opposed to keeping them here if you prefer

src/AccessibilityInsights.Extensions.GitHub/LinkValidator.cs
```diff
@@ -0,0 +1,65 @@
...
+{
+    /// <summary>
+    /// GitHub Repo Link Format Validator
+    /// </summary>
+    public static class LinkValidator
+    {
+        private static readonly string GitHubLink = Properties.Resources.GitHubLink;
+        private static readonly string AlphaNumricPattern = Properties.Resources.AlphaNumricPattern;
+
+        public static bool IsValidGitHubRepoLink(string Link)
```
flyingUnderTheRadar: would it be a good idea to support both "www.github.com/Microsoft/ai-win" and "www.github.com/Microsoft/ai-win/issues". That way we are more robust. If they have /issue at the end we simply ignore it and go ahead with business as usual. We could also punt on that (according to the discussion today morning)

src/AccessibilityInsights.Extensions.GitHub/LinkValidator.cs
```diff
@@ -0,0 +1,65 @@
...
+        {
+            string UserNamePattern = string.Format(CultureInfo.InvariantCulture, Properties.Resources.UserNamePattern, AlphaNumricPattern);
+            string RepoNamePattern = string.Format(CultureInfo.InvariantCulture, Properties.Resources.RepoNamePattern, AlphaNumricPattern);
+            string LinkPatttern = string.Format(CultureInfo.InvariantCulture, Properties.Resources.LinkPatttern, GitHubLink, UserNamePattern, RepoNamePattern);
+            Regex gitHubRepoLinRegex = new Regex(LinkPatttern, RegexOptions.IgnoreCase);
+            if (!gitHubRepoLinRegex.IsMatch(Link))
```
flyingUnderTheRadar: typo - gitHubRepoLinRegex => gitHubRepoLin*k*Regex

src/AccessibilityInsights/AccessibilityInsights.csproj
```diff
@@ -260,6 +260,14 @@
       <Project>{b2fa5150-470a-4864-a191-7bcae39df35a}</Project>
       <Name>Actions</Name>
     </ProjectReference>
+    <ProjectReference Include="..\AccessibilityInsights.Extensions.AzureDevOps\Extensions.AzureDevOps.csproj">
```
flyingUnderTheRadar: We probably don't want these references

src/AccessibilityInsights.Extensions.GitHub/LinkValidator.cs
AhmedAbdoOrtiga: I agree we need to support "www.GithUB" and "https://www.GitHub.com" but I don't think we need to handle "issues" as we agreed in the meeting, I remember the discussion was to keep it without issues and have the web team to remove it from the link

src/AccessibilityInsights.Extensions.GitHub/Properties/Resources.resx
AhmedAbdoOrtiga: let's keep all the strings in the resource file.

src/AccessibilityInsights.Extensions.GitHub/LinkValidator.cs
flyingUnderTheRadar: Yes the discussion was to have the web team remove issues but we also said time permitting we should make it robust. In this case accepting the"/issues" at the end should be ignored whenever we get to making the application more robust. Not necessary as a part of this pr

PR #214, opened 2019-03-22
Use IE feature control to change how webpages are rendered in WebBrowser

karanbirsingh:
#### PR checklist

- [ ] Includes UI changes?

#### Describe the change

We have occasionally noticed the WebBrowser used in the AzureDevops extension is unable to render the page and instead shows a white screen (even though viewing the source shows proper HTML). This PR applies a suggested fix via feature controls for a similar problem. IE provides several feature controls in order to enhance the way the WebBrowser behaves. Earlier we set the Browser Emulation registry key to a high number [99999] which doesn't seem accurate - according to the docs and suggested fix, this key should be set to the values defined [here](https://docs.microsoft.com/en-us/previous-versions/windows/internet-explorer/ie-developer/general-info/ee330730%28v%3dvs.85%29). This affects what 'mode' web pages are rendered in (read more [here](https://docs.microsoft.com/en-us/previous-versions/windows/internet-explorer/ie-developer/compatibility/cc288325(v=vs.85))). In this PR the value is chosen based on the installed version of IE. This change was tested on a few machines from our team and those with the white screen issue reported successful rendering afterwards. There is one case where the issue still exists - will have to troubleshoot on their box. We are exploring other options to fix or replace the WebBrowser; in the meantime this fix should improve the situation for some users.

### DIFF
--- a/src/AccessibilityInsights.SharedUx/FileBug/BrowserEmulation.cs
+++ b/src/AccessibilityInsights.SharedUx/FileBug/BrowserEmulation.cs
@@ -1,10 +1,6 @@
 ﻿using Microsoft.Win32;
 using System;
-using System.Collections.Generic;
 using System.Diagnostics;
-using System.Linq;
-using System.Text;
-using System.Threading.Tasks;
 
 namespace AccessibilityInsights.SharedUx.FileBug
 {
--- a/src/AccessibilityInsights.SharedUx/FileBug/BrowserEmulation.cs
+++ b/src/AccessibilityInsights.SharedUx/FileBug/BrowserEmulation.cs
@@ -20,72 +20,61 @@ namespace AccessibilityInsights.SharedUx.FileBug
             // http://msdn.microsoft.com/en-us/library/ee330720(v=vs.85).aspx
 
             // FeatureControl settings are per-process
-            dynamic fileName = System.IO.Path.GetFileName(Process.GetCurrentProcess().MainModule.FileName);
+            var fileName = System.IO.Path.GetFileName(Process.GetCurrentProcess().MainModule.FileName);
 
-            // make the control is not running inside Visual Studio Designer
-            if (String.Compare(fileName, "devenv.exe", true) == 0 || String.Compare(fileName, "XDesProc.exe", true) == 0)
+            // make sure the control is not running inside Visual Studio Designer
+            if (string.Compare(fileName, "devenv.exe", true, System.Globalization.CultureInfo.InvariantCulture) == 0 || string.Compare(fileName, "XDesProc.exe", true, System.Globalization.CultureInfo.InvariantCulture) == 0)
             {
                 return;
             }
 
-            SetBrowserFeatureControlKey("FEATURE_BROWSER_EMULATION", fileName, GetBrowserEmulationMode());
-            // Webpages containing standards-based !DOCTYPE directives are displayed in IE10 Standards mode.
-            SetBrowserFeatureControlKey("FEATURE_AJAX_CONNECTIONEVENTS", fileName, 1);
-            SetBrowserFeatureControlKey("FEATURE_GPU_RENDERING", fileName, 1);
+            try
+            {
+                SetBrowserFeatureControlKey("FEATURE_BROWSER_EMULATION", fileName, GetBrowserEmulationMode());
+                SetBrowserFeatureControlKey("FEATURE_AJAX_CONNECTIONEVENTS", fileName, 1);
+                SetBrowserFeatureControlKey("FEATURE_GPU_RENDERING", fileName, 1);
+            } 
+            catch (Exception)
+            {
+                // silently ignore
+            }
         }
 
         private static void SetBrowserFeatureControlKey(string feature, string appName, uint value)
         {
-            using (var key = Registry.CurrentUser.CreateSubKey(String.Concat("Software\\Microsoft\\Internet Explorer\\Main\\FeatureControl\\", feature), RegistryKeyPermissionCheck.ReadWriteSubTree))
+            using (var key = Registry.CurrentUser.CreateSubKey(String.Concat(@"Software\Microsoft\Internet Explorer\Main\FeatureControl\", feature), RegistryKeyPermissionCheck.ReadWriteSubTree))
             {
-                key.SetValue(appName, (UInt32)value, RegistryValueKind.DWord);
+                key.SetValue(appName, value, RegistryValueKind.DWord);
             }
         }
 
-        private static UInt32 GetBrowserEmulationMode()
+        /// <summary>
+        /// Get the browser emulation mode most appropriate for the installed version of IE,
+        /// or the default mode for applications hosting the WebBrowser control
+        /// </summary>
+        /// <returns></returns>
+        private static uint GetBrowserEmulationMode()
         {
             int browserVersion = 7;
-            using (var Key = Registry.LocalMachine.OpenSubKey("SOFTWARE\\Microsoft\\Internet Explorer", RegistryKeyPermissionCheck.ReadSubTree, System.Security.AccessControl.RegistryRights.QueryValues))
+            using (var Key = Registry.LocalMachine.OpenSubKey(@"SOFTWARE\Microsoft\Internet Explorer", RegistryKeyPermissionCheck.ReadSubTree, System.Security.AccessControl.RegistryRights.QueryValues))
             {
-                dynamic version = Key.GetValue("svcVersion");
-                if (null == version)
+                var version = Key.GetValue("svcVersion") ?? Key.GetValue("Version");
+                if (version == null || int.TryParse(version.ToString().Split('.')[0], out browserVersion) == false)
                 {
-                    version = Key.GetValue("Version");
-                    if (null == version)
-                    {
-                        throw new ApplicationException("Microsoft Internet Explorer is required!");
-                    }
+                    return 7000;
                 }
-                int.TryParse(version.ToString().Split('.')[0], out browserVersion);
             }
-
-            UInt32 mode = 10000;
-            // Internet Explorer 10. Webpages containing standards-based !DOCTYPE directives are displayed in IE10 Standards mode. Default value for Internet Explorer 10.
+            
             switch (browserVersion)
             {
-                case 7:
-                    mode = 7000;
-                    // Webpages containing standards-based !DOCTYPE directives are displayed in IE7 Standards mode. Default value for applications hosting the WebBrowser Control.
-                    break;
-                case 8:
-                    mode = 8000;
-                    // Webpages containing standards-based !DOCTYPE directives are displayed in IE8 mode. Default value for Internet Explorer 8
-                    break;
-                case 9:
-                    mode = 9000;
-                    // Internet Explorer 9. Webpages containing standards-based !DOCTYPE directives are displayed in IE9 mode. Default value for Internet Explorer 9.
-                    break;
-                case 10:
-                    mode = 10000;
-                    // Internet Explorer 10. Webpages containing standards-based !DOCTYPE directives are displayed in IE9 mode. Default value for Internet Explorer 9.
-                    break;
-                case 11:
-                    mode = 11001;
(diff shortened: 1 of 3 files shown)


src/AccessibilityInsights.SharedUx/FileBug/BrowserEmulation.cs
```diff
@@ -0,0 +1,91 @@
...
+    /// https://docs.microsoft.com/en-us/previous-versions/windows/internet-explorer/ie-developer/general-info/ee330730%28v%3dvs.85%29#browser-emulation
+    /// </summary>
+    public static class BrowserEmulation
+    {
+        public static void SetBrowserFeatureControl()
+        {
+            // http://msdn.microsoft.com/en-us/library/ee330720(v=vs.85).aspx
+
+            // FeatureControl settings are per-process
+            dynamic fileName = System.IO.Path.GetFileName(Process.GetCurrentProcess().MainModule.FileName);
```
RobGallo: Is var or string not more appropriate here? I think dynamic may have some overhead which is unnecessary in this case.

src/AccessibilityInsights.SharedUx/FileBug/BrowserEmulation.cs
```diff
@@ -0,0 +1,91 @@
...
+
+            // make the control is not running inside Visual Studio Designer
```
RobGallo:
```suggestion
            // make sure the control is not running inside Visual Studio Designer
```

src/AccessibilityInsights.SharedUx/FileBug/BrowserEmulation.cs
```diff
@@ -0,0 +1,91 @@
...
+                key.SetValue(appName, (UInt32)value, RegistryValueKind.DWord);
+            }
+        }
+
+        private static UInt32 GetBrowserEmulationMode()
+        {
+            int browserVersion = 7;
+            using (var Key = Registry.LocalMachine.OpenSubKey("SOFTWARE\\Microsoft\\Internet Explorer", RegistryKeyPermissionCheck.ReadSubTree, System.Security.AccessControl.RegistryRights.QueryValues))
+            {
+                dynamic version = Key.GetValue("svcVersion");
```
RobGallo: Dynamic again.

src/AccessibilityInsights.SharedUx/FileBug/BrowserEmulation.cs
```diff
@@ -0,0 +1,91 @@
...
+                if (null == version)
+                {
+                    version = Key.GetValue("Version");
+                    if (null == version)
+                    {
+                        throw new ApplicationException("Microsoft Internet Explorer is required!");
```
RobGallo: Do we catch this anywhere?

src/AccessibilityInsights.SharedUx/FileBug/BrowserEmulation.cs
```diff
@@ -0,0 +1,91 @@
...
+
+            SetBrowserFeatureControlKey("FEATURE_BROWSER_EMULATION", fileName, GetBrowserEmulationMode());
+            // Webpages containing standards-based !DOCTYPE directives are displayed in IE10 Standards mode.
+            SetBrowserFeatureControlKey("FEATURE_AJAX_CONNECTIONEVENTS", fileName, 1);
+            SetBrowserFeatureControlKey("FEATURE_GPU_RENDERING", fileName, 1);
+        }
+
+        private static void SetBrowserFeatureControlKey(string feature, string appName, uint value)
+        {
+            using (var key = Registry.CurrentUser.CreateSubKey(String.Concat("Software\\Microsoft\\Internet Explorer\\Main\\FeatureControl\\", feature), RegistryKeyPermissionCheck.ReadWriteSubTree))
```
DaveTryon:
```suggestion
            using (var key = Registry.CurrentUser.CreateSubKey(String.Concat(@"Software\Microsoft\Internet Explorer\Main\FeatureControl\", feature), RegistryKeyPermissionCheck.ReadWriteSubTree))
```

src/AccessibilityInsights.SharedUx/FileBug/BrowserEmulation.cs
```diff
@@ -0,0 +1,91 @@
...
+            {
+                key.SetValue(appName, (UInt32)value, RegistryValueKind.DWord);
+            }
+        }
+
+        private static UInt32 GetBrowserEmulationMode()
+        {
+            int browserVersion = 7;
+            using (var Key = Registry.LocalMachine.OpenSubKey("SOFTWARE\\Microsoft\\Internet Explorer", RegistryKeyPermissionCheck.ReadSubTree, System.Security.AccessControl.RegistryRights.QueryValues))
```
DaveTryon:
```suggestion
            using (var Key = Registry.LocalMachine.OpenSubKey(@"SOFTWARE\Microsoft\Internet Explorer", RegistryKeyPermissionCheck.ReadSubTree, System.Security.AccessControl.RegistryRights.QueryValues))
```

src/AccessibilityInsights.SharedUx/FileBug/BrowserEmulation.cs
```diff
@@ -0,0 +1,91 @@
...
+    /// <summary>
+    /// This class is adapted from https://stackoverflow.com/questions/18333459/c-sharp-webbrowser-ajax-call
+    /// and sets several registry values in order to ensure that the hosted WebBrowser control renders
+    /// documents in standards mode. Without these changes the hosted WebBrowser is unable to properly
+    /// render some pages when filing issues
+    /// https://docs.microsoft.com/en-us/previous-versions/windows/internet-explorer/ie-developer/general-info/ee330730%28v%3dvs.85%29#browser-emulation
+    /// </summary>
+    public static class BrowserEmulation
+    {
+        public static void SetBrowserFeatureControl()
```
DaveTryon: >SetBrowserFeatureControl [](start = 27, length = 24)

Registry methods can throw. Do we need a try/catch here somewhere?

src/AccessibilityInsights.SharedUx/FileBug/BrowserEmulation.cs
```diff
@@ -0,0 +1,91 @@
...
+                    mode = 10000;
+                    // Internet Explorer 10. Webpages containing standards-based !DOCTYPE directives are displayed in IE9 mode. Default value for Internet Explorer 9.
+                    break;
+                case 11:
+                    mode = 11001;
+                    // Internet Explorer 11. Webpages containing standards-based !DOCTYPE directives are displayed in IE9 mode. Default value for Internet Explorer 9.
+                    break;
+            }
+
+            return mode;
```
DaveTryon: In a switch statement like this, it's sometimes clearer with a more direct approach, such as
```suggestion
                    case 7: return 7000;
                    case 8: return 8000; 
                    case 9: return 9000;
                    case 11: return 11001
                    case 10:
                    default:
                        return 10000;
```

src/AccessibilityInsights.SharedUx/FileBug/BrowserEmulation.cs
```diff
@@ -0,0 +1,91 @@
...
+                    version = Key.GetValue("Version");
+                    if (null == version)
+                    {
+                        throw new ApplicationException("Microsoft Internet Explorer is required!");
+                    }
+                }
+                int.TryParse(version.ToString().Split('.')[0], out browserVersion);
+            }
+
+            UInt32 mode = 10000;
```
DaveTryon: In a switch statement like this, it's sometimes clearer with a more direct approach, such as
```suggestion
                    case 7: return 7000;
                    case 8: return 8000; 
                    case 9: return 9000;
                    case 11: return 11001
                    case 10:
                    default:
                        return 10000;
```

DaveTryon: >UInt32 [](start = 12, length = 6)

In a switch statement like this, it's sometimes clearer with a more direct approach, such as
switch (browserVersion)
{
    case 7: return 7000;  // Relevant comment
    case 8: return 8000;  // Relevant comment
    case 9: return 9000;  // Relevant comment
    case 11: return 11001;  // Relevant comment
    case 10:
    default:
        return 10000;  // Relevant comment
}

src/AccessibilityInsights.SharedUx/FileBug/FileBugAction.cs
```diff
@@ -292,10 +292,7 @@ private static bool GuidsMatchInReproSteps(string targetGuid, string reproSteps)
         private static int? FileBugWindow(Uri url, bool onTop, int zoomLevel, Action<int> updateZoom)
         {
             System.Diagnostics.Trace.WriteLine(Invariant($"Url is {url.AbsoluteUri.Length} long: {url}"));
-            // To force latest IE to show up
-            var appName = System.Diagnostics.Process.GetCurrentProcess().ProcessName + ".exe";
-            using (var Key = Microsoft.Win32.Registry.CurrentUser.OpenSubKey(@"SOFTWARE\Microsoft\Internet Explorer\Main\FeatureControl\FEATURE_BROWSER_EMULATION", true))
-                Key.SetValue(appName, 99999, Microsoft.Win32.RegistryValueKind.DWord);
+            BrowserEmulation.SetBrowserFeatureControl();
```
DaveTryon: >SetBrowserFeatureControl [](start = 29, length = 24)

Will this code need to move into the AzureDevOps extension?

src/AccessibilityInsights.SharedUx/FileBug/BrowserEmulation.cs
karanbirsingh: next iteration returns the default value & also wraps the call in a try/catch given Dave's comment about registry methods potentially throwing

src/AccessibilityInsights.SharedUx/FileBug/BrowserEmulation.cs
karanbirsingh: updated - we would have to create the constants ourselves; it's a bit of an odd case since I wasn't sure how much of the original code to leave intact. Happy to create the constants if you'd prefer

src/AccessibilityInsights.SharedUx/FileBug/FileBugAction.cs
karanbirsingh: Yes, it will need to move. I will let @jalkire know about this change

src/AccessibilityInsights.SharedUx/FileBug/BrowserEmulation.cs
```diff
@@ -0,0 +1,80 @@
...
+
+            // FeatureControl settings are per-process
+            var fileName = System.IO.Path.GetFileName(Process.GetCurrentProcess().MainModule.FileName);
+
+            // make sure the control is not running inside Visual Studio Designer
+            if (string.Compare(fileName, "devenv.exe", true, System.Globalization.CultureInfo.InvariantCulture) == 0 || string.Compare(fileName, "XDesProc.exe", true, System.Globalization.CultureInfo.InvariantCulture) == 0)
+            {
+                return;
+            }
+
```
neocln: this part of code is not necessary for us since we are not devenv or XDesProc.

src/AccessibilityInsights.SharedUx/FileBug/BrowserEmulation.cs
```diff
@@ -0,0 +1,74 @@
...
+            try
+            {
+                SetBrowserFeatureControlKey("FEATURE_BROWSER_EMULATION", fileName, GetBrowserEmulationMode());
+                SetBrowserFeatureControlKey("FEATURE_AJAX_CONNECTIONEVENTS", fileName, 1);
+                SetBrowserFeatureControlKey("FEATURE_GPU_RENDERING", fileName, 1);
+            }
+            catch (Exception)
+            {
+                // silently ignore
```
DaveTryon: >// silently ignore [](start = 16, length = 18)

Perhaps call ReportException on this Exception, so that we have some visibility that it's happening?

PR #225, opened 2019-03-27
Add appropriate feature control registry keys to improve WebBrowser rendering

karanbirsingh:
We have occasionally noticed the WebBrowser used in the AzureDevops extension is unable to render the page and instead shows a white screen (even though viewing the source shows proper HTML). This PR applies a potential fix via feature controls. IE provides several feature controls in order to enhance the way the WebBrowser behaves. Earlier we set the Browser Emulation registry key to a high number [99999] which doesn't seem accurate - according to the docs and suggested fix, this key should be set to the values defined [here](https://docs.microsoft.com/en-us/previous-versions/windows/internet-explorer/ie-developer/general-info/ee330730(v%3dvs.85)#browser-emulation). This affects what 'mode' web pages are rendered in (read more [here](https://docs.microsoft.com/en-us/previous-versions/windows/internet-explorer/ie-developer/general-info/ee330730%28v%3dvs.85%29)). In this PR the value is chosen based on the installed version of IE. This change was tested on a few machines from our team and those with the white screen issue reported successful rendering afterwards. There is one case where the issue still exists - will have to troubleshoot on their box. We are exploring other options to fix or replace the WebBrowser; in the meantime this fix should improve the situation for some users.

### DIFF
--- a/src/AccessibilityInsights.SharedUx/FileBug/IEBrowserEmulation.cs
+++ b/src/AccessibilityInsights.SharedUx/FileBug/IEBrowserEmulation.cs
@@ -1,0 +1,71 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using AccessibilityInsights.Desktop.Telemetry;
+using Microsoft.Win32;
+using System;
+using System.Diagnostics;
+
+namespace AccessibilityInsights.SharedUx.FileBug
+{
+    /// <summary>
+    /// This class sets several registry values in order to ensure that the hosted WebBrowser control renders
+    /// documents in standards mode. Without these changes the hosted WebBrowser is unable to properly
+    /// render some pages when filing issues
+    /// https://docs.microsoft.com/en-us/previous-versions/windows/internet-explorer/ie-developer/general-info/ee330730%28v%3dvs.85%29#browser-emulation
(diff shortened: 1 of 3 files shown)


src/AccessibilityInsights.SharedUx/FileBug/IEBrowserEmulation.cs
```diff
@@ -0,0 +1,71 @@
...
+    /// <summary>
+    /// This class sets several registry values in order to ensure that the hosted WebBrowser control renders
+    /// documents in standards mode. Without these changes the hosted WebBrowser is unable to properly
+    /// render some pages when filing issues
+    /// https://docs.microsoft.com/en-us/previous-versions/windows/internet-explorer/ie-developer/general-info/ee330730%28v%3dvs.85%29#browser-emulation
+    /// http://msdn.microsoft.com/en-us/library/ee330720(v=vs.85).aspx
+    /// </summary>
+    public static class IEBrowserEmulation
+    {
+        public static void SetFeatureControls()
```
neocln: Don't we need to clean up after all?

karanbirsingh: These feature controls are per process so we are OK

karanbirsingh:
Holding this PR until the ADO refactoring changes come in. Will need to integrate into the extension & see if we can find a good place to do this (has to occur periodically to check for IE changes, but perhaps not every time a bug is filed)

PR #230, opened 2019-03-27
Support GitHub Issue Filing for AI-Win

AhmedAbdoOrtiga:
#### PR checklist

- [ ] Includes UI changes?

#### Describe the change
- Refactor AI-Win code to support multiple bug filing extensions.
- Support GitHub Issue Filing for AI-Win.
- Refactor Azure DevOps bug filing.

**Test cases (you can try more):**
1. User has not configured any reporter - startup.
2. User has configured any reporter - startup. Make sure the logo is updated when the user is signed in.
3. No issue reporters - play around.
4. Use the connection logo to get to the connections tab and play around. Configure a connection, save and exit. Use the settings button to go to the connections tab and make sure that you see the previously saved connection and the corresponding UI from the extension.
5. Reverse of 4. Set from Settings check from the logo.
6. Save with only application settings changes.
7. Save with only connection settings changes.
8. Save with both types of changes.
9. Switch  between the bug reporting options without saving and make sure that there is no impact.
10. Make sure that logo is updated when the issue reporting option changes (when saved).
11. File bug using all possible ways.     From Live inspect Hierarchy tree     From Automated checks Hierarchy tree (For issue and without issue)     From the how to fix ((For issue and without issue).

**Screenshots**
- No extension is configured.
image
image
image
image

- Extension is configured but none is selected
image
image
image
image
image
image
image
image

- Multiple extensions are configured
image
image
image
image

- Restore configuration
image
image

- Single Issue
image

- No issue
image


src/AccessibilityInsights.SharedUx/Controls/SettingsTabs/ConnectionControl.xaml
```diff
@@ -5,183 +5,33 @@
...
+                <RowDefinition Height="Auto"/>
+                <RowDefinition Height="Auto"/>
+            </Grid.RowDefinitions>
+            <Label x:Name="lblIssueFiling" Grid.Row="0" Grid.Column="0" Margin="-4 15 0 7" VerticalAlignment="Center" FontSize="{StaticResource ConstXLTextSize}"
+                Content="{x:Static Properties:Resources.lblIssueFiling}" Focusable="True" IsTabStop="True"/>
+            <TextBlock x:Name="connectionInstr" Grid.Row="1" MaxWidth="336" HorizontalAlignment="Left" FontSize="{StaticResource ConstStandardTextSize}"
+                        TextWrapping="Wrap" Margin="0 0 0 15" Foreground="{DynamicResource ResourceKey=TextBrushGray}" Focusable="True">
+                <Run Text="{x:Static Properties:Resources.connectionInstrText}"/>
+               <TextBlock>
+                    <Hyperlink Name="hlLink" FocusVisualStyle="{DynamicResource {x:Static SystemParameters.FocusVisualStyleKey}}" TextDecorations="None" RequestNavigate="hlLink_RequestNavigate" NavigateUri="https://www.bla.com">
```
AhmedAbdoOrtiga: @risanaka need your input to update this link

src/AccessibilityInsights.CommonUxComponents/Controls/CustomControlOverridingAutomationPeer.cs
```diff
@@ -0,0 +1,69 @@
...
+    /// </summary>
+    public class CustomControlOverridingAutomationPeer: UserControlAutomationPeer
+    {
+        /// <summary>
+        /// Value for custom IsControlElement
+        /// </summary>
+        private bool IsControlElem;
+
+        /// <summary>
+        /// Value for custom IsContentElem
```
DaveTryon:
```suggestion
        /// Value for custom IsContentElement
```

karanbirsingh:
Build is failing due to fakes issue - in [this commit](https://github.com/Microsoft/accessibility-insights-windows/pull/230/commits/a30b627a13e4de1b7222dd54472b56f88222a020) we lost some FAKES_SUPPORTED conditions which may be contributing to it

jalkire:
Just flagging a change caused by this PR: we will now store the URL of filed bugs for ADO with the test file along with the bug number. Previously, we stored the bug number and used the user's current ADO account to generate a URL. 

Pros to old approach:
- Safe against changes to ADO URL schema

Pros to new approach:
- If you're logged into a different account than the account containing the bug, you will not be taken to an arbitrary, different bug with no warning (which is the current behavior)

@DaveTryon @neocln @karanbirsingh

src/AccessibilityInsights.CommonUxComponents/Controls/DualFabricIconControl.xaml.cs
```diff
@@ -0,0 +1,201 @@
...
+            set { SetValue(IconSizeFrontProperty, value); }
+        }
+
+        public static void OnIconSizeFrontChanged(DependencyObject o, DependencyPropertyChangedEventArgs e)
+        {
+            DualFabricIconControl sender = o as DualFabricIconControl;
+
+            if (sender != null)
+            {
+
```
DaveTryon: Is there really lots of extra newlines here, or is it just the web review tool?

src/AccessibilityInsights.CommonUxComponents/Controls/DualFabricIconControl.xaml.cs
```diff
@@ -0,0 +1,201 @@
...
+            set { SetValue(ForegroundProperty, value); }
+        }
+
+        public static void OnForegroundChanged(DependencyObject o, DependencyPropertyChangedEventArgs e)
+        {
+            DualFabricIconControl sender = o as DualFabricIconControl;
+
+            if (sender != null)
+
```
DaveTryon: Remove this newline?

src/AccessibilityInsights.CommonUxComponents/Controls/FabricIconControl.xaml.cs
```diff
@@ -0,0 +1,3278 @@
...
+
+        public static void OnFontSizeChanged(DependencyObject o, DependencyPropertyChangedEventArgs e)
+        {
+            FabricIconControl sender = o as FabricIconControl;
+
+            if (sender != null)
+            {
+                if (sender.GlyphSize != GlyphContext.Custom)
+                {
+                    //throw new NotSupportedException("GlyphSize must be set to 'Custom' in order to directly set FontSize.");
```
DaveTryon: This looks incomplete

src/AccessibilityInsights.Extensions.AzureDevOps/AzureBoardsIssueReporting.cs
```diff
@@ -0,0 +1,83 @@
...
+
+                try
+                {
+                    if (!FileIssueHelpers.AttachIssueData(issueInfo, newIssueId, issueId.Value).Result)
+                    {
+                        MessageDialog.Show(Properties.Resources.There_was_an_error_identifying_the_created_issue_This_may_occur_if_the_ID_used_to_create_the_issue_is_removed_from_its_Azure_DevOps_description_Attachments_have_not_been_uploaded);
+                    }
+                    return new IssueResult() { DisplayText = newIssueId, IssueLink = null };
+                }
+                catch (Exception)
```
DaveTryon: I'd suggest calling ReportException here to capture the failure in our telemetry stream

src/AccessibilityInsights.Extensions.AzureDevOps/ConnectionInfo.cs
```diff
@@ -21,9 +20,9 @@ public class ConnectionInfo : IConnectionInfo
         [JsonIgnore]
         public Uri ServerUri => ConcreteServerUri;
         [JsonIgnore]
-        public IProject Project => ConcreteProject;
+        public TeamProject Project => ConcreteProject;
```
DaveTryon: Now that there's no longer an interface, I suspect we could simplify this (lose ConcreteProject and ConcreteTeam, have just Project and Team, and remove the [JsonIgnore] tags). That would break saved connection configs from existing users, but since we'll probably break them already, this doesn't seem so terrible. Doesn't necessarily need to be part of this PR, but if we release the build with the Concrete names in the serialized objects, we're sort of stuck with them for the long term

src/AccessibilityInsights.Extensions.AzureDevOps/ConnectionInfo.cs
```diff
@@ -128,7 +127,7 @@ public void SetLastUsage(DateTime? updateTime = null)
             ConcreteLastUsage = newTime;
         }
 
-        public bool Equals(IConnectionInfo other)
+        public bool Equals(ConnectionInfo other)
```
DaveTryon: I forget--are we removing the definitions for  IConnectionInfo, ITeam, and IProject, since they're no longer used in the interface?

src/AccessibilityInsights.Extensions.AzureDevOps/Models/AzureDevOpsEntity.cs
```diff
@@ -1,14 +1,13 @@
...
-using AccessibilityInsights.Extensions.Interfaces.BugReporting;
 using System;
 
 namespace AccessibilityInsights.Extensions.AzureDevOps.Models
 {
     /// <summary>
     /// Represents an entity with a name and id
     /// </summary>
-    public class AzureDevOpsEntity : IEntity
+    public class AzureDevOpsEntity
```
DaveTryon: Are removing the IEntity interface?

src/AccessibilityInsights.Extensions.GitHub/LinkValidator.cs
```diff
@@ -0,0 +1,80 @@
...
+            }
+
+            string userName = parts[0];
+            string repoName = parts[1];
+            if (!CheckUserNameLength(userName) || !CheckUserNameAtLeastOneChar(userName) || !CheckRepoNameLength(repoName) || !CheckRepoNameSpecialCases(repoName))
+            {
+                return false;
+            }
+
+            return true;
```
DaveTryon: If this is supposed to be a fully formed Uri, then it might make sense to verify that it's not going to cause an Exception when we use it.
```suggestion
            return Uri.IsWellFormedString(link, UriKind.Absolute);
```

src/AccessibilityInsights.Extensions.GitHub/PlaceHolderTextBox.cs
```diff
@@ -0,0 +1,50 @@
...
+    public class PlaceHolderTextBox : TextBox
+    {
+        public string PlaceHolder { get; }
+        public SolidColorBrush BlackBrush { get; }
+        public SolidColorBrush GrayBrush { get; }
+
+        public PlaceHolderTextBox()
+        {
+            PlaceHolder = Properties.Resources.PlaceHolder;
+            BlackBrush = Application.Current.Resources["TextBrush"] as SolidColorBrush;
```
DaveTryon: Will this be responsive to color theme changes that occur while the TextBox is displayed? It feels like it would retain the pre-change colors ☹️

src/AccessibilityInsights.Extensions.GitHubUnitTests/Extensions.GitHubUnitTests.csproj
```diff
@@ -0,0 +1,96 @@
...
+  <ItemGroup>
+    <ProjectReference Include="..\AccessibilityInsights.Extensions.GitHub\Extensions.GitHub.csproj">
+      <Project>{f421aafd-d85f-49b8-b95e-ca17718c38dd}</Project>
+      <Name>Extensions.GitHub</Name>
+    </ProjectReference>
+    <ProjectReference Include="..\AccessibilityInsights.Extensions\Extensions.csproj">
+      <Project>{EAA85D0D-712D-4D85-A246-D3C699C6C602}</Project>
+      <Name>Extensions</Name>
+    </ProjectReference>
+  </ItemGroup>
```
DaveTryon: According to NewProject.md, we should enable delay signing for this project. I don't see it in the csproj file

src/AccessibilityInsights.Extensions/Interfaces/IssueReporting/IIssueResult.cs
```diff
@@ -0,0 +1,16 @@
...
+using System;
+
+namespace AccessibilityInsights.Extensions.Interfaces.IssueReporting
+{
+    /// <summary>
+    /// Interface for the details displayed on the new issue button after filing is complete
+    /// </summary>
+    public interface IIssueResult
+    {
+        string DisplayText { get; }
```
DaveTryon: Some meaningful /// comments would be appropriate for the interface members

src/AccessibilityInsights.Extensions/Interfaces/IssueReporting/IssueInformation.cs
```diff
@@ -1,90 +1,95 @@
...
     /// <summary>
-    /// Class to communicate bug information across the extension interface
+    /// Class to communicate Issue information across the extension interface
     /// </summary>
-    public class BugInformation
+    public class IssueInformation
     {
         /// <summary>
-        /// The title of the window where a bug is located
+        /// The title of the window where a Issue is located
```
DaveTryon:
```suggestion
        /// The title of the window where an Issue is located
```

src/AccessibilityInsights.Extensions/Interfaces/IssueReporting/IssueType.cs
```diff
@@ -1,11 +1,9 @@
...
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
-namespace AccessibilityInsights.Extensions.Interfaces.BugReporting
+namespace AccessibilityInsights.Extensions.Interfaces.IssueReporting
 {
-    public enum  BugType
+    public enum  IssueType
     {
-        Unknown,
-        ColorContrast,
         SingleFailure,
```
DaveTryon: Please keep **Unknown** as the first item in the enum. The code should never set it, but it serves as a guard against places where we fail to set the enum

src/AccessibilityInsights.SharedUx/Controls/ScannerResultControl.xaml.cs
```diff
@@ -275,65 +276,46 @@ private void lvDetails_SizeChanged(object sender, SizeChangedEventArgs e)
...
-                            vm.LoadingVisibility = Visibility.Collapsed;
-                        }
-                        catch (Exception)
-                        {
-                            vm.LoadingVisibility = Visibility.Collapsed;
-                        }
+                        vm.IssueDisplayText = issueResult.DisplayText;
+                        vm.IssueLink = issueResult.IssueLink;
                     }
+                    File.Delete(issueInformation.TestFileName);
```
DaveTryon: This cleanup gets skipped if an exception gets thrown. Should it be in a finally block to ensure proper cleanup?

src/AccessibilityInsights.SharedUx/Settings/ConfigurationConstants.cs
```diff
@@ -71,6 +69,7 @@ public partial class ConfigurationModel
         private const string keyTestReportPath = "TestReportPath";
         private const string keyTreeViewMode = "TreeViewMode";
         private const string keyVersion = "Version";
-        private const string keyZoomLevel = "ZoomLevel";
+        private const string keyIssueReporting= "IssueReportingGuid";
+        private const string keyIssueReporterSerializedConfig = "IssueReporterSerializedConfig";
```
DaveTryon: Does this mean that if I want to use both GitHub and ADO, I need to log in every time I switch? I'd sure like to be able to persist within each provider...

src/AccessibilityInsights/MainWindow.xaml.cs
```diff
@@ -701,18 +703,46 @@ private void btnAccountConfig_Click(object sender, RoutedEventArgs e)
         }
 
         /// <summary>
-        /// Initialize server integration and try logging in implicitly 
-        /// to the saved connection in the configuration if it exists.
+        /// Set saved issue reporter and try restoring configuration if it exists.
         /// </summary>
-        private void ConnectToSavedServerConnection(Action callback = null)
+        private async void RestoreConfigurationAsync()
```
DaveTryon: Won't an Exception in the async method crash the app?

src/AccessibilityInsights.Extensions.AzureDevOps/FileIssue/FileIssueHelpers.cs
```diff
@@ -0,0 +1,506 @@
...
+            return AzureDevOps.PopulateUserProfile();
+        }
+
+        public static Task<int?> ReplaceIssueDescriptionAsync(string description, int issueId)
+        {
+            return AzureDevOps.ReplaceIssueDescription(description, issueId);
+        }
+
+        /// <summary>
+        /// Extract the template name from a IssueInformation object
```
DaveTryon:
```suggestion
        /// Extract the template name from an IssueInformation object
```

src/AccessibilityInsights.Extensions.AzureDevOps/FileIssue/FileIssueHelpers.cs
```diff
@@ -0,0 +1,506 @@
...
+                fieldPairs.Add(AzureDevOpsField.AreaPath, areaPathTask.Result);
+            }
+            if (iterationPathTask.Result != null)
+            {
+                fieldPairs.Add(AzureDevOpsField.IterationPath, iterationPathTask.Result);
+            }
+        }
+
+        /// <summary>
+        /// Generates a dictionary of AzureDevOpsField/string pairs for creating a issue in AzureDevOps
```
DaveTryon:
```suggestion
        /// Generates a dictionary of AzureDevOpsField/string pairs for creating an issue in AzureDevOps
```

src/AccessibilityInsights.Extensions.AzureDevOps/FileIssue/FileIssueHelpers.cs
```diff
@@ -0,0 +1,506 @@
...
+            // Template & add description field
+            string issueDescTemplate = string.Concat(File.ReadAllLines(Path.Combine(Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location), $"IssueTemplates\\{templateName}.html")));
+            templatedAzureDevOpsFieldPairs.Add(AzureDevOpsField.ReproSteps, PopulateIssueTemplateString(issueDescTemplate, issueFieldPairs));
+
+            return templatedAzureDevOpsFieldPairs;
+        }
+
+        /// <summary>
+        /// Returns a dictionary that maps from IssueField names to string values that should be used
+        /// to file a issue. Substitutions can be defined in a separate json file.
```
DaveTryon:
```suggestion
        /// to file an issue. Substitutions can be defined in a separate json file.
```

src/AccessibilityInsights.CommonUxComponents/Controls/FabricIconControl.xaml
```diff
@@ -0,0 +1,14 @@
+<!-- Copyright (c) Microsoft. All rights reserved.
+     Licensed under the MIT license. See LICENSE file in the project root for full license information.-->
+<UserControl x:Class="AccessibilityInsights.CommonUxComponents.Controls.FabricIconControl"
```
DaveTryon: There are still many XAML files that reference AccessibliityInsights.DesktopUI.Controls, presumably to use the *FabricIconControl. Will these be cleaned up in this PR or in a separate change?

src/AccessibilityInsights.Extensions.GitHub/LinkValidator.cs
AhmedAbdoOrtiga: I don't think it's necessary, we only check for regex

src/AccessibilityInsights.Extensions.GitHub/PlaceHolderTextBox.cs
AhmedAbdoOrtiga: it doesn't, we already have this issue in other places and we still didn't fix it, I'm not sure if we can do something about that

src/AccessibilityInsights.CommonUxComponents/Controls/FabricIconControl.xaml
AhmedAbdoOrtiga: these should be cleaned up in a separate change
@jalkire

src/AccessibilityInsights.Extensions.AzureDevOps/AzureBoardsIssueReporting.cs
jalkire: Can we access telemetry from extensions?

src/AccessibilityInsights.Extensions.AzureDevOps/ConnectionInfo.cs
jalkire: Yes

src/AccessibilityInsights.Extensions.AzureDevOps/Models/AzureDevOpsEntity.cs
jalkire: Yes

src/AccessibilityInsights.SharedUx/Settings/ConfigurationConstants.cs
jalkire: No, data will be saved for all extensions simultaneously. You're in luck.

src/AccessibilityInsights.Extensions/Interfaces/IssueReporting/IssueType.cs
AhmedAbdoOrtiga: it isn't used in master.

src/AccessibilityInsights.Extensions.AzureDevOps/AzureBoardsIssueReporting.cs
AhmedAbdoOrtiga: use ExceptionExtensions.ReportException(e) in AccessibilityInsights.Extensions.Helpers

src/AccessibilityInsights.SharedUx/Controls/ScannerResultControl.xaml.cs
AhmedAbdoOrtiga: Nope, if you mean exception in FileIssueAction.FileIssueAsync, it's already caught in the same function so still will hit this line

src/AccessibilityInsights.CommonUxComponents/Controls/FabricIconControl.xaml.cs
AhmedAbdoOrtiga: it was just copied from the DesktopUI so I believe it should be fine
@jalkire, any thought?

src/AccessibilityInsights.Actions/Actions/PrivacyExtensions.cs
```diff
@@ -57,7 +57,7 @@ internal static A11yElement CreateScrubbedCopyWithoutRelationships(this A11yElem
         {
             A11yElement clone = new A11yElement
             {
-                BugId = element.BugId,
+                IssueDisplayText = element.IssueDisplayText,
```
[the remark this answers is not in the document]
jalkire: The old bug number data will not be available, but we are able to load the file successfully apart from that.

src/AccessibilityInsights.Extensions.AzureDevOps/ConnectionInfo.cs
jalkire: Agreed, removed concrete fields.

src/AccessibilityInsights/MainWindow.xaml.cs
jalkire: Yes, but that's true with a lot of code in MainWindow.xaml.cs. Is there a particular part of this you're worried about? I'm thinking maybe 
`await issueReporter.RestoreConfigurationAsync(serializedConfig).ConfigureAwait(true);`
should be try-caught?

src/AccessibilityInsights.sln
```diff
@@ -145,7 +140,16 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Extensions.TelemetryTests",
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Extensions.AutoUpdate", "AccessibilityInsights.Extensions.AutoUpdate\Extensions.AutoUpdate.csproj", "{BBDB2193-7738-4E6D-8061-1DAA64FCCBB0}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "UnitTestSharedLibrary", "AccessibilityInsights.UnitTestSharedLibrary\UnitTestSharedLibrary.csproj", "{01DC8363-73BC-47FA-BD23-12AC45618E1D}"
```
DaveTryon: This shouldn't be deleted

src/AccessibilityInsights/MainWindow.xaml.cs
DaveTryon: Exceptions gets handled if the caller uses await or .ConfigureAwait or even .Result. Exceptions are not handled if you just call an async method and do nothing with it. Code like this in the AutoUpdate modules include their own try/catch block inside the async method so that the caller is free to do whatever is necessary. 

I don't have a specific line of code that concerns me, but since any exception will likely kill the process, it seems worth thinking about

src/AccessibilityInsights.Extensions/Interfaces/IssueReporting/IssueType.cs
DaveTryon: That's correct. It exists to differentiate between intentionally set values and default (i.e., not set by the code owner) values. Common pattern

src/AccessibilityInsights.SharedUx/Controls/ScannerResultControl.xaml.cs
DaveTryon: Not what I meant. Suppose that an exception occurs inside FileIssueAction.AttachIssueData, but after the temp file has been created. The Exception will cause the File.Delete to get skipped, so the temp file is orphaned. If lines 306-313 were wrapped in a try block, and the File.Delete put into a finally block, then you always clean up successfully. One note: File.Delete throws if the file doesn't exist, so the finally block would be `if(File.Exists()) File.Delete;`

src/AccessibilityInsights.Extensions.AzureDevOpsTests/AzureBoardsIssueReportingTests.cs
```diff
@@ -0,0 +1,268 @@
...
+            Assert.AreEqual(ruleSourceContent, modifiedRuleSource);
+            Assert.IsNull(issueFieldPairs[IssueField.ProcessName]);
+            Assert.IsNull(issueFieldPairs[IssueField.Glimpse]);
+            Assert.IsNull(issueFieldPairs[IssueField.TestMessages]);
+        }
+
+#if FAKES_SUPPORTED
+        [TestMethod]
+        [Timeout(1000)]
+        public void CreateIssuePreviewAsync_TeamNameIsNull_ChainsThroughCorrectly()
```
DaveTryon: This tests fails 100% of the time on my machine

src/AccessibilityInsights.Extensions.AzureDevOpsTests/AzureBoardsIssueReportingTests.cs
```diff
@@ -0,0 +1,268 @@
...
+
+                Assert.AreEqual(expectedUri, actualUri);
+                Assert.AreEqual(expectedProjectName, actualProjectName);
+                Assert.IsNull(actualTeamName);
+            }
+        }
+
+        [TestMethod]
+        [Timeout(1000)]
+        public void CreateIssuePreviewAsync_TeamNameIsNotNull_ChainsThroughCorrectly()
```
DaveTryon: This test fails 100% of the time on my machine

src/AccessibilityInsights.Extensions.AzureDevOpsTests/AzureBoardsIssueReportingTests.cs
jalkire: Tests should work now. Not sure about the Run all issue...

src/AccessibilityInsights.Extensions.AzureDevOpsTests/AzureBoardsIssueReportingTests.cs
jalkire: Tests should work now. Not sure about the Run all issue...

src/AccessibilityInsights.Extensions.AzureDevOpsTests/AzureBoardsIssueReportingTests.cs
DaveTryon: I've pushed a commit that unblocks the unit tests forme

src/AccessibilityInsights.Extensions.AzureDevOpsTests/AzureBoardsIssueReportingTests.cs
DaveTryon: Unit tests pass. I've pushed a commit that unblocks the unit tests for me

src/AccessibilityInsights.SharedUx/Controls/ScannerResultControl.xaml.cs
```diff
@@ -303,15 +303,30 @@ private void btnFileBug_Click(object sender, RoutedEventArgs e)
...
+                        }
                     }
-                    File.Delete(issueInformation.TestFileName);
+                    catch (Exception ex)
+                    {
+                        ex.ReportException();
+                    }
+                    finally
+                    {
+                        if(issueInformation != null)
```
DaveTryon: This should also check for the file's existence:

if (issueInformation?.TestFilename != null && File.Exists(issueInformation.TestFilename))

src/AccessibilityInsights.SharedUx/Controls/ScannerResultControl.xaml.cs
```diff
@@ -303,15 +303,30 @@ private void btnFileBug_Click(object sender, RoutedEventArgs e)
...
-                    IssueInformation issueInformation = vm.GetIssueInformation();
-                    FileIssueAction.AttachIssueData(issueInformation, this.EcId, vm.Element.BoundingRectangle, vm.Element.UniqueId);
-                    IIssueResult issueResult = FileIssueAction.FileIssueAsync(issueInformation);
-                    if (issueResult != null)
+                    IssueInformation issueInformation = null ;
+                    try
                     {
-                        vm.IssueDisplayText = issueResult.DisplayText;
-                        vm.IssueLink = issueResult.IssueLink;
+                        issueInformation = vm.GetIssueInformation();
```
DaveTryon: can issueInformation ever be null?

src/AccessibilityInsights.Extensions.AzureDevOps/AzureBoardsIssueReporting.cs
```diff
@@ -0,0 +1,83 @@
...
+                // Check whether issue was filed once dialog closed & process accordingly
+                if (!issueId.HasValue) return null;
+
+                try
+                {
+                    if (!FileIssueHelpers.AttachIssueData(issueInfo, newIssueId, issueId.Value).Result)
+                    {
+                        MessageDialog.Show(Properties.Resources.There_was_an_error_identifying_the_created_issue_This_may_occur_if_the_ID_used_to_create_the_issue_is_removed_from_its_Azure_DevOps_description_Attachments_have_not_been_uploaded);
+                    }
+                    return new IssueResult() { DisplayText = newIssueId, IssueLink = null };
```
karanbirsingh: why is the returned IssueLink null?

src/AccessibilityInsights.Extensions.AzureDevOps/AzureDevOpsIntegration.cs
```diff
@@ -432,26 +435,47 @@ public Uri GetTeamProjectUrl(string projectName, string teamName)
             return builder.Uri;
         }
 
+
+        internal async Task HandleLoginAsync(CredentialPromptType showDialog=CredentialPromptType.DoNotPrompt, Uri serverUri = null)
```
karanbirsingh: this might be easier to read if is refactored to:
```
serverUri = serverUri ?? Configuration.SavedConnection.ServerUri;
if (serverUri != null) {
   try {
      await...
   } catch () { … }
}
```

src/AccessibilityInsights.Extensions.AzureDevOps/Models/IssueResult.cs
```diff
@@ -0,0 +1,23 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using AccessibilityInsights.Extensions.Interfaces.IssueReporting;
+using System;
+
+namespace AccessibilityInsights.Extensions.AzureDevOps.Models
+{
+    /// <summary>
+    /// Contains data to be displayed in AI-Win after an issue is filed
+    /// </summary>
+    public class IssueResult : IIssueResult
```
karanbirsingh: something like this would be helpful to encapsulate the same entities in Core.Results.RuleResult.cs but I'm not sure if it can be shared

src/AccessibilityInsights.Extensions.GitHub/ConfigurationModelControl.xaml.cs
```diff
@@ -0,0 +1,108 @@
...
+                canSave = false;
+            }
+            UpdateSaveButton();
+        }
+
+        private bool canSave = false;
+        /// <summary>
+        /// Can the save button be clicked
+        /// </summary>
+        public override bool CanSave => canSave;
```
karanbirsingh: I remember discussing moving `UpdateSaveButton` into the setter of the `CanSave` property so that we don't need to keep `canSave` changes and `UpdateSaveButton()` calls together explicitly.  Did that approach run into issues?

src/AccessibilityInsights.SharedUx/Controls/ScannerResultControl.xaml.cs
jalkire: It should never be null

src/AccessibilityInsights.Extensions.GitHub/LinkValidator.cs
```diff
@@ -0,0 +1,80 @@
...
+
+namespace AccessibilityInsights.Extensions.GitHub
+{
+    /// <summary>
+    /// GitHub Repo Link Format Validator
+    /// </summary>
+    public static class LinkValidator
+    {
+        private static readonly string GitHubLink = Properties.Resources.GitHubLink;
+        private static readonly string AlphaNumricPattern = Properties.Resources.AlphaNumricPattern;
```
karanbirsingh: should be 'numeric'

src/AccessibilityInsights.Extensions.GitHub/LinkValidator.cs
```diff
@@ -0,0 +1,80 @@
...
+
+        public static bool IsValidGitHubRepoLink(string Link)
+        {
+            Link = Link.Replace(@"\", "/").Trim(' ').TrimEnd('/');
+            string UserNamePattern = string.Format(CultureInfo.InvariantCulture, Properties.Resources.UserNamePattern, AlphaNumricPattern);
+            string RepoNamePattern = string.Format(CultureInfo.InvariantCulture, Properties.Resources.RepoNamePattern, AlphaNumricPattern);
+            string LinkPatttern = string.Format(CultureInfo.InvariantCulture, Properties.Resources.LinkPatttern, GitHubLink, UserNamePattern, RepoNamePattern);
```
karanbirsingh: pattern has an extra t

src/AccessibilityInsights.Extensions.GitHub/ConfigurationModelControl.xaml.cs
AhmedAbdoOrtiga: I replied to the comment in the PR, the thing was CanSave has no setter by design, so we can't apply that approach

karanbirsingh: Got it, I see - I am not sure how or if we can make it private, will take a look afterwards as a separate issue

src/AccessibilityInsights.Extensions.AzureDevOps/AzureDevOpsIntegration.cs
jalkire: Good idea. Also mixed in some of @RobGallo's nesting-avoidance approach

src/AccessibilityInsights.Extensions.AzureDevOps/AzureBoardsIssueReporting.cs
jalkire: Oops

PR #236, opened 2019-03-29
Add appropriate feature control registry keys to improve WebBrowser rendering

karanbirsingh:
We have occasionally noticed the WebBrowser used in the AzureDevops extension is unable to render the page and instead shows a white screen (even though viewing the source shows proper HTML). This PR applies a potential fix via feature controls. IE provides several feature controls in order to enhance the way the WebBrowser behaves. Earlier we set the Browser Emulation registry key to a high number [99999] which doesn't seem accurate - according to the docs and suggested fix, this key should be set to the values defined [here](https://docs.microsoft.com/en-us/previous-versions/windows/internet-explorer/ie-developer/general-info/ee330730(v%3dvs.85)#browser-emulation). This affects what 'mode' web pages are rendered in (read more [here](https://docs.microsoft.com/en-us/previous-versions/windows/internet-explorer/ie-developer/general-info/ee330730%28v%3dvs.85%29)). In this PR the value is chosen based on the installed version of IE. This change was tested on a few machines from our team and those with the white screen issue reported successful rendering afterwards. There is one case where the issue still exists - will have to troubleshoot on their box. We are exploring other options to fix or replace the WebBrowser; in the meantime this fix should improve the situation for some users.

### DIFF
--- a/src/AccessibilityInsights.Extensions.AzureDevOps/Extensions.AzureDevOps.csproj
+++ b/src/AccessibilityInsights.Extensions.AzureDevOps/Extensions.AzureDevOps.csproj
@@ -142,6 +142,7 @@
     <Compile Include="ConfigurationControl.xaml.cs">
       <DependentUpon>ConfigurationControl.xaml</DependentUpon>
     </Compile>
+    <Compile Include="FileIssue\IEBrowserEmulation.cs" />
     <Compile Include="IssueInformationHelpers.cs" />
     <Compile Include="Enums\IssueField.cs" />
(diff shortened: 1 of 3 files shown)


DaveTryon:
> using AccessibilityInsights.Extensions.Helpers;

No License header - PR build will fail

---
Refers to: src/AccessibilityInsights.Extensions.AzureDevOps/FileIssue/IEBrowserEmulation.cs:1 in 5fcd574. [](commit_id = 5fcd57430958d610ebb64120ebde8f3f53ef8c41, deletion_comment = False)

PR #237, opened 2019-03-29
Remove > characters before commands in readme

karanbirsingh:
I find myself copy-pasting these commands occasionally and the > characters have to be removed. It would be easier if I could copy them at once to run.


docs/GitBranchSetup.md
```diff
@@ -32,21 +32,21 @@ primary **accessibility-insights-windows** repository.
...
-> git remote add upstream https://github.com/Microsoft/accessibility-insights-windows
+git remote add upstream https://github.com/Microsoft/accessibility-insights-windows
 ```
 
 -   Now, running `git remote -v` should show the upstream repository also
 
 ```
-> git remote -v
+git remote -v
 origin  https://github.com/ada-cat/accessibility-insights-windows (fetch)
```
karanbirsingh: I could understand if someone feels strongly that the command here should be distinct from the output, suggestions are welcome

PR #243, opened 2019-04-02
Use CommonUxComponents controls in SharedUx and AccessibilityInsights

jalkire:
#### Describe the change
Previously, MessageDialog was duplicated into the new CommonUxComponents project. This PR follows that up by removing the original MessageDialog from SharedUx and updating all references accordingly. There should be no user impact. It also fixes the dialog's button styling.

FabricIconControl and DualFabricIconControl were duplicated into CommonUxComponents. This PR updates all usage of these controls in SharedUx and AccessibilityInsights. However, the original controls still reside in DesktopUI as well. They will be removed when (if) the highlighters are moved into SharedUx/out of the runtime.

#### PR checklist

- [ ] Includes UI changes?


src/AccessibilityInsights.CommonUxComponents/Dialogs/MessageDialog.xaml.cs
```diff
@@ -25,6 +25,7 @@ public MessageDialog()
         {
             var dlg = new MessageDialog() { Message = message };
             dlg.Owner = Application.Current.MainWindow;
+            dlg.Resources = Application.Current.MainWindow.Resources;
```
neocln: Is there any implication from it? such as a certain resource is expected in App level resources?

jalkire: This is here to give the dialog access to the resources in Styles.xaml, specifically BtnBlue. In general, it gives the dialog access to whatever resources MainWindow has.

neocln: If so, don't we need to consider splitting Styles.xaml into two parts(common and AI-Win specific)?

jalkire: Consider, yes. I would call that out of scope for this PR. There are more implications to deal with. Currently, SharedUx still owns theme, including styles, brushes, and font size. I'm not positive about separating those things out.

src/AccessibilityInsights.CommonUxComponents/Controls/FabricIconControl.xaml.cs
```diff
@@ -31,7 +31,7 @@ public FabricIconControl()
         /// Name of icon to be drawn
         /// </summary>
         #region GlyphName (Dependency Property)
-
```
DaveTryon: Remove line 33? (extra line and trailing spaces)

PR #253, opened 2019-04-03
Merge current Rings of Validation work to master

DaveTryon:
#### Describe the change
Integrate the current Rings of Validation work into master. These changes are almost all "under the hood" at this time, and the only UI is in terms of messaging on some failure cases
- Configuration includes a **ReleaseChannel** key to store the channel (defaults to Production)
- Setup-related code has been centralized into a new assembly named `AccessibilityInsights.SetupLibrary`
- Upgrades (and channel changes) are now handled by a separate process named `AccessibilityInsights.VersionSwitcher`. This process gets installed as part of the MSI but does nothing when AIWin is running. When a version change is required (due to upgrade or a channel change), this process gets copied to the temp folder, then invoked from there. The VersionSwitcher app then does the following:
  - Prompts for admin privileges
  - Downloads the MSI for the new version
  - Ensures that the MSI is properly signed
  - Uninstalls the existing version of AIWin
  - Installs the new version of AIWin
  - Updates the config file to the new ReleaseChannel (doesn't apply to upgrades)
  - Launches the new version of AIWin (this launch is without elevated privileges)
- The file-based AutoUpdate assembly is no longer being built, and the code will be removed (but probably not as part of this PR)

With the AxeWindows refactorings going on, Min suggested that we merge this sooner, rather than later. Bi and I have already tested the upgrade scenario, so existing functionality shouldn't be impacted.

Signing issues have been validated by successfully building in the signed build definition.

#### PR checklist

- [ ] Includes UI changes?


src/MSI/Product.wxs
```diff
@@ -157,6 +159,17 @@ Licensed under the MIT license. See LICENSE file in the project root for full li
                  </Directory>
                </Directory>
             </Directory>
+
+            <Directory Id="VersionSwitcher" Name="VersionSwitcher">
+              <Component Id="VersionSwitcherComp" Guid="1AAD6099-09E1-4E8F-A28B-57806F4A29DF">
+                <File Source="..\AccessibilityInsights.VersionSwitcher\bin\Release\AccessibilityInsights.VersionSwitcher.exe" Id="version_switcher"/>
+                <File Source="..\AccessibilityInsights.VersionSwitcher\bin\Release\AccessibilityInsights.VersionSwitcher.exe.config" Id="version_switcher_config"/>
+                <File Source="..\AccessibilityInsights.VersionSwitcher\bin\Release\AccessibilityInsights.SetupLibrary.dll" Id ="version_switcher_extension"/>
+                <File Source="..\AccessibilityInsights.VersionSwitcher\bin\Release\AccessibilityInsights.Win32.dll" Id ="version_switcher_win32"/>
```
DaveTryon: This also needs to include NewtonSoft

src/AccessibilityInsights.Extensions.GitHub/Properties/AssemblyInfo.cs
```diff
@@ -10,7 +10,7 @@
 [assembly: AssemblyTitle("AccessibilityInsights.Extensions.GitHub")]
 [assembly: AssemblyDescription("")]
 [assembly: AssemblyConfiguration("")]
-[assembly: AssemblyCompany("Microsoft")]
+[assembly: AssemblyCompany("")]
```
DaveTryon: We had a few AssemblyInfo.cs files where the company was set to "Microsoft". We've been told not to do that, so I changed these as I noticed them. Happy to make this a separate PR in master if there's a strong preference to handle it that way

src/AccessibilityInsights.Extensions.GitHubAutoUpdate/AutoUpdate.cs
```diff
@@ -39,25 +35,26 @@ namespace AccessibilityInsights.Extensions.GitHubAutoUpdate
     [Export(typeof(IAutoUpdate))]
     public class AutoUpdate : IAutoUpdate
```
DaveTryon: A lot of this code has been moved into the VersionSwitcher

src/AccessibilityInsights/MainWindowHelpers/AutoUpdate.cs
```diff
@@ -193,32 +192,14 @@ private static string GetTimeSpanTelemetryString(TimeSpan? timeSpan)
...
-            {
-                case UpdateResult.DownloadFailed:
-                    return Messages.OptionalUpdateFailedOnDownload;
-                case UpdateResult.VerificationFailed:
-                    return Messages.OptionalUpdateFailedVerification;
-                case UpdateResult.Unknown:
-                default:
-                    return Messages.OptionalUpdateFailedUnknown;
-            } // switch
+            return Messages.OptionalUpdateFailedUnknown;
```
DaveTryon: The old code had different messages for mandatory and optional upgrades. The new code only tells if the version switcher failed. Version switcher doesn't yet know about the concept of a mandatory upgrade

src/AccessibilityInsights.VersionSwitcher/MainWindow.xaml.cs
```diff
@@ -0,0 +1,53 @@
...
+            try
+            {
+                InstallationEngine engine = new InstallationEngine(ProductName, SafelyGetAppInstalledPath());
+                engine.PerformInstallation();
+            }
+            catch(Exception e)
+            {
+                EventLogger.WriteErrorMessage(e.ToString());
+                ExceptionReporter.ReportException(e);
+                MessageBox.Show(e.Message, "An error occurred during install");
```
jalkire: This could be an opportunity to reuse MessageDialog from CommonUxComponents

src/AccessibilityInsights/MainWindowHelpers/ControlHelper.cs
```diff
@@ -261,11 +263,20 @@ private void UpdateTitleString()
...
         /// Set version text
-        /// Release Type - Version - UIAccess state
+        /// [(ReleaseChannel) - ]Version - UIAccess state
         /// </summary>
         private void UpdateVersionString()
         {
-            StringBuilder sb = new StringBuilder(string.Format(CultureInfo.InvariantCulture, Properties.Resources.UpdateVersionStringVer, AccessibilityInsights.Core.Misc.Utility.GetAppVersion()));
+            StringBuilder sb = new StringBuilder();
+
+            ReleaseChannel? channel = ConfigurationManager.GetDefaultInstance()?.AppConfig?.ReleaseChannel;
```
jalkire: This means that we will display the currently configured channel, which may not be the actual installed/running channel in rare cases. Is this ok for our users? @risanaka @neocln

src/AccessibilityInsights.VersionSwitcher/MainWindow.xaml.cs
DaveTryon: Maybe later, but I really don't want to try to sort out the dependencies at this time.

src/AccessibilityInsights/MainWindowHelpers/ControlHelper.cs
DaveTryon: Are you thinking of cases where someone should be configured for Insider or Canary, but some issue prevents the config from being read? That would look like Production, even though it's ahead of the current build for Production. That feels like the same issue as someone installing a Canary build directly from the server, but not actually setting the Canary channel. If we can live with the latter case, then it seems like we would also be able to live with the former case.

src/AccessibilityInsights.VersionSwitcher/InstallationEngine.cs
```diff
@@ -0,0 +1,277 @@
...
+            try
+            {
+                RemoveOldVersion();
+                InstallNewVersion(options.LocalInstallerFile);
+                success = true;
+            }
+            finally
+            {
+                if (success)
+                    transaction.Commit();
```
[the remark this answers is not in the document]
DaveTryon: I generally use braces if there are multiple lines in the bodies, or if the body is the last thing in a parent body (line 155 is the last thing in a parent body). I'll insert the braces for consistency

src/AccessibilityInsights.SetupLibrary/GitHubWrapper.cs
```diff
@@ -0,0 +1,89 @@
...
+            if (!double.TryParse(configTimeout, NumberStyles.Number, CultureInfo.InvariantCulture, out double timeoutInSeconds) || timeoutInSeconds <= 0)
+                timeoutInSeconds = defaultTimeout;
+
+            _timeout = TimeSpan.FromSeconds(timeoutInSeconds);
+        }
+
+        /// <summary>
+        /// Implements <see cref="IGitHubWrapper.LoadUriContentsIntoStream(Uri, Stream)"/>
+        /// </summary>
+        public void LoadUriContentsIntoStream(Uri uri, Stream stream)
```
karanbirsingh: is this method necessary? I'm confused how it relates to the one after it (`LoadChannelInfoIntoStream`) and whether they can be collapsed

src/AccessibilityInsights.SetupLibrary/VersionSwitcherWrapper.cs
```diff
@@ -0,0 +1,162 @@
...
+        /// <param name="newChannel">If not null, the new channel to select</param>
+        /// <returns>true if the Version Switcher process started successfully, false if not</returns>
+        private static bool DownloadAndInstall(Uri installerUri, string newChannel)
+        {
+            List<FileStream> fileLocks = new List<FileStream>();
+            try
+            {
+                string temporaryFolder = GetTemporaryVersionSwitcherFolder();
+                string installedFolder = GetInstalledVersionSwitcherFolder();
+                RemoveFolder(temporaryFolder);
```
karanbirsingh: small comment - upon first read I thought we were creating the temp folder & immediately deleting it; not sure if any renames or rearranging would help

src/AccessibilityInsights.SharedUx/Settings/ConfigurationModel.cs
```diff
@@ -413,6 +414,15 @@ public bool IsUnderElementScope
             set => SetDataValue<bool>(keyIsUnderElementScope, value);
         }
 
+        /// <summary>
+        /// The release channel configured for this client
+        /// </summary>
+        public ReleaseChannel ReleaseChannel
```
karanbirsingh: do we need to update any config version values because of this change?

src/AccessibilityInsights.SharedUxTests/Resources/ConfigSettings.json
```diff
@@ -32,6 +32,7 @@
     "IsUnderElementScope": true,
     "MouseSelectionDelayMilliSeconds": 200,
     "PlayScanningSound": false,
+    "ReleaseChannel": "Canary",
```
karanbirsingh: Is this the default / should it be production instead?

src/AccessibilityInsights.VersionSwitcher/InstallationEngine.cs
```diff
@@ -0,0 +1,281 @@
...
+                SettingsDictionary settings = FileHelpers.LoadDataFromJSON<SettingsDictionary>(configFile);
+                settings[Constants.ReleaseChannelKey] = newChannel;
+                FileHelpers.SerializeDataToJSON(settings, configFile);
+            }
+        }
+
+        /// <summary>
+        /// Locate the MSI-installed app and launch it without elevated privileges
+        /// </summary>
+        private void LaunchPostInstallApp()
```
karanbirsingh: the existing MSI usually launches itself upon installation already; what is the thinking behind this method?

src/AccessibilityInsights.SharedUx/Settings/ConfigurationModel.cs
DaveTryon: The config version is essentially a constant now that the settings are explicitly stored as a dictionary. We wanted to enable people to switch between channels (forward or backward) without having to reconfigure their settings. The config version would only need to change if some future config file moved away from the Dictionary

src/AccessibilityInsights.VersionSwitcher/InstallationEngine.cs
DaveTryon: The launch on install doesn't trigger when we install the MSI through the API. I'm not really sure why that's true, but it ultimately works to our advantage. Since we are elevated when we install the MSI, the auto-launched app would also be elevated (this is easy to confirm by starting the MSI directly from an elevated command prompt, then saving a test file to C:\). We call LaunchPostInstallApp to launch AIWin after we make the change, and we launch it indirectly through explorer.exe to run it in a non-elevated state.

src/AccessibilityInsights.SetupLibrary/GitHubWrapper.cs
```diff
@@ -0,0 +1,89 @@
...
+        {
+            GitHubClient.LoadUriContentsIntoStream(uri, stream, _timeout);
+        }
+
+        /// <summary>
+        /// Implements <see cref="IGitHubWrapper.LoadUriContentsIntoStream(Uri, Stream)(Uri, Stream)"/>
```
karanbirsingh: I think the comment should instead reference the `LoadChannelInfo...` method

src/AccessibilityInsights.SetupLibrary/GitHubWrapper.cs
```diff
@@ -0,0 +1,81 @@
...
+
+            _timeout = TimeSpan.FromSeconds(timeoutInSeconds);
+        }
+
+        /// <summary>
+        /// Implements <see cref="IGitHubWrapper.LoadChannelInfoIntoStream(string, Stream)(Uri, Stream)"/>
+        /// </summary>
+        public void LoadChannelInfoIntoStream(string releaseChannel, Stream stream)
+        {
+            GitHubClient.LoadUriContentsIntoStream(_productionConfigFileUri, stream, _timeout);
```
jalkire: releaseChannel isn't geting used here. Is that right?

PR #256, opened 2019-04-03
Remove TestConfig and Suites

karanbirsingh:
#### Describe the change
This PR removes the code and infrastructure previously used to support various test suites. This feature is no longer in use. Checked automation + test flows.


src/AccessibilityInsights.Actions/Actions/SaveAction.cs
```diff
@@ -86,7 +86,7 @@ private static void SaveSnapshotFromElement(int? focusedElementId, A11yFileMode
                 }
             }
 
-            var meta = new SnapshotMetaInfo(mode, SuiteFactory.GetRuleVersion(), focusedElementId, ec.DataContext.ScreenshotElementId, otherProperties);
+            var meta = new SnapshotMetaInfo(mode, RuleRunner.GetRuleVersion(), focusedElementId, ec.DataContext.ScreenshotElementId, otherProperties);
```
karanbirsingh: I'm not sure if we still want this rule version (has always been at 1.0). I could imagine it being useful in the future so I've left it in for now.

src/AccessibilityInsights.Desktop/UIAutomation/TreeWalkers/TreeWalkerForTest.cs
```diff
@@ -87,7 +87,8 @@ public void RefreshTreeData(TreeViewMode mode)
             // run tests
             list.AsParallel().ForAll(e =>
             {
-                SuiteFactory.RunRules(e);
+                e.ScanResults?.Items.Clear();
```
karanbirsingh: this used to be inside `SuiteFactory.RunRules`. Since `SuiteFactory` is no longer needed, I've pulled out the one pre-processing step before calling the inner `RuleRunner`

src/AccessibilityInsights.SharedUxTests/Settings/ConfigurationModelTests.cs
```diff
@@ -4,9 +4,9 @@
 using AccessibilityInsights.Desktop.UIAutomation;
 using AccessibilityInsights.DesktopUI.Enums;
 using AccessibilityInsights.RuleSelection;
-using AccessibilityInsights.SetupLibrary;
 using AccessibilityInsights.SharedUx.Enums;
 using AccessibilityInsights.SharedUx.Settings;
+using AccessibilityInsights.SharedUx.Utilities;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using System;
 using System.Globalization;
```
DaveTryon: >35 [](start = 28, length = 2)

My merge to master will cause a merge conflict here

karanbirsingh: thanks for the heads up, I have updated it

src/AccessibilityInsights.Actions/Actions/SaveAction.cs
neocln: IMO we need since when an error is reported by a user, we want to understand which build of rules were used. 

---
In reply to: [271959148](https://github.com/Microsoft/accessibility-insights-windows/pull/256#discussion_r271959148) [](ancestors = 271959148)

karanbirsingh: @neocln I agree, it means we need to update the version number when the rules change. We don't do that at the moment - maybe it can include the axe-windows version at some point.

PR #261, opened 2019-04-05
Fix Exception being thrown when copying VersionSwitcher

DaveTryon:
#### Describe the change
Upgrade is broken because of an Exception being thrown. This used to work, then apparently I made a bad change without catching it. Tested by confirming that upgrade now works.

#### PR checklist

- [ ] Includes UI changes?

### DIFF
--- a/src/AccessibilityInsights.SetupLibrary/VersionSwitcherWrapper.cs
+++ b/src/AccessibilityInsights.SetupLibrary/VersionSwitcherWrapper.cs
@@ -99,7 +99,7 @@ namespace AccessibilityInsights.SetupLibrary
                 foreach (string file in Directory.GetFiles(source))
                 {
                     FileInfo fileInfo = new FileInfo(file);
-                    fileInfo.CopyTo(Path.Combine(dest, file), true);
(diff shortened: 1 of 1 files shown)


src/AccessibilityInsights.SetupLibrary/VersionSwitcherWrapper.cs
```diff
@@ -99,7 +99,7 @@ private static bool TryRecursiveCopy(string source, string dest, List<FileStream
                 foreach (string file in Directory.GetFiles(source))
                 {
                     FileInfo fileInfo = new FileInfo(file);
-                    fileInfo.CopyTo(Path.Combine(dest, file), true);
+                    fileInfo.CopyTo(Path.Combine(dest, fileInfo.Name), true);
```
DaveTryon: >fileInfo.Name [](start = 55, length = 13)

`file` is the fully qualified path, which causes an Exception when passed to Path.Combine. By using `fileInfo.Name`, we get the file name with no path, which is exactly what we need. Bi had this code working, and I must have broken this when I picked up the code for the second set of changes.

PR #262, opened 2019-04-05
Fix problems when copying VersionSwitcher

DaveTryon:
#### Describe the change
The code to copy VersionSwitcher had 2 bugs:
- It was passing the wrong data to Path.Combine, which threw an Exception
- If wasn't properly locking the file after copy

This code was working, then I did some code shuffling and apparently broke it. Tested by forcing an upgrade situation, ensuring that VersionSwitcher gets copied correctly, then ensuring that the upgrade version gets installed. Without this fix, it's impossible to upgrade or switch channels (change already exists in the branch with the channel selection UI, but I wanted to get it into master to fix potential upgrade cases).

#### PR checklist

- [ ] Includes UI changes?

### DIFF
--- a/src/AccessibilityInsights.SetupLibrary/VersionSwitcherWrapper.cs
+++ b/src/AccessibilityInsights.SetupLibrary/VersionSwitcherWrapper.cs
@@ -99,8 +99,9 @@ namespace AccessibilityInsights.SetupLibrary
                 foreach (string file in Directory.GetFiles(source))
                 {
                     FileInfo fileInfo = new FileInfo(file);
-                    fileInfo.CopyTo(Path.Combine(dest, fileInfo.Name), true);
-                    fileLocks.Add(File.OpenRead(dest));
+                    string destFile = Path.Combine(dest, fileInfo.Name);
+                    fileInfo.CopyTo(destFile, true);
+                    fileLocks.Add(File.OpenRead(destFile));
                 }
 
                 // copy folders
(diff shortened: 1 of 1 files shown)


src/AccessibilityInsights.SetupLibrary/VersionSwitcherWrapper.cs
```diff
@@ -99,8 +99,9 @@ private static bool TryRecursiveCopy(string source, string dest, List<FileStream
                 foreach (string file in Directory.GetFiles(source))
                 {
                     FileInfo fileInfo = new FileInfo(file);
-                    fileInfo.CopyTo(Path.Combine(dest, file), true);
```
DaveTryon: >file [](start = 55, length = 4)

`Path.Combine` was throwing here because `file` is the fully qualified path instead of just the name of the file. By using `fileInfo.Name`, we get just the file name.

src/AccessibilityInsights.SetupLibrary/VersionSwitcherWrapper.cs
```diff
@@ -99,8 +99,9 @@ private static bool TryRecursiveCopy(string source, string dest, List<FileStream
...
-                    fileLocks.Add(File.OpenRead(dest));
+                    string destFile = Path.Combine(dest, fileInfo.Name);
+                    fileInfo.CopyTo(destFile, true);
+                    fileLocks.Add(File.OpenRead(destFile));
```
DaveTryon: >destFile [](start = 48, length = 8)

Old code was trying to open `dest`, which is the folder, not the file. Changed this to use `destFile`, which is the file itself.

PR #265, opened 2019-04-06
Adding the releases clean up script 

flyingUnderTheRadar:
#### Describe the change
- This script will run in a power shell task to clean up releases.
- It will run after a custom nuget task to install OcktoKit is run.
- Sample run - https://dev.azure.com/mseng/1ES/_releaseProgress?_a=release-environment-logs&releaseId=990&environmentId=1110
#### PR checklist

- [ ] Includes UI changes?

### DIFF
--- a/tools/scripts/release-cleanup.ps1
+++ b/tools/scripts/release-cleanup.ps1
@@ -90,7 +90,7 @@ foreach($releaseKV in $releaseMap){
         }
         $deleteList += $release
     } elseif ($release.Name -match "Canary"){
-        if($insiderCount -gt 0 -and $canaryCount -ge 2){
+        if(($insiderCount -gt 0 -or $prodCount - gt 0) -and $canaryCount -ge 2){
             $deleteList += $release
             continue
         }
--- a/tools/scripts/release-cleanup.ps1
+++ b/tools/scripts/release-cleanup.ps1
@@ -1,0 +1,108 @@
+﻿# Copyright (c) Microsoft. All rights reserved.
+# Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#This script assumes that the OctoKit Nuget package is a custom install and is in the current folder.
+
+$deleteList = @()
+
+# Extracts a version from the release name.
+function Extract-Version($specificRelease){
+    $versionString = $specificRelease.Name
+    if($versionString -match ".*v(\d+\.\d+\.\d+\.\d+).*"){
+        return $matches[1]
+    } else {
+        $deleteList += $specificRelease
+        return [string]::Empty
+    }
+}
+
+# Sorts the releases based on the version number
+function Sort-Releases($releases){
+    $releaseMap = @{}
+    foreach($release in $releases.Result){
+        $releaseVersion = [System.version](Extract-Version $release)
+        if(-not [string]::IsNullOrEmpty($releaseVersion)){
+            $releaseMap.Add($releaseVersion,$release)
+        }
+    }
+
+    $releaseMap = $releaseMap.GetEnumerator() | Sort-Object -Descending -Property Name
+    return $releaseMap
+}
+
+# Delete a release
+function Delete-Release($release){
+    # Null checks
+    if([string]::IsNullOrEmpty($release.Id)){
+        Write-Host "No Release Id found"
+        return;
+    }
+
+    Write-Host "Deleting release" + $release.Name
+    $response = $instance.Repository.Release.Delete($repoId, $release.Id);
+
+    if(-not ($response.Result -eq "NoContent" -and $response.IsCompleted -eq $True)){
+        Write-Host "Failed to delete release -" $release.Name
+    } 
+}
+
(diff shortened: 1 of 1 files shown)


tools/scripts/release-cleanup.ps1
```diff
@@ -0,0 +1,108 @@
...
+
+# Load the octokit dll
+Add-Type -Path ((Get-Location).Path + "\Octokit.0.32.0\lib\net45\Octokit.dll")
+
+# Get a new client with random product header value
+$productHeader = [Octokit.ProductHeaderValue]::new("AIWindows-CleanupScript")
+$instance = [Octokit.GitHubClient]::new($productHeader)
+
+# Add credentials for authentication
+$instance.Credentials = [Octokit.Credentials]::new("$(gitPATX)")
```
DaveTryon: >gitPATX [](start = 54, length = 7)

How is the actual PAT specified?

tools/scripts/release-cleanup.ps1
```diff
@@ -0,0 +1,108 @@
...
+# Delete a release
+function Delete-Release($release){
+    # Null checks
+    if([string]::IsNullOrEmpty($release.Id)){
+        Write-Host "No Release Id found"
+        return;
+    }
+
+    Write-Host "Deleting release" + $release.Name
+    $response = $instance.Repository.Release.Delete($repoId, $release.Id);
```
DaveTryon: >Release [](start = 37, length = 7)

Does this also delete the tag, or is that something that we need to add?

tools/scripts/release-cleanup.ps1
```diff
@@ -0,0 +1,108 @@
...
+        }
+        $deleteList += $release
+    } elseif ($release.Name -match "Production"){
+        if($prodCount -lt 2){
+            $prodCount++
+            continue
+        }
+        $deleteList += $release
+    } elseif ($release.Name -match "Canary"){
+        if($insiderCount -gt 0 -and $canaryCount -ge 2){
```
DaveTryon: This should also consider prodCount:
```suggestion
        if(($insiderCount -gt 0 -or $prodCount - gt 0) -and $canaryCount -ge 2){
```

tools/scripts/release-cleanup.ps1
flyingUnderTheRadar: The PAT needs to be a secret variable in the pipeline. I have documented the instructions [here](
image
).

tools/scripts/release-cleanup.ps1
```diff
@@ -0,0 +1,108 @@
...
+# Get all releases
+$releases = $instance.Repository.Release.GetAll($repoId)
+
+Write-Host "Releases found" 
+$releases.Result | Select-Object -Property Name, TagName, Id | Format-Table
+
+$releaseMap = Sort-Releases $releases
+Write-Host "==========================================================="
+Write-Host "Sorted Releases" 
+Write-Host ($releaseMap | Out-Default)
```
DaveTryon: >$releaseMap [](start = 12, length = 11)

If there are rogue releases, they'll get deleted without being displayed. Is that what we want?

tools/scripts/release-cleanup.ps1
```diff
@@ -0,0 +1,108 @@
...
+Write-Host "==========================================================="
+
+$prodCount = 0
+$insiderCount = 0
+$canaryCount = 0;
+
+foreach($releaseKV in $releaseMap){
+    $release = $releaseKV.Value
+    if ($release.Name -match "Insider"){
```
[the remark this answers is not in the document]
flyingUnderTheRadar: It does match anywhere. I tried with these 
image

tools/scripts/release-cleanup.ps1
flyingUnderTheRadar: When we delete a release (even from the UI) we do not delete the tag that the release was created at.

What that means is that the MSI will be deleted however you can still download the code from which the MSI was created. 

Git does not have an API to delete tags -  https://developer.github.com/v3/git/tags/

What that means to us is that after a release is deleted we still see the tag in the release view. Personally I feel that this should be okay (and I would actually advocate that we keep these tags) as they help us identify the cut points in the code for these releases

tools/scripts/release-cleanup.ps1
flyingUnderTheRadar: They won't get deleted. They will be added to a list. All delete candidates get added to a list.
We only delete at the end of the script. It will always display what it its attempting to delete.

tools/scripts/release-cleanup.ps1
flyingUnderTheRadar: Discussed offline.

tools/scripts/release-cleanup.ps1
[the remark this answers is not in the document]
flyingUnderTheRadar: Oops. Misread it and discussed offline. Change made to match only at the beginning if the string as suggested.

PR #277, opened 2019-04-08
Enable ReleaseChannel in AutoUpdate

DaveTryon:
#### Describe the change
AutoUpdate needs to be responsive to ReleaseChannel settings (it currently only supports Production)

#### PR checklist

- [ ] Includes UI changes?

### DIFF
--- a/src/AccessibilityInsights.Extensions.GitHubAutoUpdate/AutoUpdate.cs
+++ b/src/AccessibilityInsights.Extensions.GitHubAutoUpdate/AutoUpdate.cs
@@ -170,7 +170,7 @@ namespace AccessibilityInsights.Extensions.GitHubAutoUpdate
         /// Production ctor - MUST be a default ctor or extensions will break
         /// </summary>
         public AutoUpdate() :
-            this(SetupLibrary.ReleaseChannel.Production, () => MsiUtilities.GetInstalledProductVersion(ExceptionReporter),
+            this(ProductionReleaseChannelProvider, () => MsiUtilities.GetInstalledProductVersion(ExceptionReporter),
                 new ProductionChannelInfoProvider(new GitHubWrapper(ExceptionReporter), ExceptionReporter))
         {
         }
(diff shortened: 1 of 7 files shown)


src/AccessibilityInsights.Extensions.GitHubAutoUpdate/ProductionChannelInfoProvider.cs
```diff
@@ -28,7 +28,7 @@ public ProductionChannelInfoProvider(IGitHubWrapper gitHubWrapper, IExceptionRep
         /// </summary>
         public bool TryGetChannelInfo(ReleaseChannel releaseChannel, out ChannelInfo channelInfo)
         {
-            return ChannelInfoUtilities.TryGetChannelInfo(releaseChannel, out channelInfo, _gitHubWrapper, null, _exceptionReporter);
```
DaveTryon: >null [](start = 107, length = 4)

Passing null here was a bug. We want to use the default parameter, hence the named parameter syntax

src/AccessibilityInsights/MainWindowHelpers/Initialize.cs
```diff
@@ -114,6 +115,9 @@ private static void PopulateConfigurations()
             // when AccessibilityInsights starts up. 
             ConfigurationManager.GetDefaultInstance().AppConfig.IsUnderElementScope = true;
 
+            // Configure the correct ReleaseChannel for autoupdate
+            Container.SetAutoUpdateReleaseChannel(ConfigurationManager.GetDefaultInstance().AppConfig.ReleaseChannel.ToString());
+
             // enable/disable telemetry
             if (ConfigurationManager.GetDefaultInstance().AppConfig.EnableTelemetry)
                 TelemetryController.EnableTelemetry();
```
DaveTryon: >TelemetryController [](start = 16, length = 19)

This is where the Container currently gets initialized, so we need to set the ReleaseChannel before reaching this point

src/AccessibilityInsights.Extensions/Container.cs
```diff
@@ -145,6 +145,16 @@ public static Container GetDefaultInstance()
 
             return _defaultInstance;
         }
+
+        /// <summary>
+        /// Set the ReleaseChannel for update
+        /// </summary>
+        /// <param name="releaseChannel">The value to use in the AutoUpdate code</param>
+        public static void SetAutoUpdateReleaseChannel(string releaseChannel)
```
DaveTryon: >string [](start = 55, length = 6)

This uses a string to keep Extensions definitions decoupled from the SetupLibrary extension

src/AccessibilityInsights.Extensions.GitHubAutoUpdate/AutoUpdate.cs
```diff
@@ -170,26 +170,37 @@ private UpdateResult Update()
...
         {
-            _strongReleaseChannel = releaseChannel;
+            _strongReleaseChannel = releaseChannelProvider();
             _installedVersionProvider = installedVersionProvider;
             _channelInfoProvider = channelInfoProvider;
             ReleaseChannel = _strongReleaseChannel.ToString();
             _initTask = Task.Run(() => InitializeWithTimer());
         }
 
+        private static ReleaseChannel ProductionReleaseChannelProvider()
```
karanbirsingh: can we consider renaming so it doesn't have Production in the name?

DaveTryon: Renamed to ConfiguredReleaseChannelProvider

---
In reply to: [273264454](https://github.com/Microsoft/accessibility-insights-windows/pull/277#discussion_r273264454) [](ancestors = 273264454)

PR #278, opened 2019-04-08
Updates to the docs after axe-windows rename

karanbirsingh:
We want to keep the docs relatively up-to-date, so this reflects the rename to axe-windows for some assemblies. We could probably restructure some of the headings but since this area is still in some flux I don't think it makes sense to do until we have settled fully.

### DIFF
--- a/docs/Overview.md
+++ b/docs/Overview.md
@@ -18,20 +18,19 @@ These assemblies provide the interaction with UIA, as well as layers that allow 
 
 Assembly | Responsibility
 --- | ---
-AccessibilityInsights.Core | Provide data abstractions which represent accessibility data in a platform-agnostic way.
-AccessibilityInsights.Desktop | Provide platform-specific (Windows) implementations of the platform-agnostic data abstractions. The low-level interactions with UIA occur in this assembly.
-AccessibilityInsights.DesktopUI | Provide platform-specific (Windows) implementations of platform-agnostic UI abstractions. This includes things like element highlighters.
-AccessibilityInsights.Actions | Provide a high-level set of Actions that are the primary interface into the Runtime components.
-AccessibilityInsights.Extensions | Provide extension points that allow certain non-core functionality to be implemented in a loosely coupled way.
-AccessibilityInsights.Win32 | Provide a wrapper around Win32-specific code that is needed by other assemblies.
+Axe.Windows.Actions | Provide a high-level set of Actions that are the primary interface into the Runtime components.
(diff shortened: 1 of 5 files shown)


docs/Overview.md
```diff
@@ -40,9 +39,11 @@ Assembly | Responsibility
...
 AccessibilityInsights.CommonUxComponents | Provide non-specialized visual elements used by the main app and extensions. This allows the main app and extensions to share ux components that are unrelated to the runtime.
+AccessibilityInsights.DesktopUI | Provide platform-specific (Windows) implementations of platform-agnostic UI abstractions. This includes things like element highlighters.
+AccessibilityInsights.Extensions | Provide extension points that allow certain non-core functionality to be implemented in a loosely coupled way.
 AccessibilityInsights.SharedUx | Provide visual elements used by the main app. This code is in a separate assembly for historical reasons.
 AccessibilityInsights.WebApiHost | Provide a local service that exposes scanning functionality on locally running applications.
-AccessibilityInsights.Automation | Provide a layer that wraps key actions behind a simplified interface. This layer can then be used either from a .NET application or from PowerShell scripts.
+Axe.Windows.Automation | Provide a layer that wraps key actions behind a simplified interface. This layer can then be used either from a .NET application or from PowerShell scripts.
```
DaveTryon: >Windows [](start = 4, length = 7)

Axe-Windows components appear across 3 different categories. Would it make sense to try to create a separate Axe-Windows category? They'll eventually be removed when the second repo is created, so maybe it should wait until then?

docs/Overview.md
```diff
@@ -72,23 +73,24 @@ The packaging projects exist to gather assemblies into their shipping vehicles:
 Project | Responsibility
 --- | ---
 MSI | Builds the MSI file used by most users.
-ApplicationInsights-CI | Builds the NuGet package that will be used by users who wish to scan via automation.
+AxeWindows-CI | Builds the NuGet package that will be used by users who wish to scan via automation.
```
DaveTryon: >scan via automation [](start = 80, length = 19)

When the package comes from a second repo, the package will cover more than just automation. Not sure what that means for this comment

DaveTryon:
> #### Runtime components

AxeWindows components?

---
Refers to: docs/Overview.md:16 in 926c8e2. [](commit_id = 926c8e2496916ae897120f197e1c2705406f5e5a, deletion_comment = False)

docs/Overview.md
karanbirsingh: IMO I agree with the sentiment but don't see much value in doing it until we do the switch

Issue #274, opened 2019-04-08
[BUG] ControlViewTabStructure should allow Button


LiLoDavis:
**Current rule**
The given element is expected to have the following structure: Tab and AnyChild(IsControlElement) and AllChildren(not(IsControlElement) or (Group or ScrollBar or TabItem)).

**Proposed rule**
Rule should allow Button.
See the documentation: https://docs.microsoft.com/en-us/windows/desktop/winauto/uiauto-supporttabcontroltype#typical-tree-structure

karanbirsingh:
We have added a new rule that validates ScrollBar structure. That rule ensures that a ScrollBar has either a button or a thumb as its child - this should indirectly resolve this issue. The next canary [release](https://github.com/Microsoft/accessibility-insights-windows/releases) (4/11) should include this fix.

ferBonnin:
Hi @LiLoDavis this issue has been resolved. Could you verify?
Thanks!

LiLoDavis:
Solution makes sense to me.

ferBonnin:
Closing issue then :)

PR #279, opened 2019-04-10
New rule for scrollbar structure

karanbirsingh:
#### Describe the change

This PR adds a rule that checks whether Control View scrollbars adhere to the structure given here: https://docs.microsoft.com/en-us/windows/desktop/WinAuto/uiauto-supportscrollbarcontroltype

It is partly in response to issue #274 - when a complex structure like a tab has children controls like a scrollbar, the scrollbar's children can be checked using this rule instead of checking the full tree at the tab-level.

### DIFF
--- a/src/AccessibilityInsights.Rules/Library/Structure/ControlView/Scrollbar.cs
+++ b/src/AccessibilityInsights.Rules/Library/Structure/ControlView/Scrollbar.cs
@@ -1,0 +1,33 @@
+﻿using Axe.Windows.Core.Bases;
+using Axe.Windows.Core.Enums;
+using Axe.Windows.Rules.PropertyConditions;
+using Axe.Windows.Rules.Resources;
+using System;
+using System.Globalization;
+using static Axe.Windows.Rules.PropertyConditions.ControlType;
+
+namespace Axe.Windows.Rules.Library
+{
+    [RuleInfo(ID = RuleId.ControlViewScrollbarStructure)]
+    class ControlViewScrollbarStructure : Rule
+    {
+        public ControlViewScrollbarStructure()
+        {
+            this.Info.Description = string.Format(CultureInfo.InvariantCulture, Descriptions.Structure, ControlView.ScrollbarStructure);
+            this.Info.HowToFix = string.Format(CultureInfo.InvariantCulture, HowToFix.Structure, ControlView.ScrollbarStructure);
+            this.Info.Standard = A11yCriteriaId.InfoAndRelationships;
+        }
+
+        public override EvaluationCode Evaluate(IA11yElement e)
+        {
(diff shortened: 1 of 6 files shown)


src/AccessibilityInsights.Rules/Library/Structure/ControlView/Scrollbar.cs
```diff
@@ -0,0 +1,33 @@
...
+        public override EvaluationCode Evaluate(IA11yElement e)
+        {
+            if (e == null) throw new ArgumentException(nameof(e));
+
+            return ControlView.ScrollbarStructure.Matches(e) ? EvaluationCode.Pass : EvaluationCode.Note;
+        }
+
+        protected override Condition CreateCondition()
+        {
+            return ScrollBar;
```
neocln: this rule only run when the ScrollBar control is marked with IsControlElementProperty true value. 
Please make sure that condition is added.

src/AccessibilityInsights.RulesTest/Library/Structure/ControlView/ScrollbarTest.cs
```diff
@@ -0,0 +1,112 @@
...
+using Moq;
+using System.Collections.Generic;
+using System.Linq;
+using EvaluationCode = Axe.Windows.Rules.EvaluationCode;
+
+namespace Axe.Windows.RulesTest.Library
+{
+    [TestClass]
+    public class ControlViewScrollbarStructureTest
+    {
```
neocln: Test cases should be updated based on the other comment on ScrollBar.cs file.

src/AccessibilityInsights.Rules/Library/Structure/ControlView/Scrollbar.cs
karanbirsingh: I think this is abstracted inside the `ScrollBar` condition:
https://github.com/Microsoft/accessibility-insights-windows/blob/e455215a219a975d9ac05be7d25f37719a6e613a/src/AccessibilityInsights.Rules/PropertyConditions/ControlType.cs#L33

karanbirsingh: I mistakenly thought the ScrollBar condition handled this. I will check with @RobGallo - it seems like none of the rules consider this property so we may need to fix it separately:
> Another such failing is that the children are not actually checked to make sure they have the IsContentElement or IsControlElement properties set to true. This failing is common throughout these rules. And again, there is no intent to remedy that as part of this change.
from #281

karanbirsingh: tracking in #282

Issue #263, opened 2019-04-05
[BUG] Chrome menus can exceed parent bounding rects but AI-Win flags this as an error 


karanbirsingh:
**Describe the bug**
In Chrome, the settings menu pane does not clip based on the browser's size. Instead, the pane will exceed the size of the browser and always show its children. Because of this, ai-win flags the pane because its bounding rectangle is not contained within the bounding rectangle of its parent. The pane itself seems like a window. This rule should be revised to either omit the check in this case (using NativeWindowHandle + Pane control type + class name heuristics) or be changed to a 'best practice' rule.

**To Reproduce**
1. Open any website in Google Chrome.
2. Decrease the size of the browser so the following menu exceeds the browser size.
3. Select the browser in ai-win and start a test timer for 5 seconds.
4. Select the customize google chrome button (labeled '...') on the top right to open the menu.
5. View test results.

**Expected behavior**
The pane should not be flagged for an invalid bounding rectangle.

**Actual behavior**
The pane is flagged for a bounding rectangle that exceeds the size of its parents.

**Screenshots or .GIF**

![menu overlaying browser](https://user-images.githubusercontent.com/7775527/55649546-204be100-5798-11e9-824e-c8e3a8e1116b.PNG)

**Desktop (please complete the following information):**
 - OS: Windows 10 1809
 - Accessibility Insights for Windows Version: 1.1.794.1
 - Target Application: Chrome
 - Target Application Version: 73.0.3683.86

karanbirsingh:
rename as .a11ytest to open in ai-win: [ChromeMenu.zip](https://github.com/Microsoft/accessibility-insights-windows/files/3049298/ChromeMenu.zip)

karanbirsingh:
fixed in https://github.com/Microsoft/accessibility-insights-windows/releases/tag/v1.1.0831.01

PR #280, opened 2019-04-10
Fixing Issue #263 Chrome menus can exceed parent bounding rects but A I-Win flags this as an error

RobGallo:
Some Chrome panes are essentially windows, and we already treat windows as exceptions to this rule. Added code to make the rule recognize these Chrome panes as windows.

Also, fixed code in ValueCondition which was not producing the correct description strings.

#### Describe the change

#### PR checklist

- [ ] Includes UI changes?


src/AccessibilityInsights.RulesTest/Library/BoundingRectangleContainedInParentTest.cs
```diff
@@ -294,5 +294,27 @@ public void BoundingRectangleContainedInParent_HorizontallyScrollablePass()
 
             Assert.AreEqual(Rule.Evaluate(e), EvaluationCode.Pass);
         }
+
+        [TestMethod]
+        public void BoundingRectangleContainedInParent_ChromeWindow_NotApplicable()
+        {
+            using (var e = new MockA11yElement())
```
karanbirsingh: I wonder if we should use moq instead for new tests?

karanbirsingh: discussed offline, we will consider refactoring MockA11yElement over time to internally use moq

src/AccessibilityInsights.RulesTest/PropertyConditions/StringPropertiesTest.cs
```diff
@@ -264,5 +264,38 @@ public void TestStringPropertyIsEqualBothNullFalse()
                 Assert.IsFalse(condition.Matches(e));
             } // using
         }
+
+        [TestMethod]
+        public void StringProperty_MatchesRegEx_True()
```
DaveTryon:
```suggestion
        public void MatchesRegEx_CaseMatches_True()
```

src/AccessibilityInsights.RulesTest/PropertyConditions/StringPropertiesTest.cs
```diff
@@ -264,5 +264,38 @@ public void TestStringPropertyIsEqualBothNullFalse()
...
+            using (var e = new MockA11yElement())
+            {
+                e.Name = "Eat your food";
+                var condition = Name.MatchesRegEx(@"fo+");
+                Assert.IsTrue(condition.Matches(e));
+            } // using
+        }
+
+        [TestMethod]
+        public void StringProperty_MatchesRegEx_False()
```
DaveTryon:
```suggestion
        public void MatchesRegEx_CaseDoesNotMatch_False()
```

src/AccessibilityInsights.RulesTest/PropertyConditions/StringPropertiesTest.cs
```diff
@@ -264,5 +264,38 @@ public void TestStringPropertyIsEqualBothNullFalse()
...
+            using (var e = new MockA11yElement())
+            {
+                e.Name = "food";
+                var condition = Name.MatchesRegEx(@"Fo+");
+                Assert.IsFalse(condition.Matches(e));
+            } // using
+        }
+
+        [TestMethod]
+        public void StringProperty_MatchesRegExNoCase_True()
```
DaveTryon:
```suggestion
        public void MatchesRegExNoCase_CaseDoesNotMatch_True()
```

src/AccessibilityInsights.Rules/PropertyConditions/StringProperty.cs
```diff
@@ -118,5 +140,25 @@ private bool IsEqualTo(IA11yElement e, StringProperty that)
...
+
+        public Condition MatchesRegEx(string s, RegexOptions options)
+        {
+            return Condition.Create(e =>
+            {
+                Regex r = new Regex(s, options);
+                return r.IsMatch(GetStringPropertyValue(e));
+                },
+                String.Format(ConditionDescriptions.MatchesRegExWithOptions, PropertyDescription, s, options.ToString()));
+        }
```
DaveTryon: Lines 160 and 161 each have one extra indent

src/AccessibilityInsights.Rules/Library/BoundingRectangleContainedInParent.cs
```diff
@@ -70,10 +72,15 @@ private static bool IsBoundingRectangleContained(IA11yElement container, IA11yEl
         protected override Condition CreateCondition()
         {
             // Windows can be any size, regardless of their parents
+            // Certain chrome panes are treated as windows because that's how they behave.
```
neocln: is this issue only for chrome? IMO, it can happen to any application. if then, how about not considering this case for Menu than excluding Chrome case?

RobGallo: This fix is specifically for Chrome, based on an issue we got from a user. As for making this more widely applicable to all menus, I feel like we just don't have enough info right now to justify making that change. I think it would be just as likely to cause an issue as to solve one. Also, if we relax the requirements, that means any cases where a menu should (for some reason) remain in a window would never get called out and therefore be invisible forever. In short, I would like to wait for more data.

neocln: I disagree with you on the resolution. I'll share one other case which I can think. Let's talk about this more.

PR #284, opened 2019-04-10
Add dialog to confirm release channel change

jalkire:
#### Describe the change
- Created ContainedDialog abstract class for dialog-like controls that block the ui but aren't separate windows
- Refactored telemetry opt-in experience to use ContainedDialog (should not affect user experience)
- Created ReleaseChannelContainedDialog to confirm changes to release channel

#### PR checklist

- [x] Includes UI changes?

image


src/AccessibilityInsights.SharedUx/Dialogs/ChangeChannelContainedDialog.xaml.cs
```diff
@@ -0,0 +1,38 @@
...
+using AccessibilityInsights.SetupLibrary;
+using System.Globalization;
+using System.Windows;
+
+namespace AccessibilityInsights.SharedUx.Dialogs
+{
+    /// <summary
+    /// Interaction logic for ChangeChannelContainedDialog.xaml
+    /// </summary>
+    public partial class ChangeChannelContainedDialog : ContainedDialog
```
neocln: how does it handle {ESC} button? typically, we close a dialog when ESC key is pressed as a cancel action. I don't see a code for that.

src/AccessibilityInsights.SharedUx/Dialogs/ContainedDialog.cs
```diff
@@ -0,0 +1,57 @@
...
+namespace AccessibilityInsights.SharedUx.Dialogs
+{
+    public abstract class ContainedDialog : UserControl
+    {
+        /// <summary>
+        /// Used to determine when the "dialog" is ready to return its result
+        /// </summary>
+        protected EventWaitHandle WaitHandle { get; } = new AutoResetEvent(true);
+
+        protected bool DialogResult { get; set; } = false;
```
neocln: There is a known dialog result type to indicate many different return values. how about using that type here than "bool"? 
also having it in the base class would be good.

src/AccessibilityInsights/Modes/ContainedDialogModeControl.xaml.cs
```diff
@@ -0,0 +1,45 @@
...
+            InitializeComponent();
+        }
+
+        /// <summary>
+        /// Override LocalizedControlType
+        /// </summary>
+        /// <returns></returns>
+        protected override AutomationPeer OnCreateAutomationPeer()
+        {
+            return new CustomControlOverridingAutomationPeer(this, "page");
```
neocln: how about pulling out "page" string into resource file? 
since "page" is the localized control type, it makes sense to be a resource string.

src/AccessibilityInsights.SharedUx/Dialogs/ContainedDialog.cs
jalkire: You might be thinking of the [Win Forms DialogResult](https://docs.microsoft.com/en-us/dotnet/api/system.windows.forms.dialogresult?view=netframework-4.7.2). I'm following the [Window.ShowDialog pattern](https://docs.microsoft.com/en-us/dotnet/api/system.windows.window.showdialog?view=netframework-4.7.2), which returns `bool?`. This is what wwe use for our own `MessageDialog` class as well. I could make it a nullable bool for the sake of consistency, but even that doesn't seem necessary to me.

neocln: I see. it makes sense.

src/AccessibilityInsights.SharedUx/Dialogs/ChangeChannelContainedDialog.xaml.cs
jalkire: ESC now handled

src/AccessibilityInsights/Modes/ContainedDialogModeControl.xaml.cs
jalkire: page pulled out here & in other mode controls

src/AccessibilityInsights/Modes/ContainedDialogModeControl.xaml.cs
```diff
@@ -0,0 +1,45 @@
...
+using System.Windows.Controls;
+
+namespace AccessibilityInsights.Modes
+{
+    /// <summary>
+    /// Interaction logic for ContainedDialogModeControl.xaml
+    /// </summary>
+    public partial class ContainedDialogModeControl : UserControl
+    {
+        public ContainedDialogModeControl()
```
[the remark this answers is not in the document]
jalkire: Changed name to `DialogContainerModeControl`

src/AccessibilityInsights/Modes/ConfigurationModeControl.xaml.cs
```diff
@@ -141,22 +143,38 @@ private void buttonOk_Click(object sender, RoutedEventArgs e)
...
+
+            Dispatcher.Invoke(() => MainWin.TransitionToSelectActionMode());
+        }
+
+        private async Task<bool> HandleReleaseSwitch()
+        {
+            var channelDialog = new ChangeChannelContainedDialog(appSettingsCtrl.SelectedReleaseChannel);
+
+            if (appSettingsCtrl.SelectedReleaseChannel == Configuration.ReleaseChannel ||
+                !await MainWin.ctrlTelemetryDialog.ShowControl(channelDialog).ConfigureAwait(false))
```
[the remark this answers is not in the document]
jalkire: Forgot to rename that...

src/AccessibilityInsights/Modes/ConfigurationModeControl.xaml.cs
```diff
@@ -141,22 +143,38 @@ private void buttonOk_Click(object sender, RoutedEventArgs e)
...
+                VersionSwitcherWrapper.ChangeChannel(appSettingsCtrl.SelectedReleaseChannel);
+                Dispatcher.Invoke(() => MainWin.Close());
+                return true;
+            }
+            catch (Exception ex)
+            {
+                ex.ReportException();
+
+                Dispatcher.Invoke(() =>
+                    MessageDialog.Show(string.Format(CultureInfo.InvariantCulture, Properties.Resources.ConfigurationModeControl_VersionSwitcherException, ex.Message))
```
DaveTryon: >ex.Message [](start = 155, length = 10)

As we discussed, using the raw exception text here (at least when I have no network) is neither actionable nor pretty. We've already reported the Exception via telemetry, so let's display a message here that is more applicable to the user

jalkire: Message updated

jalkire: Message updated

Issue #270, opened 2019-04-08
[BUG] ControlShouldNotSupportValuePattern should consider Name property


LiLoDavis:
**Current rule**
Currently looks for text controls that support the value pattern. Returns as uncertain instead of error.

**Proposed rule**
1. Look for text controls that support the value pattern.
2. Compare content of name and value.
3. Return error if name and value differ.

neocln:
@RobGallo Based on MSDN, ValuePattern is not supported in Text Control. 
however, because there are cases of having value patterns in Text Control in some frameworks, we had it as warning than error. 
https://docs.microsoft.com/en-us/windows/desktop/winauto/uiauto-supporttextcontroltype
I would like to understand what this rule change is based on.

karanbirsingh:
@LiLoDavis I checked in with @neocln & @RobGallo - since other rules ensure that an automation name must exist, the additional presence of a value pattern does not have high user impact - thus it is listed as 'uncertain'. The name/value comparison error is interesting and we can consider it if it is reflected in the MSDN docs

karanbirsingh:
accidental close, marking as resolve instead

PR #286, opened 2019-04-10
Update test to ensure any ValuePattern info for Text controls is also exposed via the name property

karanbirsingh:
This change updates the rules in response to #270 so that the name property is considered. Specifically, any Text controls that support the ValuePattern should not provide extra information via the value that is not already in the name.

### DIFF
--- a/src/AccessibilityInsights.Rules/Library/ControlShouldNotSupportValuePattern.cs
+++ b/src/AccessibilityInsights.Rules/Library/ControlShouldNotSupportValuePattern.cs
@@ -3,6 +3,7 @@
 using System;
 using Axe.Windows.Core.Bases;
 using Axe.Windows.Core.Enums;
+using Axe.Windows.Core.Types;
 using Axe.Windows.Rules.PropertyConditions;
 using Axe.Windows.Rules.Resources;
 using static Axe.Windows.Rules.PropertyConditions.ControlType;
@@ -22,15 +23,19 @@ namespace Axe.Windows.Rules.Library
         public override EvaluationCode Evaluate(IA11yElement e)
         {
             if (e == null) throw new ArgumentException(nameof(e));
+            
+            var name = e.Name.Trim().ToLower();
+            var value = e.GetPattern(PatternType.UIA_ValuePatternId).GetValue<string>("Value").Trim().ToLower();
 
-            return Patterns.Value.Matches(e) ? EvaluationCode.Warning : EvaluationCode.Pass;
+            bool valuePatternExceedsName = value.Length > e.Name.Length || !name.Contains(value);
+            return valuePatternExceedsName ? EvaluationCode.Error : EvaluationCode.Pass;
         }
 
         protected override Condition CreateCondition()
         {
             // Removed Calendar, even though it is specified in MSDN
             // This seems to be an error in the documentation, or at least it seems to no longer be valid
-            return Text;
+            return Text & Patterns.Value & ValuePattern.ValueProperty.NotNullOrEmpty;
         }
     } // class
 } // namespace
--- a/src/AccessibilityInsights.Rules/PropertyConditions/ValuePattern.cs
+++ b/src/AccessibilityInsights.Rules/PropertyConditions/ValuePattern.cs
@@ -1,0 +1,19 @@
+﻿using Axe.Windows.Core.Bases;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Axe.Windows.Rules.PropertyConditions
+{
+    /// <summary>
+    /// Contains commonly used conditions for testing against the Value pattern of an IA11yElement.
+    /// </summary>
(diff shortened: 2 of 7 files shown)


RobGallo:
> @RobGallo Based on MSDN, ValuePattern is not supported in Text Control.
> however, because there are cases of having value patterns in Text Control in some frameworks, we had it as a warning than an error.
> https://docs.microsoft.com/en-us/windows/desktop/winauto/uiauto-supporttextcontroltype
> I would like to understand what this rule change is based on.

@neocln
The intent of forbidding the value pattern by the docs is to prevent people from adding content in the value pattern that they don't include in the Name property. We marked it as a warning because sometimes the value pattern exists, but it may or may not be an error because there may also be content in the Name property. This change makes it so that if the value pattern exists and contains <= the data contained in the Name property, no warning is given because the user has not been deprived of any information. However, if there is data provided by the value pattern which is not part of the name, information has been provided which will likely not make it to the user, and the result is therefore an error.

src/AccessibilityInsights.Rules/Library/ControlShouldNotSupportValuePattern.cs
```diff
@@ -22,15 +23,19 @@ public ControlShouldNotSupportValuePattern()
         public override EvaluationCode Evaluate(IA11yElement e)
         {
             if (e == null) throw new ArgumentException(nameof(e));
+            
+            var name = e.Name.Trim().ToLower();
+            var value = e.GetPattern(PatternType.UIA_ValuePatternId).GetValue<string>("Value").Trim().ToLower();
```
RobGallo: Even though you have already checked to make sure the value pattern exists and the value is not null as part of the condition, it may not be a bad idea to confirm they are not null here. My concern is that the condition could easily change without the developer realizing the evaluation depends on those checks.

src/AccessibilityInsights.Rules/PropertyConditions/ValuePattern.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using Axe.Windows.Core.Bases;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
```
RobGallo: Remove unused "using" statements.

src/AccessibilityInsights.Rules/Library/ControlShouldNotSupportValuePattern.cs
RobGallo: Should we use `"Value"` here or `ValuePattern.ValuePropertyString`?

karanbirsingh:
> @RobGallo Based on MSDN, ValuePattern is not supported in Text Control.
> however, because there are cases of having value patterns in Text Control in some frameworks, we had it as a warning than an error.
> https://docs.microsoft.com/en-us/windows/desktop/winauto/uiauto-supporttextcontroltype
> I would like to understand what this rule change is based on.
> 
> @neocln
> The intent of forbidding the value pattern by the docs is to prevent people from adding content in the value pattern that they don't include in the Name property. We marked it as a warning because sometimes the value pattern exists, but it may or may not be an error because there may also be content in the Name property. This change makes it so that if the value pattern exists and contains <= the data contained in the Name property, no warning is given because the user has not been deprived of any information. However, if there is data provided by the value pattern which is not part of the name, information has been provided which will likely not make it to the user, and the result is therefore an error.

@neocln @RobGallo I wonder if another solution could include returning 'warning' instead of 'pass' to match the MSDN docs while still returning error for egregious cases

neocln:
> > @RobGallo Based on MSDN, ValuePattern is not supported in Text Control.
> > however, because there are cases of having value patterns in Text Control in some frameworks, we had it as a warning than an error.
> > https://docs.microsoft.com/en-us/windows/desktop/winauto/uiauto-supporttextcontroltype
> > I would like to understand what this rule change is based on.
> 
> @neocln
> The intent of forbidding the value pattern by the docs is to prevent people from adding content in the value pattern that they don't include in the Name property. We marked it as a warning because sometimes the value pattern exists, but it may or may not be an error because there may also be content in the Name property. This change makes it so that if the value pattern exists and contains <= the data contained in the Name property, no warning is given because the user has not been deprived of any information. However, if there is data provided by the value pattern which is not part of the name, information has been provided which will likely not make it to the user, and the result is therefore an error.

@RobGallo I see your point. if then, isn't it better to split the rule into two rules? one for checking not to have. and the other for checking the contents of ValuePattern when the pattern exist? 
the whole purpose of this scan rule is to check the existence of the pattern when it should not. 
mixing these two, IMHO, is not aligned with what we have done in our rule set, is it?

karanbirsingh:
Adding do-not-merge tag while we hash out the best fix to match with docs & check name vs. value

karanbirsingh:
Closing until we have more clarity

Issue #298, opened 2019-04-15
[BUG] Link to release notes always goes to Production


DaveTryon:
**Describe the bug**
The link to the release notes always points to the production release notes, regardless of release channel

**To Reproduce**
1. Go to the What's new screen
2. Click on "go to to the latest release notes"
3. See error

**Expected behavior**
This should go either to the release notes for the Channel or the specific Release

**Actual behavior**
This goes to the releases notes on accessibilityinsights.io (which is always production)

**Screenshots or .GIF**

**Desktop (please complete the following information):**
 - OS: Windows 10 1903
 - Accessibility Insights for Windows Version: 1.1.0834.01

**Additional context**
Priority requested -

AhmedAbdoOrtiga:
- Do we have a release note page for each channel or version to use as suggested in the "Expected behavior"?
- Also since the official place to download and use AI-Win is [AI-Win](https://aka.ms/accessibilityinsights-windows/download) which matches with the production release notes [Notes](https://accessibilityinsights.io/docs/en/windows/reference/releases), is this still a valid issue?

PR #302, opened 2019-04-15
Issue [#298] - direct link from What's new page to release-specific page

DaveTryon:
#### Describe the change
Make the release notes page in the What's New dialog point to a release-specific page, instead of a generic page in the docs. This makes it work for all channels, and eventually lets us reduce the amount of duplicated content that we need to maintain

#### PR checklist

- [x] Does this address an existing issue? If yes, Issue# - [#298]
- [ ] Includes UI changes?

Uri before change was a fwlink that pointed to https://accessibilityinsights.io/docs/en/windows/reference/releases. 
Uri after the change is build-specific. With build 1.1.0831.02 installed, it points to https://github.com/Microsoft/accessibility-insights-windows/releases/tag/v1.1.0831.02.

### DIFF
--- a/src/AccessibilityInsights/Modes/StartUpModeControl.xaml.cs
+++ b/src/AccessibilityInsights/Modes/StartUpModeControl.xaml.cs
@@ -11,10 +11,10 @@ using System.Diagnostics;
 using System.Globalization;
 using System.Threading.Tasks;
 using System.Windows;
+using System.Windows.Automation.Peers;
 using System.Windows.Controls;
 using System.Windows.Input;
 using System.Windows.Navigation;
-using System.Windows.Automation.Peers;
 
 namespace AccessibilityInsights.Modes
 {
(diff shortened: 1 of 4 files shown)


src/AccessibilityInsights/Modes/StartUpModeControl.xaml.cs
```diff
@@ -1,19 +1,20 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using AccessibilityInsights.CommonUxComponents.Dialogs;
+using AccessibilityInsights.SetupLibrary;
```
DaveTryon: Accessibility.SetupLibrary is the only addition here--the rest have just been sorted

src/AccessibilityInsights/Modes/StartUpModeControl.xaml.cs
```diff
@@ -1,19 +1,20 @@
...
 using System;
+using System.Diagnostics;
+using System.Globalization;
 using System.Threading.Tasks;
 using System.Windows;
 using System.Windows.Controls;
 using System.Windows.Input;
 using System.Windows.Navigation;
-using AccessibilityInsights.SharedUx.Settings;
 using System.Windows.Automation.Peers;
```
DaveTryon: This is not sorted :(

DaveTryon: Fixed in commit 2

PR #309, opened 2019-04-17
Move GetAppVersion out of Axe.Core

AhmedAbdoOrtiga:
#### Describe the change

#### PR checklist

- [ ] Includes UI changes?


DaveTryon:
I notice that this is from a branch. In the future, let's try to use forks instead of branches, as discussed in retro and/or engineering meetings

src/AccessibilityInsights.SharedUx/Settings/ConfigurationModel.cs
```diff
@@ -652,5 +652,15 @@ public object Clone()
             return new ConfigurationModel(this);
         }
         #endregion
+
+        /// <summary>
+        /// Get version from AccessibilityInsights.SharedUx Assembly
+        /// </summary>
+        /// <returns></returns>
+        public static string GetAppVersionFromAssembly()
```
DaveTryon: >GetAppVersionFromAssembly [](start = 29, length = 25)

The ConfigurationModel class is intended to hold user config data. The assembly version is not user config data, and therefore doesn't belong in this class. There are many places where it might logically live. One easy option would be new static class in SharedUx.Misc.

AhmedAbdoOrtiga: I think we still keep it in the settings folder, I don't like creating a separate utility classes as we used to do in Core to get the app version, cause thinking OOP, as you say there should be some class logically responsible for getting the assembly app version, so I think that the "ConfigurationManager" can do that and there's no need for a new class, what do you think?

jalkire:
> I notice that this is from a branch. In the future, let's try to use forks instead of branches, as discussed in retro and/or engineering meetings

Why?

src/AccessibilityInsights.SharedUx/Settings/ConfigurationModel.cs
jalkire: Agree with Dave that this shouldn't be in ConfigurationModel. Maybe ConfigurationManager makes more sense, but it still seems like this isn't "configuration"

DaveTryon: This isn't config. It's a static value that we're exposing. The "Single-responsibility principle" (the "S" in "SOLID") tells us that this belongs in a class with similar responsibilities. If we don't have one, we should create one.

---
In reply to: [276050404](https://github.com/Microsoft/accessibility-insights-windows/pull/309#discussion_r276050404) [](ancestors = 276050404)

RobGallo: @DaveTryon  is right on the money here. This should be a separate class and definitely not part of the configuration manager. One way to tell this is not the right class for the job is to ask the question: why does the configuration manager class change. The answer: it changes when configuration options change. Classes change for one reason and one reason only. Therefore, version does not belong; because if the version logic changes, we should not change the configuration class.

DaveTryon:
@John Alkire, I suppose there are several possible answers--here are 4, in (IMO) decreasing importance:
1. We discussed forks vs. branches at length in the 2/15 engineering discussion and concluded that we ought to be following the same patterns that will be followed by external contributors. To my knowledge, the web team uses exclusively forks, and most of the windows team uses exclusively forks. With the possible exception of things like testing build triggers, forks provide everything needed for development and debugging (and since each fork can have an unlimited number of branches, you still retain the full flexibility of how to manage branches within your individual fork).
2. Forks are inherently lean--branches in forks appear only in their "home" fork or in forks derived from it. Branches in the main project get replicated to *every* fork, and to *every* clone
3. The main repo has 3 specific branches, each with a specific purpose and policy. There's virtue in keeping that list small and uncluttered
4. Using forks means you don't have to deal with my annoying "can we delete this branch" emails

Happy to discuss this further offline if you wish

---
In reply to: [484155452](https://github.com/Microsoft/accessibility-insights-windows/pull/309#issuecomment-484155452) [](ancestors = 484155452)

PR #313, opened 2019-04-17
Simplify bug URLs with special characters

karanbirsingh:
#### Describe the change

This PR proactively limits the characters allowed in the Azure DevOps bug URLs to a basic (e.g. < 127 ascii) set to limit navigation issues. For example, filing a bug from any rule with how-to-fix text containing the middle dot · character can lead to a redirect loop if the user signs out from the hosted web browser prior to filing the issue. After this PR, filing an issue from any of these rules properly navigates to the page. Any non-basic characters are omitted from the text.

### DIFF
--- a/src/AccessibilityInsights.Extensions.AzureDevOps/AzureDevOpsIntegration.cs
+++ b/src/AccessibilityInsights.Extensions.AzureDevOps/AzureDevOpsIntegration.cs
@@ -20,6 +20,7 @@ using System.Collections.Generic;
 using System.Globalization;
 using System.IO;
 using System.Linq;
+using System.Text;
 using System.Threading.Tasks;
 using System.Windows;
 using static System.FormattableString;
@@ -476,7 +477,7 @@ namespace AccessibilityInsights.Extensions.AzureDevOps
         internal Uri CreateIssuePreview(string projectName, string teamName, IReadOnlyDictionary<AzureDevOpsField, string> AzureDevOpsFieldPairs)
         {
             var escaped = from pair in AzureDevOpsFieldPairs
-                          select Uri.EscapeDataString($"[{pair.Key.ToApiString()}]") + "=" + Uri.EscapeDataString(pair.Value);
+                          select EscapeForUrl($"[{pair.Key.ToApiString()}]") + "=" + EscapeForUrl(pair.Value);
 
             // Uri.EscapeDataString converts space to %20, but it seems safe for us to instead use a "+" for the contents
             //  of the issue description, which will be interpreted as a space in the browser. This saves us characters, so we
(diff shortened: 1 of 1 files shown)


src/AccessibilityInsights.Extensions.AzureDevOps/AzureDevOpsIntegration.cs
```diff
@@ -486,6 +487,29 @@ internal Uri CreateIssuePreview(string projectName, string teamName, IReadOnlyDi
...
+        /// <returns></returns>
+        internal static string EscapeForUrl(string str)
+        {
+            var sb = new StringBuilder();
+            foreach (var c in str)
+            {
+                // characters such as the middle dot (ascii 183) can 
+                // cause issues during navigation, so we only include
+                // a basic set of characters when building the URL
+                if (c >= 1 && c <= 127)
```
DaveTryon: >c >= 1 [](start = 20, length = 6)

Do we need the lower bounds check? I'd expect it to be unsigned and maybe even support null

src/AccessibilityInsights.Extensions.AzureDevOps/AzureDevOpsIntegration.cs
```diff
@@ -486,6 +487,29 @@ internal Uri CreateIssuePreview(string projectName, string teamName, IReadOnlyDi
...
             return result;
         }
+
+        /// <summary>
+        /// Retains characters within ascii range 1-127 and returns
+        /// an escaped representation of the string
+        /// </summary>
+        /// <param name="str"></param>
+        /// <returns></returns>
+        internal static string EscapeForUrl(string str)
```
DaveTryon: >EscapeForUrl [](start = 31, length = 12)

There may also be some value in having a Dictionary of substitutions (non-breaking space to space, middle dot to a star, etc.) to try to preserve more natural formatting

jalkire: This seems like something we would want at least for the characters we use internally, like the ● bullet

src/AccessibilityInsights.Extensions.AzureDevOps/AzureDevOpsIntegration.cs
```diff
@@ -486,6 +487,29 @@ internal Uri CreateIssuePreview(string projectName, string teamName, IReadOnlyDi
...
+            var sb = new StringBuilder();
+            foreach (var c in str)
+            {
+                // characters such as the middle dot (ascii 183) can 
+                // cause issues during navigation, so we only include
+                // a basic set of characters when building the URL
+                if (c >= 1 && c <= 127)
+                {
+                    sb.Append(c);
+                }
```
DaveTryon: >} [](start = 16, length = 1)

Not sure how I feel about completely removing characters. Would it make sense to have a default substitution character to represent a character that was removed? Since most cases here are likely to be errors in our code or data files, having some sort of way to detect this case would also be interesting.

src/AccessibilityInsights.Extensions.AzureDevOps/AzureDevOpsIntegration.cs
karanbirsingh: I think it's an interesting idea, although these strings do get truncated. For the bullet case, it's the difference between `patterns: Modify a b c...see more in the a11y test file` and `patterns: {substitution} Modify a b c...see more in the a11y test file`. I wonder whether it's more complexity than we need

PR #316, opened 2019-04-17
Report more exceptions

DaveTryon:
#### Describe the change
At one point, we had telemetry for Exceptions, then we removed it. It was recently partially restored, but several places that used to report Exceptions were skipped, and lots of code was added without ever trying to report Exception. This change is basically applies the same pattern over and over:

1. Find a catch block
2. If the catch block eats the Exception, call ReportException to get it into the data stream
3. Add the necessary using block, depending on where the code lives--either Axe.Windows.Telemetry or AccessibilityInsights.SharedUx.Telemetry.

There are a few cases where we don't call ReportException:
- Code in CommonUxComponents can't log without creating a circular dependency or splitting telemetry into a separate assembly (the separate assembly is the correct long-term answer)
- Code in the Logger doesn't report in some cases that would introduce a potential to throw unexpectedly or call recursively
- Some code at shutdown doesn't call because it wouldn't get sent anyway
- Exceptions caused by user action (user typed in bad data or moved the magnifier to the edge of the screen), and that we literally can't do anything about.
- Exceptions where we literally can't do anything (like during Container's construction)

It's probable that some of the reported Exceptions will be non-actionable and will eventually get removed. The goal here is to get _some_ sort of signal that can be used to assess product stability (and that might feed into the promotion decisions of a given build.

#### PR checklist

- [ ] Includes UI changes?


src/AccessibilityInsights.Core/Fingerprint/OutputFileLocation.cs
```diff
@@ -82,6 +83,8 @@ public bool Open()
                 {
                     throw;
                 }
+
+                e.ReportException();
```
jalkire: Is there a particular reason to only report the exceptions that weren't filtered out above?

src/AccessibilityInsights.SharedUx/Highlighting/ClearOverlayDriver.cs
```diff
@@ -63,17 +64,19 @@ public static void BringMainWindowOfPOIElementToFront()
...
-            } 
-            catch (InvalidOperationException)
+            catch (Exception e)
             {
+                if (e is ArgumentException ||
+                    e is COMException ||
+                    e is NullReferenceException ||
+                    e is InvalidOperationException)
+                {
+                    e.ReportException();
```
jalkire: Same question as before about only reporting the exceptions we aren't throwing

src/AccessibilityInsights.Automation/ExecutionWrapper.cs
```diff
@@ -25,6 +25,7 @@ internal static T ExecuteCommand<T>(Func<T> command, Func<string, T> errorFactor
                 }
                 catch (Exception ex)
                 {
+                    // No need to report this Exception, since telemetry is disabled in automation
```
RobGallo: Disabled as in not possible or not desirable?

src/AccessibilityInsights.Core/Fingerprint/OutputFileLocation.cs
DaveTryon: We only want to report each Exception once, and only when it gets eaten. If we were to call ReportException on the Exceptions that are being rethrown, they would be multiply-counted.

src/AccessibilityInsights.SharedUx/Highlighting/ClearOverlayDriver.cs
DaveTryon: Same as above. We only want to call ReportExceptions when we eat the Exception, and want to avoid multiple counts

src/AccessibilityInsights.Automation/ExecutionWrapper.cs
DaveTryon: Updated the comment to say "telemetry is not available in automation"

src/AccessibilityInsights.Automation/ExecutionWrapper.cs
```diff
@@ -25,7 +25,7 @@ internal static T ExecuteCommand<T>(Func<T> command, Func<string, T> errorFactor
                 }
                 catch (Exception ex)
                 {
-                    // No need to report this Exception, since telemetry is disabled in automation
+                    // No need to report this Exception, since telemetry not available in automation
```
RobGallo: Missing "is".

DaveTryon: fixed

Issue #317, opened 2019-04-18
[BUG] AI-Win fails to start on some machines


DaveTryon:
**Describe the bug**
When starting AccessibilityInsights on some machines, the app hangs, then crashes. Looking in the event viewer, it's throwing an Exception:

Application: AccessibilityInsights.exe
Framework Version: v4.0.30319
Description: The process was terminated due to an unhandled exception.
Exception Info: System.IO.FileNotFoundException
   at AccessibilityInsights.SetupLibrary.MsiUtilities.GetInstalledProductVersion(AccessibilityInsights.SetupLibrary.IExceptionReporter)
   at AccessibilityInsights.SetupLibrary.MsiUtilities.GetReleaseNotesUri(AccessibilityInsights.SetupLibrary.IExceptionReporter)
   at AccessibilityInsights.Modes.StartUpModeControl..ctor()

Upon further investigation, the problem is that the Microsoft.Deployment.WindowsInstaller assembly needs to be in the app's main folder instead of installed with the AutoUpdate extension. 

**To Reproduce**
1. Start AIWin on a susceptible machine (finding the susceptible machine is the hard part)

**Expected behavior**
App should start normally

**Actual behavior**
Splash screen displays, then the app just... stops. The Exception shows up in the Event Viewer.

**Desktop (please complete the following information):**
 - OS: Windows 10 1903
 - Accessibility Insights for Windows Version: 1.1.835.1

DaveTryon:
Confirmed fixed in 1..1.0837.02

PR #318, opened 2019-04-18
Move Microsoft.Deployment.Windows.Installer.dll up to main app

DaveTryon:
#### Describe the change
Build 1.1.835.1 introduced a change that caused the app to hang when starting on a specific machine. The root cause was that Microsoft.Deployment.WindowsInstaller.dll wasn't being found because it was installed as part of an extension instead of part of the app. Most machines successfully loaded the assembly as part of the AutoUpdate code--it's not clear why this machine behaves differently, but it's 100% reproducible on this machine. This change moves the assembly into the same folder as the main app so it will reliably be found when needed, either by the main app or by the AutoUpdate code.

Tested in 2 stages:
1. On an affected machine, manually moved Microsoft.Deployment.WindowsInstaller.dll to the app folder and successfully launched the app
2. Built a private MSI, installed it on the affected machine, and successfully launched the app

#### PR checklist

- [x] Does this address an existing issue? If yes, Issue# - #317
- [ ] Includes UI changes?

### DIFF
--- a/src/MSI/Product.wxs
+++ b/src/MSI/Product.wxs
@@ -67,6 +67,7 @@ Licensed under the MIT license. See LICENSE file in the project root for full li
               <File Source="..\AccessibilityInsights\bin\Release\AccessibilityInsights.SharedUx.dll" />
               <File Source="..\AccessibilityInsights\bin\Release\AccessibilityInsights.SharedUx.dll.config" />
               <File Source="..\AccessibilityInsights\bin\Release\Axe.Windows.Win32.dll" />
+              <File Source="..\AccessibilityInsights\bin\Release\Microsoft.Deployment.WindowsInstaller.dll" />
               <File Source="..\AccessibilityInsights\bin\Release\AccessibilityInsights.exe.config" />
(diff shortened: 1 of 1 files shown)


src/MSI/Product.wxs
```diff
@@ -100,8 +101,6 @@ Licensed under the MIT license. See LICENSE file in the project root for full li
 
                    <!-- AutoUpdate Extension Assemblies -->
                    <File Source="..\AccessibilityInsights.Extensions.GitHubAutoUpdate\bin\Release\AccessibilityInsights.Extensions.GitHubAutoUpdate.dll" Id="autoupdate_extension"/>
-                   <File Source="..\AccessibilityInsights.Extensions.GitHubAutoUpdate\bin\Release\Microsoft.Deployment.WindowsInstaller.dll" Id="windowsinstaller_extension"/>
-                   <File Source="..\AccessibilityInsights.Extensions.GitHubAutoUpdate\bin\Release\Newtonsoft.Json.dll" Id="newtonsoft_json_extension"/>
```
DaveTryon: NewtonSoft is already in the app folder, so it's redundant to have it here

PR #323, opened 2019-04-19
Substituting characters to be more friendly to bug-url generation

karanbirsingh:
This PR follows up on #313 in which we limited the set of allowable characters in our bug URLs to respond to odd behavior when the user is redirected to login - as suggested by @DaveTryon , instead of removing characters outside our allowable range, we now replace them with preconfigured alternatives or a '?'.

### DIFF
--- a/src/AccessibilityInsights.Extensions.AzureDevOps/AzureDevOpsIntegration.cs
+++ b/src/AccessibilityInsights.Extensions.AzureDevOps/AzureDevOpsIntegration.cs
@@ -491,27 +491,43 @@ namespace AccessibilityInsights.Extensions.AzureDevOps
         }
 
         /// <summary>
-        /// Retains characters within ascii range below 127 and returns
-        /// an escaped representation of the string
+        /// Substitutions for characters when creating the bug URL
+        /// </summary>
+        internal static IReadOnlyDictionary<char, char> CharacterSubstitutions = new Dictionary<char, char>()
+        {
+            { '·', '-' }, // middle dot (183) to dash (45)
+        };
+
+        /// <summary>
+        /// replaces characters if substitution defined and
+        /// returns an escaped representation of the string
         /// </summary>
         /// <param name="str"></param>
         /// <returns></returns>
         internal static string EscapeForUrl(string str)
         {
+            // characters such as the middle dot (ascii 183) can 
+            // cause issues during navigation, so we only allow
+            // a basic set of characters when building the URL
             var sb = new StringBuilder();
             foreach (var c in str)
             {
-                // characters such as the middle dot (ascii 183) can 
-                // cause issues during navigation, so we only include
-                // a basic set of characters when building the URL
-                if (c <= 127)
+                if (CharacterSubstitutions.TryGetValue(c, out char replaced))
+                {
+                    sb.Append(replaced);
+                }
+                else if (c <= 127)
                 {
                     sb.Append(c);
+                }
+                else
+                {
+                    sb.Append('?');
                 }
             }
 
             return Uri.EscapeDataString(sb.ToString());
         }
-#endregion // IssueFiling
+        #endregion // IssueFiling
     }
 }
--- a/src/AccessibilityInsights.Extensions.AzureDevOps/AzureDevOpsIntegration.cs
+++ b/src/AccessibilityInsights.Extensions.AzureDevOps/AzureDevOpsIntegration.cs
@@ -495,7 +495,7 @@ namespace AccessibilityInsights.Extensions.AzureDevOps
         /// </summary>
         internal static IReadOnlyDictionary<char, char> CharacterSubstitutions = new Dictionary<char, char>()
         {
-            { '·', '-' }, // middle dot (183) to dash (45)
+            { '\u00b7', '-' }, // middle dot (183) to dash (45)
         };
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


src/AccessibilityInsights.Extensions.AzureDevOps/AzureDevOpsIntegration.cs
```diff
@@ -491,27 +491,43 @@ internal Uri CreateIssuePreview(string projectName, string teamName, IReadOnlyDi
...
+                {
+                    sb.Append(replaced);
+                }
+                else if (c <= 127)
                 {
                     sb.Append(c);
                 }
+                else
+                {
+                    sb.Append('?');
```
DaveTryon: >? [](start = 31, length = 1)

This is a distinct improvement, but my gut tells me that this will eventually cause pain for international users, where characters > 127 are commonplace. That said, I'm ok with taking the James Shore approach of maximizing the work that we don't do

src/AccessibilityInsights.Extensions.AzureDevOps/AzureDevOpsIntegration.cs
```diff
@@ -491,27 +491,43 @@ internal Uri CreateIssuePreview(string projectName, string teamName, IReadOnlyDi
         }
 
         /// <summary>
-        /// Retains characters within ascii range below 127 and returns
-        /// an escaped representation of the string
+        /// Substitutions for characters when creating the bug URL
+        /// </summary>
+        internal static IReadOnlyDictionary<char, char> CharacterSubstitutions = new Dictionary<char, char>()
+        {
+            { '�', '-' }, // middle dot (183) to dash (45)
```
DaveTryon: Let's be very explicit how the unicode values are represented in code (especially since this file doesn't have a UTF encoding header)
```suggestion
            { '\B7', '-' }, // middle dot (183) to dash (45)
```

src/AccessibilityInsights.Extensions.AzureDevOps/AzureDevOpsIntegration.cs
```diff
@@ -491,27 +491,43 @@ internal Uri CreateIssuePreview(string projectName, string teamName, IReadOnlyDi
...
+
+        /// <summary>
+        /// replaces characters if substitution defined and
+        /// returns an escaped representation of the string
         /// </summary>
         /// <param name="str"></param>
         /// <returns></returns>
         internal static string EscapeForUrl(string str)
         {
+            // characters such as the middle dot (ascii 183) can
```
DaveTryon: >characters such as the middle dot (ascii 183) [](start = 15, length = 45)

I don't think I'd mention the 183 value here, since it's handled in the table and this code just follows the table

src/AccessibilityInsights.Extensions.AzureDevOps/AzureDevOpsIntegration.cs
```diff
@@ -491,27 +491,43 @@ internal Uri CreateIssuePreview(string projectName, string teamName, IReadOnlyDi
...
+            // cause issues during navigation, so we only allow
+            // a basic set of characters when building the URL
             var sb = new StringBuilder();
             foreach (var c in str)
             {
-                // characters such as the middle dot (ascii 183) can 
-                // cause issues during navigation, so we only include
-                // a basic set of characters when building the URL
-                if (c <= 127)
+                if (CharacterSubstitutions.TryGetValue(c, out char replaced))
```
DaveTryon: >replaced [](start = 67, length = 8)

I'd probably call this substitute, but it's not a big deal

src/AccessibilityInsights.Extensions.AzureDevOps/AzureDevOpsIntegration.cs
karanbirsingh:
```suggestion
            // characters such as the middle dot can 
```

karanbirsingh: updated

src/AccessibilityInsights.Extensions.AzureDevOps/AzureDevOpsIntegration.cs
```diff
@@ -491,27 +491,43 @@ internal Uri CreateIssuePreview(string projectName, string teamName, IReadOnlyDi
...
+                {
+                    sb.Append(replaced);
+                }
+                else if (c <= 127)
                 {
                     sb.Append(c);
                 }
+                else
+                {
+                    sb.Append('?');
```
jalkire: I feel like we might want something other than a common character (?) here to make it clear that we modified the string. I would suggest �, but that won't work for obvious reasons...maybe some kind of enclosed ? like <?> ?

src/AccessibilityInsights.Extensions.AzureDevOps/AzureDevOpsIntegration.cs
karanbirsingh: If it's similar to you either way, I prefer the noun form; plan to leave as-is

DaveTryon: "replaced" isn't a noun. "replacement" or "substitution" are nouns (and "substitution" has the benefit of matching the collection name)

---
In reply to: [277098982](https://github.com/Microsoft/accessibility-insights-windows/pull/323#discussion_r277098982) [](ancestors = 277098982)

jalkire: substitute is commonly used as a noun, as well

src/AccessibilityInsights.Extensions.AzureDevOps/AzureDevOpsIntegration.cs
karanbirsingh: I was hoping to keep the replacement a single character; if you feel strongly about this I'm happy to update. IMHO the ? seems sufficient for the case we're trying to handle - lmk

PR #324, opened 2019-04-19
Add documentation to public members of A11yElement

RobGallo:
#### Describe the change

This content needs to be created so we can generate xml documentation files to accompany the axe-windows NuGet package. This is only part of the work needed to get Axe.Windows.Core ready, but it's a start.

#### PR checklist

- [ ] Includes UI changes?

### DIFF
--- a/src/AccessibilityInsights.Core/Bases/A11yElement.cs
+++ b/src/AccessibilityInsights.Core/Bases/A11yElement.cs
@@ -497,7 +497,7 @@ namespace Axe.Windows.Core.Bases
         }
 
         /// <summary>
-        /// Finds the first decendent element matching the given condition.
+        /// Finds the first descendant element matching the given condition.
         /// </summary>
         /// <param name="condition">a function that returns true if the given descendant element meets a set of criteria</param>
         /// <returns>an <see cref="IA11yElement"/> object representing the matching descendant if one exists; otherwise, false</returns>
--- a/src/AccessibilityInsights.Core/Bases/A11yElement.cs
+++ b/src/AccessibilityInsights.Core/Bases/A11yElement.cs
@@ -86,7 +86,7 @@ namespace Axe.Windows.Core.Bases
         }
 
         /// <summary>
-        /// a text description of a keystroke which activates the element; usually a single key; only  valid within  a limited scope such as menus
+        /// a text description of a keystroke which activates the element; usually a single key; only valid within a limited scope such as menus
         /// </summary>
         [JsonIgnore]
         public string AccessKey
--- a/src/AccessibilityInsights.Core/Bases/A11yElement.cs
+++ b/src/AccessibilityInsights.Core/Bases/A11yElement.cs
@@ -24,6 +24,9 @@ namespace Axe.Windows.Core.Bases
     {
         private string _ProcessName = null;
 
+        /// <summary>
+        /// The label or primary text identifier for the element
+        /// </summary>
         [JsonIgnore]
         public string Name
         {
@@ -33,6 +36,9 @@ namespace Axe.Windows.Core.Bases
             }
         }
 
+        /// <summary>
+        /// The control type identifier for the element; see <see cref="Axe.Windows.Core.Types.ControlType"/>
+        /// </summary>
         [JsonIgnore]
         public int ControlTypeId
         {
@@ -43,6 +49,9 @@ namespace Axe.Windows.Core.Bases
             }
         }
 
+        /// <summary>
+        /// a localized text representation of <see cref="ControlTypeId"/>
+        /// </summary>
         [JsonIgnore]
         public string LocalizedControlType
         {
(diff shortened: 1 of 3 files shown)


src/AccessibilityInsights.Core/Bases/A11yElement.cs
```diff
@@ -70,6 +85,9 @@ public string AcceleratorKey
             }
         }
 
+        /// <summary>
+        /// a text description of a keystroke which activates the element; usually a single key; only  valid within  a limited scope such as menus
```
DaveTryon: Extra spaces in this line

```suggestion
        /// a text description of a keystroke which activates the element; usually a single key; only valid within a limited scope such as menus
```

src/AccessibilityInsights.Core/Bases/A11yElement.cs
```diff
@@ -88,6 +109,9 @@ public string Culture
             }
         }
 
+        /// <summary>
+        /// a permanent unique identifier for the specific element, or sometimes the type of element
```
DaveTryon: >permanent unique identifier [](start = 14, length = 27)

Is it worth calling out that the uniqueness is not global, but only unique within siblings in its container?

DaveTryon:
General comment: I'd suggest adding a link to the official docs within each comment. All of the links are accessible from [https://docs.microsoft.com/en-us/dotnet/framework/ui-automation/ui-automation-properties-overview](https://docs.microsoft.com/en-us/dotnet/framework/ui-automation/ui-automation-properties-overview)

src/AccessibilityInsights.Core/Bases/A11yElement.cs
```diff
@@ -336,6 +423,9 @@ public T GetPlatformPropertyValue<T>(int propertyId)
             return property.Value;
         }
 
+        /// <summary>
+        /// the index of an item within a set
```
DaveTryon: >set [](start = 42, length = 3)

Seems like "set" would be clearer with some context. I'm assuming that it refers to the siblings in the same container? Same comment applies to SizeOfSet

src/AccessibilityInsights.Core/Bases/A11yElement.cs
```diff
@@ -358,6 +451,9 @@ public int SizeOfSet
             }
         }
 
+        /// <summary>
+        /// whether the element is resizeable
```
DaveTryon: "resizeable" is an accepted alternative spelling, but "resizable" (with no 'e' after the z) is the more common spelling

```suggestion
        /// whether the element is resizable
```

src/AccessibilityInsights.Core/Bases/A11yElement.cs
```diff
@@ -393,6 +496,11 @@ public IA11yElement GetFirstChild()
             return Children[0];
         }
 
+        /// <summary>
+        /// Finds the first decendent element matching the given condition.
```
DaveTryon: Spelling: Should be "descendant" (has an 's' in the middle, ends in "ant" instead of "ent"

Usage Note from https://www.yourdictionary.com/descendent#americanheritage:

The adjective, "descending from a biological ancestor", may be spelt either with an a or with an e in the final syllable (see descendant). However the noun descendant, "one who is the progeny of someone", may be spelt only with an a.

```suggestion
        /// Finds the first descendant element matching the given condition.
```

src/AccessibilityInsights.Core/Bases/A11yElement.cs
```diff
@@ -393,6 +496,11 @@ public IA11yElement GetFirstChild()
...
+        /// </summary>
+        /// <param name="condition">a function that returns true if the given descendant element meets a set of criteria</param>
+        /// <returns>an <see cref="IA11yElement"/> object representing the matching descendant if one exists; otherwise, false</returns>
         public A11yElement FindDescendent(Func<A11yElement, Boolean> condition)
```
DaveTryon: Spelling: Should end in "ant", not "ent"

```suggestion
        public A11yElement FindDescendant(Func<A11yElement, Boolean> condition)
```

src/AccessibilityInsights.Core/Bases/A11yElement.cs
RobGallo:
```suggestion
        /// a permanent unique identifier among siblings for the specific element, or sometimes the type of element
```

RobGallo:
> General comment: The descriptions here are largely copied from the official docs. Should we add a link to the official docs within each comment? All of the links are accessible from https://docs.microsoft.com/en-us/dotnet/framework/ui-automation/ui-automation-properties-overview

Some may be similar, but they are not all the same. I curated the text to make the comments mnore generic than just UIA, since the element is intended to be generic.

src/AccessibilityInsights.Core/Bases/A11yElement.cs
RobGallo: I don't think we can be as specific as siblings. I also think we can't explain everything in our code documentation. I think we should provide a practical hint and rely on external docs to provide more info.

PR #329, opened 2019-04-22
Remove CCAControlTypeFilters from Axe.Windows.Rules

RobGallo:
#### Describe the change

The class CCAControlTypeFilters was unnecessary because the integer values of the UIA control types could more directly be enumerated by the ControlType class. In addition, CCAControlTypeFilters was removed because it was made part of the Rules assembly, whose responsibility is only to run automated rules -- the Rules assembly should never know anything about the color contrast analyzer, or report any type information other than that provided by the rules.

#### PR checklist

- [ ] Includes UI changes?


DaveTryon:
>             if (this.Dic.ContainsKey(id))

(aside): this is a perfect place for a TryGetValue

---
Refers to: src/AccessibilityInsights.Core/Types/TypeBase.cs:91 in 43e81b9. [](commit_id = 43e81b97df67229ad5e7c62435737c438a754350, deletion_comment = False)

src/AccessibilityInsights.Core/Types/TypeBase.cs
```diff
@@ -115,5 +115,10 @@ public bool Exists(int id)
         {
             return this.Dic.ContainsKey(id);
         }
+
+        /// <summary>
+        /// the values of the types contained in the inheriting class
```
DaveTryon: I'd omit "values of the ", like this:

```suggestion
        /// the types contained in the inheriting class
```

src/AccessibilityInsights.Core/Types/TypeBase.cs
```diff
@@ -115,5 +115,10 @@ public bool Exists(int id)
...
+        /// </summary>
+        public IEnumerable<int> Values => Dic.Keys;
```
DaveTryon: Should this be called Types?

```suggestion
        public IEnumerable<int> Types => Dic.Keys;
```

RobGallo: I don't think so. I consider this to be the values of the types.

PR #330, opened 2019-04-22
Production release 1.1.0837.02

DaveTryon:
#### Describe the change
Promote Release 1.1.0837.02 to Production. This copies the Insider versions of *release_info.json* and *release_notes.md* to Production, with an updated date in the release notes. Expectation is that we'll release on Wednesday, April 24th.

This also removes the bitmaps that were added to support the production release notes (the bitmaps were added at the last moment to work around issues with GitHub pages, and are not part of the long-term plan around release notes)


DaveTryon:
We'll merge this on 4/24

Issue #326, opened 2019-04-19
[BUG] AI-Win telemetry reports wrong OS versions


DaveTryon:
**Describe the bug**
The AI-Win telemetry reports only 2 values for the OS. It's either **Microsoft Windows NT 6.1.7601 Service Pack 1** for Windows 7 SP1 (our lowest supported platform) or **Microsoft Windows NT 6.2.9200.0** for everything else, meaning that we have no meaningful OS version data in our telemetry.

**The fix**
The fix is to uncomment the boilerplate **supportedOS** sections in the manifest files. That will cause the OS to report the correct OS info without any code change.

**The risk**
By uncommenting the **supportedOS** pieces in the manifest files, we also change some of our high DPI behavior. There's a good overview of this at https://walbourn.github.io/manifest-madness/. We had a ***lot*** of trouble getting high DPI support right, so if we enable this, we'll need to plan on testing (and maybe fixing high DPI issues) in at least the following areas:
- Highlighter overlay (live and in snapshot)
- Tabstop overlay
- Position and rendering of the snapshot button
- Proper activation of a scan by clicking the snapshot button

It's fine if we choose not to fix this, as long as we're also willing to live with the lack of usable OS info in the telemetry.

neocln:
@DaveTryon as we discussed, please investigate it.

PR #332, opened 2019-04-22
Add registry based version info to telemetry

DaveTryon:
#### Describe the change
Our telemetry pipeline (Application Insights) reports OS version data fetched via an API that intentionally lies about windows versions unless certain manifest flags are set. Unfortunately, those same flags also alter high DPI behavior, which would be a high impact change. Workaround is to add our own windows version information to telemetry, using Registry keys that work at least as far as back as Win7, and that are likely to be around for a very long time. This change may complicate the work to split the Win32 assembly into 2 pieces, but since that work hasn't yet started, there should be no conflict in making this change now.

#### PR checklist

- [x] Does this address an existing issue? If yes, Issue# - #326
- [ ] Includes UI changes?


src/AccessibilityInsights/AccessibilityInsights.csproj
```diff
@@ -262,6 +262,10 @@
       <Project>{eee9cd52-ca10-41ae-ac6e-98ae4bbdfa2d}</Project>
       <Name>CommonUxComponents</Name>
     </ProjectReference>
+    <ProjectReference Include="..\AccessibilityInsights.Extensions.Telemetry\Extensions.Telemetry.csproj">
```
DaveTryon: Durp. This was supposed to stay local and not get pushed

src/AccessibilityInsights/MainWindow.xaml.cs
```diff
@@ -197,7 +197,13 @@ private static void InitTelemetry()
             Logger.AddOrUpdateContextProperty(TelemetryProperty.Version, VersionTools.GetAppVersion());
             Logger.AddOrUpdateContextProperty(TelemetryProperty.AppSessionID, Guid.NewGuid().ToString());
             Logger.AddOrUpdateContextProperty(TelemetryProperty.SessionType, "Desktop");
-            Logger.AddOrUpdateContextProperty(TelemetryProperty.ReleaseChannel, ConfigurationManager.GetDefaultInstance().AppConfig.ReleaseChannel.ToString());
+            Logger.AddOrUpdateContextProperty(TelemetryProperty.WindowsVersion, ConfigurationManager.GetDefaultInstance().AppConfig.ReleaseChannel.ToString());
```
jalkire: Looks like TelemetryProperty.WindowsVersion and TelemetryProperty.ReleaseChannel got flipped between here and line 205
```suggestion
            Logger.AddOrUpdateContextProperty(TelemetryProperty.ReleaseChannel, ConfigurationManager.GetDefaultInstance().AppConfig.ReleaseChannel.ToString());
```

src/AccessibilityInsights/MainWindow.xaml.cs
```diff
@@ -197,7 +197,13 @@ private static void InitTelemetry()
...
+
+            string windowsVersion = NativeMethods.GetCurrentWindowsVersionForTelemetry();
+            if (!string.IsNullOrEmpty(windowsVersion))
+            {
+                Logger.AddOrUpdateContextProperty(TelemetryProperty.ReleaseChannel, windowsVersion);
```
jalkire:
```suggestion
                Logger.AddOrUpdateContextProperty(TelemetryProperty.WindowsVersion, windowsVersion);
```

DaveTryon: Oops! I changed the wrong one

src/AccessibilityInsights/MainWindow.xaml.cs
```diff
@@ -198,6 +198,12 @@ private static void InitTelemetry()
             Logger.AddOrUpdateContextProperty(TelemetryProperty.AppSessionID, Guid.NewGuid().ToString());
             Logger.AddOrUpdateContextProperty(TelemetryProperty.SessionType, "Desktop");
             Logger.AddOrUpdateContextProperty(TelemetryProperty.ReleaseChannel, ConfigurationManager.GetDefaultInstance().AppConfig.ReleaseChannel.ToString());
+
+            string windowsVersion = NativeMethods.GetCurrentWindowsVersionForTelemetry();
+            if (!string.IsNullOrEmpty(windowsVersion))
```
RobGallo: Should we log something if we can't get the version?

src/AccessibilityInsights.Win32Tests/Win32ApisTests.cs
```diff
@@ -263,5 +263,47 @@ public void IsWindowsRS5OrLater_BuildValueIsGreaterThanTarget_ReturnsTrue()
...
+        [TestMethod]
+        public void GetCurrentWindowsVersionForTelemetry_CurrentVersionDoesNotExist_ReturnsNull()
+        {
+            using (ShimsContext.Create())
+            {
+                List<string> returnValues = new List<string> { "" };
+                int index = 0;
+                ShimRegistry.GetValueStringStringObject = (_, __, def) => (++index > returnValues.Count ? def : returnValues[index - 1]);
+
+                Assert.IsNull(NativeMethods.GetCurrentWindowsVersionForTelemetry());
```
RobGallo: Not that I expect any change in this PR, but isn't the NativeMethods class supposed to only be for pInvoke signatures of native methods?

DaveTryon:
@neocln : The AppInsights code already uses Environment,OSVersion internally. This is the API that misreports the OS unless the manifest file is modified. That said, I like the idea of using the AppInsights-provided flag. Having both old and new formats in the same field will generate some noise in the short run, but it will drop off over time.

DaveTryon:
@neocln: Your proposal will cause some bumps in the short-term telemetry, as we will have multiple forms of data in the same string. That said, we can probably handle that change. I don't want to reference Win32 from the extension, so I'll just add the registry code directly into the extension. That also removes the need for touching Win32.

src/AccessibilityInsights.Extensions.Telemetry/Properties/AssemblyInfo.cs
```diff
@@ -25,7 +25,7 @@
 [assembly: Guid("d17e1749-28fe-44b9-a9f6-af8985124fe4")]
 
 #if ENABLE_SIGNING
-[assembly: InternalsVisibleTo("Axe.Windows.DesktopTests,PublicKey=002400000480000094000000060200000024000052534131000400000100010007d1fa57c4aed9f0a32e84aa0faefd0de9e8fd6aec8f87fb03766c834c99921eb23be79ad9d5dcc1dd9ad236132102900b723cf980957fc4e177108fc607774f29e8320e92ea05ece4e821c0a5efe8f1645c4c0c93c1ab99285d622caa652c1dfad63d745d6f2de5f17e5eaf0fc4963d261c8a12436518206dc093344d5ad293")]
```
DaveTryon: DesktopTests doesn't actually use these internals. This was probably a remnant of a pre-refactor dependency

DaveTryon:
Here's an event uploaded with the change--note that the client_OS field now says **6.3.18362** (the correct value for 19H1) instead of **Microsoft Windows NT 6.2.9200.0**:

image

src/AccessibilityInsights/MainWindow.xaml.cs
DaveTryon: We'll know if a problem occurs, so I don't think it's necessary

---
In reply to: [277715682](https://github.com/Microsoft/accessibility-insights-windows/pull/332#discussion_r277715682) [](ancestors = 277715682)

src/AccessibilityInsights.Win32Tests/Win32ApisTests.cs
DaveTryon: No longer an issue in the revised format

---
In reply to: [277719578](https://github.com/Microsoft/accessibility-insights-windows/pull/332#discussion_r277719578) [](ancestors = 277719578)

src/AccessibilityInsights.Win32/Win32Helper.cs
```diff
@@ -193,6 +193,28 @@ internal static bool IsWindowsRS5OrLater()
             return IsAtLeastWin10WithSpecificBuild(17713); // Build 17713 is confirmed in the RS5 range
         }
 
+        /// <summary>
+        /// Get the current Windows version (intended to be used for telemetry). If trying to
```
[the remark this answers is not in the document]
DaveTryon: Telemetry method is now in telemetry extension

src/AccessibilityInsights.Extensions.Telemetry/OSHelpers.cs
```diff
@@ -0,0 +1,35 @@
...
+using Microsoft.Win32;
+
+namespace AccessibilityInsights.Extensions.Telemetry
+{
+    /// <summary>
+    /// Helpers to extract data from the OS
+    /// </summary>
+    internal static class OSHelpers
+    {
+        /// <summary>
```
RobGallo: I still wish there were a remark as to why this is suitable for telemetry but not for features. There's no "why".

src/AccessibilityInsights.Extensions.TelemetryTests/OSHelpersUnitTests.cs
```diff
@@ -0,0 +1,57 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
```
RobGallo: Are we calling test files *UnitTests.cs" or "*Tests.cs" now?

src/AccessibilityInsights.Extensions.TelemetryTests/Extensions.TelemetryTests.csproj
```diff
@@ -64,6 +68,7 @@
   <ItemGroup>
     <Compile Include="AITelemetryTests.cs" Condition="$(FAKES_SUPPORTED) == 1" />
     <Compile Include="Properties\AssemblyInfo.cs" />
+    <Compile Include="OSHelpersUnitTests.cs" />
```
RobGallo: I don't see us excluding the fakes based on the environment variable anywhere. We can do this here at the file level, or at the function level in the file itself.

DaveTryon: This was added in the most recent iteration

src/AccessibilityInsights.Extensions.TelemetryTests/OSHelpersUnitTests.cs
DaveTryon: The Unit Tests documentation says *UnitTests.cs. The code to run tests looks for *Tests.dll

src/AccessibilityInsights.Extensions.Telemetry/OSHelpers.cs
DaveTryon: There are countless blog pages on the problems with the various forms of GetVersion. I don't see a lot of value in adding yet another commentary.

Issue #336, opened 2019-04-23
[BUG] Update/switch channel UAC dialog appears behind app


jalkire:
**Describe the bug**
If AI-Win attempts to install a new version of itself, whether from auto-update or from channel change, the UAC dialog is obscurred.

**To Reproduce**
1. Open AI-Win to the center of your default monitor 
2. Go to settings
3. Ensure AI-Win is always on top before proceeding
4. Change channel
5. Save and close

**Expected behavior**
The UAC prompt should be on top of AI-Win

**Actual behavior**
The UAC prompt is behind AI-Win

**Screenshots or .GIF**
Updating:
![image](https://user-images.githubusercontent.com/4615491/56621914-21c43880-65e3-11e9-8413-c7864bca3137.png)

Changing channel:
![image](https://user-images.githubusercontent.com/4615491/56621978-47e9d880-65e3-11e9-9ed4-9c31f90af28f.png)

**Desktop (please complete the following information):**
 - OS:  Windows 10 1809

neocln:
@jalkire could you take care of this? we may need to set the main windows not to be TopMost when we get into update mode.

PR #337, opened 2019-04-24
Stop main window from being topmost while UAC prompt is displayed

jalkire:
#### Describe the change
Addressing #336: If AI-Win attempts to install a new version of itself, whether from auto-update or from channel change, the UAC dialog is obscurred.
This change causes AI-Win to backup its current topmost state, set topmost to false, run the update/channel switch, and then restore its topmost state (if AI-Win is still open). This is similar to what we do for the ADO sign in dialog scenario.

#### PR checklist

- [ ] Does this address an existing issue? If yes, Issue #336
- [ ] Includes UI changes?

See bug for before screenshots. 

UAC dialog is on top of AI-Win:
image
image

### DIFF
--- a/src/AccessibilityInsights/MainWindowHelpers/AutoUpdate.cs
+++ b/src/AccessibilityInsights/MainWindowHelpers/AutoUpdate.cs
@@ -125,7 +125,7 @@ namespace AccessibilityInsights
             UpdateResult result = UpdateResult.Unknown;
 
             // The UAC prompt from the version switcher will appear behind the main window
-            // if it is topmost, so we store, change, and restore the value in this method.
+            // if it is topmost, we store, change, and restore the value in this method.
             bool oldTopMost = Topmost;
 
             try
--- a/src/AccessibilityInsights/MainWindowHelpers/AutoUpdate.cs
+++ b/src/AccessibilityInsights/MainWindowHelpers/AutoUpdate.cs
@@ -124,8 +124,9 @@ namespace AccessibilityInsights
         {
             UpdateResult result = UpdateResult.Unknown;
 
-            // The UAC prompt from the version switcher will appear behind the main window.
-            // If the window is topmost, we store, change, and restore the value in this method.
+            // If the window is topmost, the UAC prompt from the version switcher will appear behind the main window.
+            // To prevent this, save the previous topmost state, ensure that the main window is not topmost when the
+            // UAC prompt will display, then restore the previous topmost state.
             bool previousTopmostSetting = Topmost;
 
             try
--- a/src/AccessibilityInsights/MainWindowHelpers/AutoUpdate.cs
+++ b/src/AccessibilityInsights/MainWindowHelpers/AutoUpdate.cs
@@ -124,9 +124,9 @@ namespace AccessibilityInsights
         {
             UpdateResult result = UpdateResult.Unknown;
 
-            // The UAC prompt from the version switcher will appear behind the main window
-            // if it is topmost, we store, change, and restore the value in this method.
-            bool oldTopMost = Topmost;
+            // The UAC prompt from the version switcher will appear behind the main window.
+            // If the window is topmost, we store, change, and restore the value in this method.
+            bool previousTopmostSetting = Topmost;
 
             try
             {
@@ -150,7 +150,7 @@ namespace AccessibilityInsights
                 e.ReportException();
             };
 
-            Topmost = oldTopMost;
+            Topmost = previousTopmostSetting;
             ctrlProgressRing.Deactivate();
             Logger.PublishTelemetryEvent(TelemetryAction.Upgrade_InstallationError, TelemetryProperty.Error, result.ToString());
 
--- a/src/AccessibilityInsights/MainWindowHelpers/AutoUpdate.cs
+++ b/src/AccessibilityInsights/MainWindowHelpers/AutoUpdate.cs
@@ -1,10 +1,11 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
-using AccessibilityInsights.SharedUx.Telemetry;
+using AccessibilityInsights.CommonUxComponents.Dialogs;
 using AccessibilityInsights.Dialogs;
 using AccessibilityInsights.Extensions;
 using AccessibilityInsights.Extensions.Interfaces.Upgrades;
 using AccessibilityInsights.Resources;
+using AccessibilityInsights.SharedUx.Telemetry;
 using AccessibilityInsights.SharedUx.Utilities;
 using System;
 using System.Collections.Generic;
@@ -122,9 +123,15 @@ namespace AccessibilityInsights
         private async void DownLoadInstaller(IAutoUpdate autoUpdate, AutoUpdateOption updateOption)
         {
             UpdateResult result = UpdateResult.Unknown;
+
+            // The UAC prompt from the version switcher will appear behind the main window
+            // if it is topmost, so we store, change, and restore the value in this method.
+            bool oldTopMost = Topmost;
+
             try
             {
                 ctrlProgressRing.Activate();
+                Topmost = false;
                 result = await autoUpdate.UpdateAsync().ConfigureAwait(true);
 
                 Logger.PublishTelemetryEvent(TelemetryAction.Upgrade_DoInstallation, new Dictionary<TelemetryProperty, string>
@@ -143,13 +150,14 @@ namespace AccessibilityInsights
                 e.ReportException();
             };
 
+            Topmost = oldTopMost;
             ctrlProgressRing.Deactivate();
             Logger.PublishTelemetryEvent(TelemetryAction.Upgrade_InstallationError, TelemetryProperty.Error, result.ToString());
 
             string message = updateOption == AutoUpdateOption.RequiredUpgrade
                 ? GetMessageForRequiredUpdateFailure() : GetMessageForOptionalUpdateFailure();
 
-            MessageBox.Show(message);
+            MessageDialog.Show(message);
 
             if (updateOption == AutoUpdateOption.RequiredUpgrade)
             {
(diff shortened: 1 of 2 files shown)


src/AccessibilityInsights/MainWindowHelpers/AutoUpdate.cs
```diff
@@ -122,9 +123,15 @@ private UpdateDialog CreateUpgradeDialog(Uri releaseNotesUri, AutoUpdateOption u
         private async void DownLoadInstaller(IAutoUpdate autoUpdate, AutoUpdateOption updateOption)
         {
             UpdateResult result = UpdateResult.Unknown;
+
+            // The UAC prompt from the version switcher will appear behind the main window
+            // if it is topmost, so we store, change, and restore the value in this method.
```
DaveTryon: Remove "so" ?

src/AccessibilityInsights/Modes/ConfigurationModeControl.xaml.cs
```diff
@@ -170,8 +179,10 @@ private async Task<bool> HandleReleaseSwitch()
                 ex.ReportException();
 
                 Dispatcher.Invoke(() =>
-                    MessageDialog.Show(string.Format(CultureInfo.InvariantCulture, Properties.Resources.ConfigurationModeControl_VersionSwitcherException))
-                );
+                {
+                    MainWin.Topmost = oldTopMost;
```
DaveTryon: Does oldTopMost need to be restored in the non-Exception case?

src/AccessibilityInsights/Modes/ConfigurationModeControl.xaml.cs
```diff
@@ -159,8 +159,17 @@ private async Task<bool> HandleReleaseSwitch()
                 return false;
             }
 
+            // The UAC prompt from the version switcher will appear behind the main window
+            // if it is topmost, so we store, change, and restore the value in this method.
```
DaveTryon: Remove "so" ?
```suggestion
            // if it is topmost, we store, change, and restore the value in this method.
```

DaveTryon:
>                     this.Close();

Just to confirm, this will exit the app, so restoring TopMost is not needed?

---
Refers to: src/AccessibilityInsights/MainWindowHelpers/AutoUpdate.cs:144 in 9309de1. [](commit_id = 9309de1ce81b37e7fc5f3b0108eeb5ea10fe1550, deletion_comment = False)

src/AccessibilityInsights/Modes/ConfigurationModeControl.xaml.cs
jalkire: The window will have closed and the function will have returned `true` at that point. If it hasn't, there should have been an exception thrown.

jalkire:
> Just to confirm, this will exit the app, so restoring TopMost is not needed?

Yes, that is the idea.

src/AccessibilityInsights/Modes/ConfigurationModeControl.xaml.cs
DaveTryon: Gotcha. The "return false" outside of the catch block made me think there might be a separate non-Exception path. (Stylistically, I'd probably put the return false inside the catch block, so that if another path were introduced, the compiler would catch it automatically)

---
In reply to: [278254819](https://github.com/Microsoft/accessibility-insights-windows/pull/337#discussion_r278254819) [](ancestors = 278254819)

jalkire: Moved that return!

DaveTryon: :thumbsup:

---
In reply to: [278312603](https://github.com/Microsoft/accessibility-insights-windows/pull/337#discussion_r278312603) [](ancestors = 278312603)

src/AccessibilityInsights/MainWindowHelpers/AutoUpdate.cs
```diff
@@ -122,9 +123,15 @@ private UpdateDialog CreateUpgradeDialog(Uri releaseNotesUri, AutoUpdateOption u
         private async void DownLoadInstaller(IAutoUpdate autoUpdate, AutoUpdateOption updateOption)
         {
             UpdateResult result = UpdateResult.Unknown;
+
+            // The UAC prompt from the version switcher will appear behind the main window
+            // if it is topmost, we store, change, and restore the value in this method.
```
RobGallo: It's not clear what "it" refers to hear. Is it the UAC dialog or AI_Win.

src/AccessibilityInsights/MainWindowHelpers/AutoUpdate.cs
```diff
@@ -122,9 +123,15 @@ private UpdateDialog CreateUpgradeDialog(Uri releaseNotesUri, AutoUpdateOption u
...
+            bool oldTopMost = Topmost;
```
RobGallo: Could we call this variable previousTopmostSetting? I think that would be much more descriptive.

src/AccessibilityInsights/Modes/ConfigurationModeControl.xaml.cs
```diff
@@ -159,8 +159,17 @@ private async Task<bool> HandleReleaseSwitch()
                 return false;
             }
 
+            // The UAC prompt from the version switcher will appear behind the main window
+            // if it is topmost, we store, change, and restore the value in this method.
```
RobGallo: Again, "it" is ambiguous.

src/AccessibilityInsights/Modes/ConfigurationModeControl.xaml.cs
```diff
@@ -159,8 +159,17 @@ private async Task<bool> HandleReleaseSwitch()
...
+            bool oldTopMost = Configuration.AlwaysOnTop;
```
RobGallo: Again, I think a name like "previousTopmostSetting" would be more informative.

src/AccessibilityInsights/Modes/ConfigurationModeControl.xaml.cs
```diff
@@ -170,11 +179,13 @@ private async Task<bool> HandleReleaseSwitch()
                 ex.ReportException();
 
                 Dispatcher.Invoke(() =>
-                    MessageDialog.Show(string.Format(CultureInfo.InvariantCulture, Properties.Resources.ConfigurationModeControl_VersionSwitcherException))
-                );
-            }
+                {
+                    MainWin.Topmost = oldTopMost;
+                    MessageDialog.Show(string.Format(CultureInfo.InvariantCulture, Properties.Resources.ConfigurationModeControl_VersionSwitcherException));
```
RobGallo: I'm assuming we call String.Format here because there's no other way to indicate invariant culture? This looks pretty strange at first glance.

src/AccessibilityInsights/MainWindowHelpers/AutoUpdate.cs
```diff
@@ -143,13 +150,14 @@ private async void DownLoadInstaller(IAutoUpdate autoUpdate, AutoUpdateOption up
...
 
+            Topmost = oldTopMost;
             ctrlProgressRing.Deactivate();
             Logger.PublishTelemetryEvent(TelemetryAction.Upgrade_InstallationError, TelemetryProperty.Error, result.ToString());
 
             string message = updateOption == AutoUpdateOption.RequiredUpgrade
                 ? GetMessageForRequiredUpdateFailure() : GetMessageForOptionalUpdateFailure();
 
-            MessageBox.Show(message);
+            MessageDialog.Show(message);
```
RobGallo: Curious: Why the switch to MessageDialog? I'm guessing MessageDialog is a common component? If so, should we use it everywhere in our UI?

jalkire: Message dialog is our custom styled (and hopefully more accessible) version of MessageBox. Yes, I believe we should use it everywhere, unless there is a specific reason to use an alternative.

src/AccessibilityInsights/MainWindowHelpers/AutoUpdate.cs
jalkire: I had been advised that it refers to the preceding noun phrase, which is the main window in this case. Upon further research, I can find no evidence that this is a rule. Updating comment and my brain accordingly.

src/AccessibilityInsights/MainWindowHelpers/AutoUpdate.cs
jalkire: Yes

src/AccessibilityInsights/Modes/ConfigurationModeControl.xaml.cs
jalkire: Actually, we were calling string.Format because we used to be replacing format items in the string. We stopped doing that, but never removed the unneccessary string.format call. Removing it now!

DaveTryon:
> using System.Globalization;

Do we still need  System.Globalization? With the removal of the String.Format call, I suspect that this is no longer needed

---
Refers to: src/AccessibilityInsights/Modes/ConfigurationModeControl.xaml.cs:12 in d678af7. [](commit_id = d678af7472fee6c2222fee84c2becfae070226cc, deletion_comment = False)

src/AccessibilityInsights/Modes/ConfigurationModeControl.xaml.cs
```diff
@@ -159,8 +159,17 @@ private async Task<bool> HandleReleaseSwitch()
                 return false;
             }
 
+            // The UAC prompt from the version switcher will appear behind the main window.
+            // If the window is topmost, we store, change, and restore the value in this method.
```
DaveTryon: This comment still seems stilted. Suggestion:

// If the window is topmost, the UAC prompt from the version switcher will appear behind the main window.
// To prevent this, save the previous topmost state, ensure that the main window is not topmost when the
// UAC prompt will display, then restore the previous topmost state.

jalkire: Let's hope we've arrived on the perfect comment.

PR #344, opened 2019-04-30
Add Wildlife Manager to the list of protected code paths for axe-windows

RobGallo:
#### Describe the change

This code will be moved to axe-windows. Therefore, we don't want it modified in the ai-win repo.

#### PR checklist

- [ ] Includes UI changes?

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -20,3 +20,4 @@
 /src/Axe.Windows.Telemetry/ @AhmedAbdoOrtiga @RobGallo @karanbirsingh
 /src/Axe.Windows.TelemetryTests/ @AhmedAbdoOrtiga @RobGallo @karanbirsingh
 /src/UIAAssemblies/ @AhmedAbdoOrtiga @RobGallo @karanbirsingh
+/tools/WildlifeManager/ @AhmedAbdoOrtiga @RobGallo @karanbirsingh


jalkire:
Given that this app can be useful for testing AI-Win itself, do we want to remove it from here?

karanbirsingh:
> Given that this app can be useful for testing AI-Win itself, do we want to remove it from here?

I agree with @jalkire that this application would be helpful for AI-Win validation & end-to-end tests. One idea he mentioned offline is that each repo seems to require some test app (UI tests vs automation) but it's not strictly necessary those test apps need to be the same.

RobGallo:
> > Given that this app can be useful for testing AI-Win itself, do we want to remove it from here?
> 
> I agree with @jalkire that this application would be helpful for AI-Win validation & end-to-end tests. One idea he mentioned offline is that each repo seems to require some test app (UI tests vs automation) but it's not strictly necessary those test apps need to be the same.

@jalkire  and @karanbirsingh , @DaveTryon  can correct me if I'm wrong, but I believe the test app is critical for unit testing automation. If so, it _must_ be available from the axe-windows repo. And if that is indeed the case, I think it would be better to have the test app in only one place to reduce maintenance cost. if we need a test app for AI-Win that is different from the one we need for axe-windows, we should just create another test app for AI-Win. Thoughts?

DaveTryon:
> From @RobGallo: I believe the test app is critical for unit testing automation. If so, it _must_ be available from the axe-windows repo. 

The check-automation script definitely needs WildlifeManager.exe, so it can't go away at this time. That said, if Axe.Windows were to export the app via NuGet, we could reasonably have AIWin consume that NuGet package. Given the size and scope of the app, it's probably easiest to just have duplicates. That implies that they could get out of sync over time, and we'd have to make some decisions about whether or not we care. My gut is that we don't need to care about drift between the test apps.

jalkire:
@RobGallo I am not arguing that the test app should be excluded from the axe-windows repo. I see this as an opportunity to create a "new" app by duplicating wildlife manager in both repos, but with the intention that they will diverge from this point forward. This would eliminate the maintence concern.

RobGallo:
@jalkire ,

Let's duplicate when we're sure that's what we want to do. In the meantime, it's safe to defer that decision.

PR #352, opened 2019-05-09
Enable code coverage in build result 

AhmedAbdoOrtiga:
#### Describe the change
Enable code coverage to see how much we cover and act based on that.

#### PR checklist

- [ ] Includes UI changes?

### DIFF
--- a/build/prbuild.yml
+++ b/build/prbuild.yml
@@ -50,7 +50,7 @@ jobs:
         **\*test*.dll
         !**\obj\**
       vsTestVersion: 15.0
-      codeCoverageEnabled: false
+      codeCoverageEnabled: true
       platform: '$(BuildPlatform)'
       configuration: release
       rerunFailedTests: true


flyingUnderTheRadar:
I've tried this in the past. We are around about 65%. Is there a recent push to find out where we are and where we want to go towards?

AhmedAbdoOrtiga:
> I've tried this in the past. We are around about 65%. Is there a recent push to find out where we are and where we want to go towards?

@flyingUnderTheRadar  Yeah, @markreay wants to enable it even if it's not fully covered and then start covering more parts of the code down the road.

DaveTryon:
Some thoughts:
1. Why only release? Shouldn't we also be checking debug?
2. Since this is in prbuild.yml, it should be possible for anyone to access it without GitHub auth (or at least without requiring an @microsoft.com account)
3. Should the daily dashboard show code coverage results from the PR build or from what's merged into master? I'd argue for the latter
4. I would suggest that we merge this **after** the axe-windows assemblies are removed. That seems more meaningful than starting combined, then taking a big drop when axe-windows is no longer being covered

PR #369, opened 2019-05-21
Remove DesktopUI project

jalkire:
#### Describe the change
At this point, DesktopUI contains only the highlighters. I see no reason why they should get their own assembly, so this PR moves them into SharedUx. In my opinion, this PR is a step in the right direction, not the end destination. Long term, we may want to create a more logical distinction between our Ux assemblies than currently exists. If people feel this isn't even a step in the right direction, or that we should take another step at the same time, feel free to say so 😊

There should be no impact on functionality; no logic was changed.

#### PR checklist

- [ ] Includes UI changes?


src/AccessibilityInsights.Win32/Properties/AssemblyInfo.cs
```diff
@@ -13,7 +13,7 @@
 [assembly: AssemblyConfiguration("")]
 [assembly: AssemblyCompany("")]
 [assembly: AssemblyProduct("AccessibilityInsights.Win32")]
-[assembly: AssemblyCopyright("Copyright �  2018")]
+[assembly: AssemblyCopyright("Copyright �  2018")]
```
karanbirsingh: Can you take a glance at this copyright symbol changing?

jalkire: Changed back to ©

Issue #347, opened 2019-05-03
[BUG] Version bar is confusing


DaveTryon:
(This was a reworking of #335, which will be closed. It's not strictly a bug, but is bug size so is being tracked as a bug)

**Describe the bug**
The current version bar in AI-Win includes potentially confusing information and isn't generally useful. After conversations with @cheade, here's where we landed:

- In the "typcial" environment, we want the version bar to be blank (don't distract the user)
- If the installed build is not currently associated with the Production channel, inform the user in the version bar
- If the installed build does not have UIAccess support, inform the user in the version bar
- Display the full information (build version, associated channel, UIAccess availability) in the about dialog

Assigning to Chelsea to get visuals of the new version bar and about box.

DaveTryon:
After further discussion with @cheade, we're going to display the UIAccess status only in the About box, and omit it completely from the version bar

DaveTryon:
This is fixed in release https://github.com/microsoft/accessibility-insights-windows/releases/tag/v1.1.0899.01

PR #388, opened 2019-06-17
Update how version and UIAccess are displayed

DaveTryon:
#### Describe the change
Update how version and UIAccess are displayed. Make the UIAccess status more accessible as a result.

#### PR checklist

- [x] Does this address an existing issue? If yes, Issue# - #347 
- [x] Includes UI changes?

Changes to the version bar:
- UIAccess information has been removed from the version bar.
- In prerelease versions (defined as being on Canary or Insider channel, or running a build newer than the current build in the Production channel), the build is called out as a Prerelease build.
- If the Channel is Production AND the version isn't newer than the current Production build, the version bar will be empty.

Changes to the about dialog appear below:
- The version label is now a hyperlink that points to the version-specific release notes.
- UIAccess availability is now displayed in the about dialog.
Screenshot of new about page (with highlighted differences) appears below. The lack of UIAccess support and the connection dialog are expected, as I was running an unsigned build from my build environment:
image

[Uri was corrected to include leading zeros in iteration 4] Clicking on the version hyperlink opens https://github.com/microsoft/accessibility-insights-windows/releases/tag/v1.1.0897.01 in the default browser. This page will exist for all customer builds. In my case, since no published build happens to match the auto-assigned local version, the link results in a 404 error. This will not be a problem for customer builds.

Implementation note: I wanted to leverage the code that already exists to determine what builds are expected in the Production channel. To do this, I've added a new state to AutoUpdateOption. This state indicates that the current build is newer than the build supported in the current channel, and this is then used to identify cases where we're on the Production channel, but on a prerelease build.

### DIFF
--- a/src/AccessibilityInsights.SharedUx/Controls/SettingsTabs/AboutTabControl.xaml.cs
+++ b/src/AccessibilityInsights.SharedUx/Controls/SettingsTabs/AboutTabControl.xaml.cs
@@ -5,6 +5,7 @@ using AccessibilityInsights.SharedUx.Telemetry;
 using AccessibilityInsights.Win32;
 using System;
 using System.Diagnostics;
+using System.Globalization;
 using System.Windows;
 using System.Windows.Controls;
 using System.Windows.Documents;
@@ -22,11 +23,8 @@ namespace AccessibilityInsights.SharedUx.Controls.SettingsTabs
         public AboutTabControl()
         {
             InitializeComponent();
-            lbVersion.Content = VersionTools.GetAppVersion();
-            if (!NativeMethods.IsRunningWithUIAccess())
-            {
-                lbUIAccess.Content = Properties.Resources.LabelUIAccessAvailable;
-            }
+            InitializeVersionInformationLink();
+            InitializeUIAccessLabel();
         }
 
         /// <summary>
@@ -76,6 +74,23 @@ namespace AccessibilityInsights.SharedUx.Controls.SettingsTabs
                 // silently ignore. 
             }
 #pragma warning restore CA1031 // Do not catch general exception types
+        }
+
+        private void InitializeVersionInformationLink()
+        {
+            string version = VersionTools.GetAppVersion();
+            hlVersion.NavigateUri = new Uri(string.Format(CultureInfo.InvariantCulture,
+                Properties.Resources.VersionLinkFormat, version), UriKind.Absolute);
+            hlVersion.Inlines.Clear();
+            hlVersion.Inlines.Add(string.Format(CultureInfo.InvariantCulture,
+                Properties.Resources.VersionLinkContentFormat, version));
+        }
+
+        private void InitializeUIAccessLabel()
+        {
+            lbUIAccess.Content = NativeMethods.IsRunningWithUIAccess() ?
+                Properties.Resources.LabelUIAccessAvailable :
+                Properties.Resources.LabelUIAccessNotAvailable;
         }
     }
 }
--- a/src/AccessibilityInsights.SharedUx/Controls/SettingsTabs/AboutTabControl.xaml.cs
+++ b/src/AccessibilityInsights.SharedUx/Controls/SettingsTabs/AboutTabControl.xaml.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 using AccessibilityInsights.SharedUx.Misc;
 using AccessibilityInsights.SharedUx.Telemetry;
+using AccessibilityInsights.Win32;
 using System;
 using System.Diagnostics;
 using System.Windows;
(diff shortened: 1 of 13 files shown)


src/AccessibilityInsights.SharedUx/Controls/SettingsTabs/AboutTabControl.xaml.cs
```diff
@@ -21,7 +23,8 @@ public partial class AboutTabControl : UserControl
         public AboutTabControl()
         {
             InitializeComponent();
-            lbVersion.Content = VersionTools.GetAppVersion();
+            InitializeVersionInformationLink();
```
DaveTryon: >InitializeVersionInformationLink [](start = 12, length = 32)

@Jalkire, I'd love to bind these in XAML rather than in code, but my attempts were unsuccessful. I welcome any suggestions you may have

DaveTryon:
> using AccessibilityInsights.Win32;

No longer needed. Will remove

---
Refers to: src/AccessibilityInsights/MainWindowHelpers/ControlHelper.cs:11 in ad6f9ef. [](commit_id = ad6f9ef7a8cf8366881fecde3c5f065bf531563b, deletion_comment = False)

DaveTryon:
> using System.Text;

No longer needed. Will remove.

---
Refers to: src/AccessibilityInsights/MainWindowHelpers/ControlHelper.cs:15 in ad6f9ef. [](commit_id = ad6f9ef7a8cf8366881fecde3c5f065bf531563b, deletion_comment = False)

src/AccessibilityInsights.SharedUx/Controls/SettingsTabs/AboutTabControl.xaml.cs
jalkire: I think `ChannelConfigControl` and its accompanying `ChannelConfigViewModel` should be one of the simpler uses of bindings that you could model this after. One potentially unintuitive detail of bindings is that the property to which you are binding needs to be a member of the UIElement's [DataContext](https://docs.microsoft.com/en-us/dotnet/api/system.windows.frameworkelement.datacontext?view=netframework-4.8).

A route you could take is:
1. Make some `AboutTabViewModel` class which will contain the dynamic values you want to display
2. Set it as `AboutTabControl`'s DataContext
3. Bind the appropriate properties

[Official example](https://docs.microsoft.com/en-us/dotnet/framework/wpf/data/how-to-specify-the-binding-source?view=netframework-4.8)

src/AccessibilityInsights.SharedUx/Properties/Resources.resx
```diff
@@ -1561,11 +1558,32 @@ Do you want to change the channel? </value>
...
+  </data>
+  <data name="hlVersion" xml:space="preserve">
+    <value>Version</value>
+  </data>
+  <data name="Placeholder" xml:space="preserve">
+    <value>*** placeholder ***</value>
+    <comment>Gets replaced at runtime</comment>
+  </data>
+  <data name="VersionLinkContentFormat" xml:space="preserve">
+    <value>Version {0}</value>
```
DaveTryon: >{0} [](start = 19, length = 3)

This is wrong. This string strips out the leading zero's in the last 2 pieces of the Uri, and those zero's are a required pat of the Uri

src/AccessibilityInsights.SharedUx/Properties/Resources.resx
```diff
@@ -1561,11 +1558,28 @@ Do you want to change the channel? </value>
...
+  </data>
+  <data name="hlVersion" xml:space="preserve">
+    <value>Version</value>
+  </data>
+  <data name="VersionLinkContentFormat" xml:space="preserve">
+    <value>Version {0}</value>
+    <comment>{0} is the placeholder for the app version string</comment>
+  </data>
+  <data name="VersionLinkFormat" xml:space="preserve">
+    <value>https://github.com/microsoft/accessibility-insights-windows/releases/tag/v{0}.{1}.{2:0000}.{3:00}</value>
```
DaveTryon: >v{0}.{1}.{2:0000}.{3:00} [](start = 84, length = 24)

This was wrong in version 3 because it lost the leading zeros. Fixed in iteration 4

src/AccessibilityInsights.SharedUx/Controls/SettingsTabs/AboutTabControl.xaml.cs
DaveTryon: Thanks for the help here! Binding added in iteration 4

---
In reply to: [294556168](https://github.com/microsoft/accessibility-insights-windows/pull/388#discussion_r294556168) [](ancestors = 294556168)

DaveTryon: Iteration 4 uses binding

---
In reply to: [294487918](https://github.com/microsoft/accessibility-insights-windows/pull/388#discussion_r294487918) [](ancestors = 294487918)

PR #390, opened 2019-06-18
Clean up telemetry

DaveTryon:
#### Describe the change
- Connect 2 recently-added telemetry events that were never connected.
- Change the Logger unit tests to be independent of specific values for TelemetryProperty and TelemetryAction
- Remove unused values for TelemetryProperty and TelemetryAction. I've confirmed with the existing telemetry that none of these values are being used in current builds. Many of these old values were remnants of code that was removed more than a year ago.
- Create AccessibilityInsights.Misc.TelemetryEventFactory to combine several closely-related pieces of code into a single class. This replaces most of the instances in the AccessibilityInsights project where Logger.PublishTelemetryEvent was being called with one or more TelemetryProperty values)

#### PR checklist

- [ ] Includes UI changes?

### DIFF
--- a/src/AccessibilityInsights/MainWindowHelpers/TestMode.cs
+++ b/src/AccessibilityInsights/MainWindowHelpers/TestMode.cs
@@ -196,11 +196,8 @@ namespace AccessibilityInsights
 
             var v = SelectAction.GetDefaultInstance().SelectLoadedData(path, selectedElementId);
 
-            if (selectedElementId.HasValue)
-            {
-                Logger.PublishTelemetryEvent(TelemetryAction.Hierarchy_Load_Protocol, TelemetryProperty.ProtocolMode, v.Item2.Mode.ToString());
-            }
-            Logger.PublishTelemetryEvent(TelemetryAction.Hierarchy_Load_NewFormat, TelemetryProperty.FileMode, v.Item2.Mode.ToString());
+            TelemetryProperty mode = selectedElementId.HasValue ? TelemetryProperty.ProtocolMode : TelemetryProperty.FileMode;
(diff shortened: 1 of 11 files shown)


src/AccessibilityInsights/MainWindowHelpers/TestMode.cs
```diff
@@ -196,11 +196,8 @@ private void StartLoadingSnapshot(string path, int? selectedElementId = null)
 
             var v = SelectAction.GetDefaultInstance().SelectLoadedData(path, selectedElementId);
 
-            if (selectedElementId.HasValue)
-            {
-                Logger.PublishTelemetryEvent(TelemetryAction.Hierarchy_Load_Protocol, TelemetryProperty.ProtocolMode, v.Item2.Mode.ToString());
-            }
-            Logger.PublishTelemetryEvent(TelemetryAction.Hierarchy_Load_NewFormat, TelemetryProperty.FileMode, v.Item2.Mode.ToString());
+            TelemetryProperty mode = selectedElementId.HasValue ? TelemetryProperty.ProtocolMode : TelemetryProperty.FileMode;
+            Logger.PublishTelemetryEvent(TelemetryAction.Hierarchy_Load_NewFormat, mode, v.Item2.Mode.ToString());
```
karanbirsingh: This is probably intentional, but just to confirm - the original code sends 2 events if `selectedElementId.HasValue` is true, whereas this new code only sends 1 event. Is this as expected?

DaveTryon: Yes. They can be distinguished by the underlying property, if needed

PR #393, opened 2019-06-19
FXCop changes

DaveTryon:
#### Describe the change
FXCop issues: Replace StringComparison.InvariantCulture with StringComparison.Ordinal when comparing strings

#### PR checklist

- [ ] Includes UI changes?

### DIFF
--- a/src/AccessibilityInsights.Extensions.GitHub/ConfigurationModelControl.xaml.cs
+++ b/src/AccessibilityInsights.Extensions.GitHub/ConfigurationModelControl.xaml.cs
@@ -59,7 +59,7 @@ namespace AccessibilityInsights.Extensions.GitHub
         public void TextChangeUpdateSaveButtonHelper()
         {
             string curURL = this.tbURL.Text;
-            if (!string.IsNullOrEmpty(curURL) && ((this.Config != null && string.IsNullOrEmpty(this.Config.RepoLink)) || !this.Config.RepoLink.Equals(curURL, StringComparison.InvariantCulture)))
+            if (!string.IsNullOrEmpty(curURL) && ((this.Config != null && string.IsNullOrEmpty(this.Config.RepoLink)) || !this.Config.RepoLink.Equals(curURL, StringComparison.Ordinal)))
             {
                 canSave = true;
             }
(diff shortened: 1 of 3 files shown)


src/AccessibilityInsights.Extensions.GitHub/ConfigurationModelControl.xaml.cs
```diff
@@ -59,7 +59,7 @@ public void TextChangeUpdateSaveButton(object sender, EventArgs e)
         public void TextChangeUpdateSaveButtonHelper()
         {
             string curURL = this.tbURL.Text;
-            if (!string.IsNullOrEmpty(curURL) && ((this.Config != null && string.IsNullOrEmpty(this.Config.RepoLink)) || !this.Config.RepoLink.Equals(curURL, StringComparison.InvariantCulture)))
+            if (!string.IsNullOrEmpty(curURL) && ((this.Config != null && string.IsNullOrEmpty(this.Config.RepoLink)) || !this.Config.RepoLink.Equals(curURL, StringComparison.Ordinal)))
```
DaveTryon: >Ordinal [](start = 175, length = 7)

The old code ignored the case, but since GitHub repos are case-sensitive, maybe we should use case-sensitive comparisons in this class?

jalkire: Are github URLs case sensitive? https://github.com/Microsoft/accessibility-insights-windows
https://githUB.com/miCRosoft/accessibility-insights-WINDOWS

jalkire: This looks like it doesn't ignore case before or after:
https://docs.microsoft.com/en-us/dotnet/api/system.stringcomparison?view=netframework-4.7.1

flyingUnderTheRadar: Just to mention what I saw (Ortiga demoed it to me) they (at least used to be or maybe with certain browsers; not that I can repro) were case sensitive.

DaveTryon: Duh--brain cramp! The checks are/were case sensitive. Looks like we have parity

DaveTryon: @flyingUnderTheRadar : uri host resolution is case-insensitive. Once you're on the host, case sensitivity depends on the configuration of the host machine. Windows hosts are case-insensitive by default, Lunix hosts are case-sensitive by default. Defaults can be overridden on either platform.

Issue #394, opened 2019-06-20
[BUG] Upgrade prompt displays when it shouldn't


DaveTryon:
Please check whether the bug has [already been filed](https://github.com/Microsoft/accessibility-insights-windows/issues).

**Describe the bug**
The upgrade dialog is displaying when installing the latest release on a clean machine

**To Reproduce**
1. Start with a clean machine (no previous installation of AIWin & no config settings)
2. Install from https://github.com/microsoft/accessibility-insights-windows/releases/download/v1.1.0899.01/AccessibilityInsights.msi
3. Wait for the app to start

**Expected behavior**
This is the latest build. No upgrade should be prompted

**Actual behavior**
You're prompted to upgrade. If you choose to upgrade, you get an error message.

**Desktop (please complete the following information):**
 - OS: Windows 10 1809
 - Accessibility Insights for Windows Version: 1.1.899.1

**Additional context**
This is a regression caused by #388. We added a new AutoUpdateOption status, but didn't update the code that decides whether or not to display the upgrade dialog. Fix is to update the if statement that decides whether or not to display the dialog.

DaveTryon:
Confirmed fixed in release https://github.com/microsoft/accessibility-insights-windows/releases/tag/v1.1.0904.01

PR #395, opened 2019-06-20
Fix check for prompting for upgrades

DaveTryon:
#### Describe the change
Decision about showing the upgrade should return false for `AutoUpdateOption.NewerThanCurrent`. This is a new enum value, but the consuming logic in the UI wasn't updated to handle the new value.

#### PR checklist

- [x] Does this address an existing issue? If yes, Issue# - #394 
- [ ] Includes UI changes?

### DIFF
--- a/src/AccessibilityInsights/MainWindowHelpers/AutoUpdate.cs
+++ b/src/AccessibilityInsights/MainWindowHelpers/AutoUpdate.cs
@@ -84,11 +84,9 @@ namespace AccessibilityInsights
             }
 
             autoUpdateOption = updateOption;
-            if (autoUpdateOption == AutoUpdateOption.Current || autoUpdateOption == AutoUpdateOption.Unknown)
-            {
-                return false;
-            }
-            return true;
+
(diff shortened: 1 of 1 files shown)


src/AccessibilityInsights/MainWindowHelpers/AutoUpdate.cs
```diff
@@ -84,11 +84,9 @@ private static bool ShouldUserUpgrade(IAutoUpdate autoUpdate, out AutoUpdateOpti
             }
 
             autoUpdateOption = updateOption;
-            if (autoUpdateOption == AutoUpdateOption.Current || autoUpdateOption == AutoUpdateOption.Unknown)
-            {
-                return false;
-            }
-            return true;
+
+            return autoUpdateOption == AutoUpdateOption.OptionalUpgrade
```
DaveTryon: >OptionalUpgrade [](start = 56, length = 15)

Old code defaulted to true for unknown values. New code defaults to false.

Issue #379, opened 2019-05-31
[BUG] Weird behaviour with display and Jaws


ferBonnin:
Please check whether the bug has [already been filed](https://github.com/Microsoft/accessibility-insights-windows/issues).

**Describe the bug**
When connected to a different screen and using Jaws, after running fastPass, the automated checks UI goes away

**To Reproduce**
1. turn on Jaws and connect to a different display
2. run automated checks in an app
3. UI dissapears

**Expected behavior**
UI works as expected without a screen reader

**Actual behavior**
UI dissapears

**Desktop (please complete the following information):**
 - OS: [e.g. Windows 10 1809] (Get the version by running `winver` from the command line) Windows 10
 - Accessibility Insights for Windows Version: Canary
 - Target Application: [e.g. Chrome, Wildlife Manager] Wildlife Manager

**Additional context**
Found during demo.

Sirisha-Kommushetty:
@jalkire - Could you prioritize this when Rob is in town.

jalkire:
@Sirisha-Kommushetty Update on this. I have been able to repro the problem consistently on my machine, which is good and bad news. I have not been able to find a solution--our previous fix is no longer working. We've (mostly @flyingUnderTheRadar) spent a fair amount of time trying to mitigate this in the past with mixed results, so I'm not optimistic about finding a compromise-free solution. We can chat in person if you'd like some context around the issue.

flyingUnderTheRadar:
Just to give some more context
@RobGallo - this is the one that we worked together to repro on your machine. 

We were able to fix it via [this PR](https://github.com/microsoft/accessibility-insights-windows/pull/114/files). The fix that worked (evidenced by @jalkire @ the CSUN demo) simply invalidates the mainwindow visual and forces a re render with the highest priority.

**Code around this has since changed and the issue has started reappearing**

As far as I can tell this seems to be a WPF issue (can be repro'd with other apps as well). 

In my opinion if we want to pick this up I would highly suggest getting the folks at WPF involved pretty early to better guide us.

DaveTryon:
@ferBonnin, this should be in the next build (1.1.901.1, I believe)

DaveTryon:
@ferBonnin, please regress using release https://github.com/microsoft/accessibility-insights-windows/releases/tag/v1.1.0904.01

ferBonnin:
Cannot repro in my computer. Per conversation with @jalkire that could repro this in his computer, this release has fixed the issue.

PR #396, opened 2019-06-20
Update the existing Jaws repaint glitch workaround

DaveTryon:
#### Describe the change
Update the existing Jaws repaint glitch workaround. This glitch is apparently caused by interaction of WPF and the Jaws-installed screen driver. Hide the window before invalidating and forcing a repaint. To minimize flash, only do this is we have reason to believe that Jaws is on the machine.

#### PR checklist

- [x] Does this address an existing issue? If yes, Issue# - #379 
- [ ] Includes UI changes?

Old behavior without Jaws
WithoutJaws-Old

Old behavior with Jaws
WithJaws-Old

New behavior without Jaws
WithoutJaws-New

New behavior with Jaws
WithJaws-New

### DIFF
--- a/src/AccessibilityInsights/Modes/TestModeControl.xaml.cs
+++ b/src/AccessibilityInsights/Modes/TestModeControl.xaml.cs
@@ -9,6 +9,7 @@ using AccessibilityInsights.SharedUx.Interfaces;
 using AccessibilityInsights.SharedUx.Settings;
 using AccessibilityInsights.SharedUx.Telemetry;
 using AccessibilityInsights.SharedUx.Utilities;
+using AccessibilityInsights.Win32;
 using Axe.Windows.Actions;
 using Axe.Windows.Actions.Contexts;
 using Axe.Windows.Actions.Enums;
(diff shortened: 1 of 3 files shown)


src/AccessibilityInsights/Modes/TestModeControl.xaml.cs
```diff
@@ -261,8 +262,13 @@ public async Task SetElement(Guid ecId)
                 // Improves the reliability of the rendering but does not solve across the board issues seen with all WPF apps.
                 this.Dispatcher.BeginInvoke(DispatcherPriority.Send, new Action(() =>
                 {
-                    Application.Current.MainWindow.InvalidateVisual();
-                    Application.Current.MainWindow.Visibility = Visibility.Visible;
+                    // This is a workaround for a repaint issue caused by an interaction between Jaws and WPF
+                    if (NativeMethods.IsExternalScreenReaderActive())
```
jalkire: This is returning true for me after I close JAWS. Do we know if the value needs a log out or reboot to update?

DaveTryon: This is mainly a nicety to reduce unnecessary flashing. If we assume that most people never install Jaws, the "lingering" status shouldn't be a problem.

jalkire: Fair enough!

PR #397, opened 2019-06-20
Consolidate ProgressRingControl class (currently duplicated)

DaveTryon:
#### Describe the change
The ProgressRingControl class was copied from SharedUx into CommonUxControls, but the SharedUx copy was never removed. This PR removes the SharedUx version and points all SharedUx.ProgressRingControl references to the CommonUxControls.ProgressRingControl.

#### PR checklist

- [x] Run through of all [test scenarios](https://github.com/Microsoft/accessibility-insights-windows/blob/master/docs/Scenarios.md) completed?
- [ ] Does this address an existing issue? If yes, Issue# - 
- [ ] Includes UI changes? UI should be identical after the change
  - [ ] Run the production version of Accessibility Insights for Windows against a version with changes.
  - [ ] Attach any screenshots / GIF's that are applicable.

> Note: After the PR has been created, certain checks will be kicked off. All of these checks must pass before a merge.

PR #398, opened 2019-06-21
Consolidate ProgressRingControl class (currently duplicated)

jalkire:
#### Describe the change
This PR builds off of @DaveTryon's PR #397. Description from there:
The ProgressRingControl class was copied from SharedUx into CommonUxControls, but the SharedUx copy was never removed. This PR removes the SharedUx version and points all SharedUx.ProgressRingControl references to the CommonUxControls.ProgressRingControl.

In addition, this PR passes the scanning sound preference to ProgressRingControl via dependency property and binding. This needs to be done because the CommonUxControls.ProgressRingControl is unable to directly reference user preferences.

There should be no change in user experience/functionality from this PR.

#### PR checklist

- [ ] Includes UI changes?


src/AccessibilityInsights.SharedUx/Dialogs/LoadingBox.xaml
```diff
@@ -1,23 +0,0 @@
-<!-- Copyright (c) Microsoft. All rights reserved.
```
jalkire: `LoadingBox` contained a reference to `ProgressRingControl`, but is not in use. I thought it would be better to remove the class than update it.

Issue #151, opened 2019-03-04
Tab stops test does not draw the first tab stop indicator.


risanaka:
**Describe the bug**
When running the tab stops test, the number (1) isn't shown on an element if it already has focus when you turn on the tab stops test. 

**To Reproduce**
1. Open the [Enterprise app](https://www.microsoft.com/en-us/p/enterprise-database-developer-sample/9pf1wcv13501).
1. In Customers, select + New.
1. In New customer, set focus in the First name field.
1. Turn on Record tab stops.
1. Press the Tab key once.

**Expected behavior**
The number (1) appears on the First name field.

**Actual behavior**
The number (1) appears on the Last name field.

**Screenshots or .GIF**

**Desktop (please complete the following information):**
 - OS: [e.g. Windows 10 1809] (Get the version by running `winver` from the command line)
 - Accessibility Insights for Windows Version:
 - Target Application: [e.g. chrome, safari]
 - Target Application Version: [e.g. 22]

**Additional context**
Priority requested -

neocln:
Since it is a feature request, I remove [BUG] from the title.

risanaka:
@ferBonnin - this issue has a consistency problem between Web and Windows on how Tab stops is displayed. Can you drive the conversation?

ferBonnin:
@risanaka for sure :)

Sirisha-Kommushetty:
@ferBonnin - Can we connect on this. Seems like a bug that can be fixed for both Windows and Web.

ferBonnin:
@Sirisha-Kommushetty lets connect. The problem here is that Web and Windows behave differently. (Windows doesnt draw the first tab stop indicator)

jalkire:
@Sirisha-Kommushetty @ferBonnin For context, there is no good/intentional reason why Windows doesn't draw the first tab stop. It just happened to be that way. I hope we decide to make the change 👍

ferBonnin:
@jalkire thanks for the context! :) I agree

ferBonnin:
from conversation with Peter, size of this is probably bug, changing tags and marking ready for work

Sirisha-Kommushetty:
@risanaka - Could you validate this in the latest canary builds and close it? Thank you!

risanaka:
@Sirisha-Kommushetty validated and closing. Thanks for the fix :)

PR #403, opened 2019-06-24
Show first tab stop in tab stops test

jalkire:
#### Describe the change
Previously, started listening to events and then set focus to the target element. This didn't work because the UIA event listener wasn't actually registered before the focus changed--there is a slight gap between `RegisterAutomationEventListener` completing and the listener actually starting.
This PR delays setting focus on the target element until we get a message from the event listener confirming that it has started listening.

#### PR checklist

- [x] Does this address an existing issue? If yes, Issue #151
- [ ] Includes UI changes?
before

### DIFF
--- a/src/AccessibilityInsights.SharedUx/Controls/TestTabs/TabStopControl.xaml.cs
+++ b/src/AccessibilityInsights.SharedUx/Controls/TestTabs/TabStopControl.xaml.cs
@@ -396,7 +396,7 @@ namespace AccessibilityInsights.SharedUx.Controls.TestTabs
         /// </summary>
         /// <param name="message"></param>
         private void OnFocusChanged(EventMessage message)
-    {
+        {
             // only when focus is chosen
             if (message.EventId == EventType.UIA_AutomationFocusChangedEventId)
             {
--- a/src/AccessibilityInsights.SharedUx/Controls/TestTabs/TabStopControl.xaml.cs
+++ b/src/AccessibilityInsights.SharedUx/Controls/TestTabs/TabStopControl.xaml.cs
@@ -360,7 +360,7 @@ namespace AccessibilityInsights.SharedUx.Controls.TestTabs
                 IsTabStopLooped = false;
 
                 CurrentElement = null;
-                EventHandler?.RegisterAutomationEventListener(EventType.UIA_AutomationFocusChangedEventId, this.OnFocusChanged);
+                EventHandler?.RegisterAutomationEventListener(EventType.UIA_AutomationFocusChangedEventId, this.EventMessageRecieved);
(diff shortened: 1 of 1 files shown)


src/AccessibilityInsights.SharedUx/Controls/TestTabs/TabStopControl.xaml.cs
```diff
@@ -363,9 +359,9 @@ private void StartRecordingEvent()
                 TabStopCount = 0;
                 IsTabStopLooped = false;
 
-                this.EventHandler?.RegisterAutomationEventListener(EventType.UIA_AutomationFocusChangedEventId, this.OnFocusChanged);
-
-                this.lvElements.Items.Clear();
+                CurrentElement = null;
```
jalkire: `CurrentElement = null;` should already have been happening somewhere. Perhaps we lost it inadvertently.

src/AccessibilityInsights.SharedUx/Controls/TestTabs/TabStopControl.xaml.cs
```diff
@@ -448,6 +444,17 @@ private void OnFocusChanged(EventMessage message)
                     message.Dispose();
                 }
             }
+            else if (IsMessageRegisterSuccess(message))
```
karanbirsingh: I wonder if two changes would help with readability here:
- move the `IsMessageRegisterSuccess` case to the top of the function since it occurs first time-wise
- change the name of the method to `eventReceived` or `focusEventReceived` so it's a bit more clear that other messages can appear

I'm not sure how common our usage pattern is (triggering an event after registration success but before first focus event), but future thinking in Axe.Windows might be helpful to improve the interface for clients (so they don't need to know about the details of `IsMessageRegisterSuccess` as you mention)

jalkire: Agree with rename. But `IsMessageRegisterSuccess` is the significantly less commonly expected branch path and has to do string comparisons on every event property pair, so I'd prefer that we don't call it more than is necessary. Do you feel like its later location is a significant hindrance to understanding the function?

karanbirsingh: I was mainly confused that this method can be called with non-focus changed events - a text comment would work too if we want to maintain the structure as-is. The string comparisons will likely be short-circuited out, if that helps at all

jalkire: Hopefully it is now more readable

PR #407, opened 2019-06-28
Adding command line parameters to ai-windows for config path

karanbirsingh:
#### Describe the change
This is a PR that provides the following command line options to AI-Win:
- FileToOpen
- ConfigFolder
- UserDataFolder
- AttachToDebugger

These are provided to ConfigurationManager through a FixedConfigSettingsProvider. We should take extra care while reviewing this PR to ensure no breaking changes to configuration paths.

#### PR checklist

- validation of existing file-opening via command line completed
- [NA] Does this address an existing issue? If yes, Issue# - 
- [NA] Includes UI changes?
  - [NA] Run the production version of Accessibility Insights for Windows against a version with changes.
  - [NA] Attach any screenshots / GIF's that are applicable.


karanbirsingh:
I forgot to mark this as a draft before clicking create so I have edited the title

src/AccessibilityInsights.VersionSwitcher/InstallationEngine.cs
```diff
@@ -175,7 +175,8 @@ private static void UpdateConfigWithNewChannel(string newChannel)
         {
             if (newChannel != null)
             {
-                var configFile = Path.Combine(DirectoryManagement.sConfigurationFolderPath, Constants.AppConfigFileName);
+                var defaultConfigPaths = FixedConfigSettingsProvider.CreateDefaultSettingsProvider();
```
karanbirsingh: @DaveTryon flagging this because I'm not sure how the configuration argument should work with the InstallationEngine; I'm not sure it should do anything other than look in the default/prod path

src/AccessibilityInsights.SharedUxTests/Settings/ConfigurationModelTests.cs
```diff
@@ -19,7 +19,9 @@ namespace AccessibilityInsights.SharedUxTests.Settings
     [TestClass()]
     public class ConfigurationModelTests
     {
-        public static string folderPath = Path.Combine(DirectoryManagement.sUserDataFolderPath, "ConfigurationTests");
+        public static readonly FixedConfigSettingsProvider testProvider = FixedConfigSettingsProvider.CreateDefaultSettingsProvider();
```
karanbirsingh: We could provide some temp folder instead for our unit tests if that feels more natural

src/AccessibilityInsights.SetupLibrary/DirectoryManagement.cs
```diff
@@ -11,46 +11,16 @@ namespace AccessibilityInsights.SetupLibrary
...
-            CreateFolder(sUserDataFolderPath);
-            CreateFolder(sConfigurationFolderPath);
-        }
-
         /// <summary>
         /// Create folder if it doesn't exist. 
         /// </summary>
         /// <param name="path"></param>
-        private static void CreateFolder(string path)
+        public static void CreateFolder(string path)
```
DaveTryon: >CreateFolder [](start = 27, length = 12)

It feels odd to have this as the only method in this class. Could we reasonably put this method into another class?

DaveTryon: >CreateFolder [](start = 27, length = 12)

This method is the only thing left in this class, and it's only called from Initizlize.cs. Seems like maybe a migration would make sense?

src/AccessibilityInsights/MainWindow.xaml.cs
```diff
@@ -263,13 +264,17 @@ void InitPanes()
...
         {
             var args = Environment.GetCommandLineArgs();
-
-            if (args.Contains(ArgAttach))
-            {
-                var dlg = new MessageDialog();
-                dlg.Message = Properties.Resources.SupportDebuggingDialogMessage;
-                dlg.ShowDialog();
-            }
+            Parser.Default.ParseArguments<CommandOptions>(args.Skip(1))
```
DaveTryon: >ParseArguments<CommandOptions>(args.Skip(1)) [](start = 27, length = 44)

Code smell: You're calling ParseArguments in 3 different places. You're also exposing the setters on CommandOptions. You can fix both by creating a wrapper class that _contains_ a CommandOptions class, initializes it once (maybe on-demand via a Lazy), then exposes get-only properties.

src/AccessibilityInsights/Misc/CommandOptions.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using CommandLine;
+
+namespace AccessibilityInsights.Misc
+{
+    public class CommandOptions
+    {
+        [Value(0, HelpText = "File to open in Accessibility Insights", Required = false)]
```
DaveTryon: >HelpText [](start = 18, length = 8)

I'm not familiar with this framework. Where does the help text come into play?

src/AccessibilityInsights.SetupLibrary/FixedConfigSettingsProvider.cs
```diff
@@ -0,0 +1,27 @@
...
+
+        public FixedConfigSettingsProvider(string configurationFolderPath, string userDataFolderPath)
+        {
+            this.ConfigurationFolderPath = configurationFolderPath;
+            this.UserDataFolderPath = userDataFolderPath;
+        }
+
+        public static FixedConfigSettingsProvider CreateDefaultSettingsProvider()
+        {
+            var sUserDataFolderPath = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.MyDoc‌​uments), "AccessibilityInsights");
```
DaveTryon: >sUserDataFolderPath [](start = 16, length = 19)

Not a fan of the "s" prefix here. Feels too much like we're trying to slip back to Reverse Hungarian Notation

src/AccessibilityInsights.SharedUx/Settings/ConfigurationManager.cs
```diff
@@ -29,12 +29,15 @@ public class ConfigurationManager
         /// Event recorder settings
         /// </summary>
         public RecorderSetting EventConfig { get; private set; }
+
```
DaveTryon: Replace this blank line with /// comments?

src/AccessibilityInsights.SharedUx/Settings/ConfigurationManager.cs
```diff
@@ -29,12 +29,15 @@ public class ConfigurationManager
         /// Event recorder settings
         /// </summary>
         public RecorderSetting EventConfig { get; private set; }
+        
+        public FixedConfigSettingsProvider PathProvider { get; }
```
DaveTryon: >PathProvider [](start = 43, length = 12)

"PathProvider" as a name seems to imply that we're only providing paths and nothing else

src/AccessibilityInsights.SharedUxTests/Settings/AppLayoutTests.cs
```diff
@@ -96,8 +96,8 @@ public void AppLayoutTest()
         [TestMethod()]
         public void RemoveConfigurationTest()
         {
-            string path = Path.Combine(DirectoryManagement.sUserDataFolderPath, TestLayoutFile);
-            foreach (string file in System.IO.Directory.EnumerateFiles(DirectoryManagement.sUserDataFolderPath))
+            string path = Path.Combine(provider.UserDataFolderPath, TestLayoutFile);
+            foreach (string file in System.IO.Directory.EnumerateFiles(provider.UserDataFolderPath))
```
DaveTryon: >System.IO. [](start = 36, length = 10)

Line 7 specifies System.IO, but we're repeating it several times in this class. If it's not needed, can we do some opportunistic cleanup?

src/AccessibilityInsights.SharedUxTests/Settings/ConfigurationModelTests.cs
DaveTryon: IIRC, each test run iteration gets its own working directory. Could we leverage that?

---
In reply to: [298770644](https://github.com/microsoft/accessibility-insights-windows/pull/407#discussion_r298770644) [](ancestors = 298770644)

src/AccessibilityInsights.VersionSwitcher/InstallationEngine.cs
DaveTryon: Default path is correct here. The setting overrides are intended to be used only in a test environment, and not used by end users.

---
In reply to: [298770551](https://github.com/microsoft/accessibility-insights-windows/pull/407#discussion_r298770551) [](ancestors = 298770551)

src/AccessibilityInsights/Misc/CommandOptions.cs
karanbirsingh: @DaveTryon For normal console apps running `program --help` should generate some help text. It looks like that's not working for the WPF app, so I can remove it & replace with a // comment. We could also just leave it in for the same documentation purpose. Do you prefer one or the other?

src/AccessibilityInsights/MainWindow.xaml.cs
DaveTryon: Your comment in InstallationEngine brings up another reason to consolidate the parsing. We may want to place some limitations about when command line overrides are respected. Settings that are not intended to be used by end users could potentially be ignored unless some runtime condition is satisfied. Maybe an environment variable?

---
In reply to: [298772144](https://github.com/microsoft/accessibility-insights-windows/pull/407#discussion_r298772144) [](ancestors = 298772144)

src/AccessibilityInsights/Misc/CommandOptions.cs
DaveTryon: I'd suggest removing the HelpText fields

---
In reply to: [298774055](https://github.com/microsoft/accessibility-insights-windows/pull/407#discussion_r298774055) [](ancestors = 298774055)

src/AccessibilityInsights/Misc/CommandOptions.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using CommandLine;
+
+namespace AccessibilityInsights.Misc
+{
+    public class CommandOptions
```
DaveTryon: >CommandOptions [](start = 17, length = 14)

We need to ensure that the current command line (i.e., a file is specified as the first parameter) continues to work correctly

src/AccessibilityInsights/AccessibilityInsights.csproj
```diff
@@ -80,6 +80,9 @@
     <Reference Include="Axe.Windows.Win32, Version=0.1.1.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL">
       <HintPath>..\packages\Microsoft.Axe.Windows.0.1.1-prerelease\lib\net471\Axe.Windows.Win32.dll</HintPath>
     </Reference>
+    <Reference Include="CommandLine, Version=2.5.0.0, Culture=neutral, PublicKeyToken=5a870481e358d379, processorArchitecture=MSIL">
+      <HintPath>..\packages\CommandLineParser.2.5.0\lib\net461\CommandLine.dll</HintPath>
```
DaveTryon: >CommandLine.dll [](start = 63, length = 15)

This will need to be added to the WXS file.

src/AccessibilityInsights/MainWindow.xaml.cs
jalkire: I agree with the below comment about removing the help text for our command line arguments--no need to advertise our command line features. I don't think we should take any active steps to make the arguments more difficult to use, though--is there are particular scenario you're concerned about?

src/AccessibilityInsights/Misc/CommandOptions.cs
jalkire: +1 to removing these, at least the new custom config ones, since those aren't really meant for typical users.

src/AccessibilityInsights/packages.config
```diff
@@ -1,5 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
+  <package id="CommandLineParser" version="2.5.0" targetFramework="net471" />
```
jalkire: Do we need to update thirdpartynotices.html or do any OSS registration related work?

src/AccessibilityInsights/MainWindow.xaml.cs
karanbirsingh: Chatted with Dave before heading out on Friday and we settled on what you've suggested, John - I will remove the HelpText, add `//comments`, but we won't take any other steps. Using these parameters in non-testing scenarios (except the file one, perhaps) feels like relying on undocumented behavior / implementation details

src/AccessibilityInsights/Misc/CommandOptions.cs
karanbirsingh: The file scenario still works correctly

src/AccessibilityInsights/packages.config
karanbirsingh: updated thirdpartynotices.html

PR #408, opened 2019-07-01
Introduce AIWinDriver and use it in the GettingStartPage UITest

jalkire:
#### Describe the change
AIWinDriver provides an organized and human readable way to interact with AI-Win in our UITests. You can see how it is used in the GettingStartedPage test and an outline of how it can be used in the LoadTestFile test (not implemented yet).

#### PR checklist

- [ ] Includes UI changes?

### DIFF
--- a/src/UITests/LoadTestFile.cs
+++ b/src/UITests/LoadTestFile.cs
@@ -1,0 +1,44 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace UITests
+{
+    /// <summary>
+    /// Start the app, open an A11yTest file, navigate to 2 views (automated checks and UIA hierarchy), spot check loaded data, exit app
+    /// </summary>
+    [TestClass]
+    public class LoadTestFile : AIWinSession
+    {
+        [TestMethod]
+        [TestCategory("NoStrongName")]
+        public void VerifyResultsInUIATreePage()
+        {
+            driver.TestMode.AutomatedChecks.ViewInUIATree();
+            // do appropriate testing
+            driver.TestMode.UIATree.BackToAutomatedChecks();
+        }
+
+        [TestMethod]
+        [TestCategory("NoStrongName")]
+        public void VerifyGridContents()
+        {
+            driver.TestMode.AutomatedChecks.ToggleAllExpanders();
+            // do appropriate testing
+            driver.TestMode.AutomatedChecks.ToggleAllExpanders();
+        }
+
+        [ClassInitialize]
+        public static void ClassInitialize(TestContext context)
+        {
+            Setup(context);
+
+            driver.GettingStarted.DismissTelemetry();
+            driver.GettingStarted.DismissPage();
+            driver.LiveMode.OpenFile("folder", "file");
+        }
+
+        [ClassCleanup]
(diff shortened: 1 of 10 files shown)


src/UITests/LoadTestFile.cs
```diff
@@ -0,0 +1,44 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace UITests
+{
+    /// <summary>
+    /// Start the app, open an A11yTest file, navigate to 2 views (automated checks and UIA hierarchy), spot check loaded data, exit app
+    /// </summary>
+    [TestClass]
+    public class LoadTestFile : AIWinSession
```
jalkire: This test passes but does nothing--it's just here to provide an outline of how the AIWinDriver class can be used. I'm happy to delete it for this PR if anyone feels strongly. Otherwise, it will be updated shortly as a part of this feature 🙂

src/UITests/UILibrary/Settings.cs
```diff
@@ -0,0 +1,38 @@
...
+        {
+            Session = session;
+            AboutTab = new AboutTab(session);
+            ApplicationTab = new ApplicationTab(session);
+        }
+        public bool SaveAndClose() => false;
+        public bool Back() => false;
+    }
+
+    public class ApplicationTab
```
jalkire: I am aware that a few of these files contain multiple classes per file. I'm pretty split. 
On one hand, the only reason they're in multiple classes is so that when someone uses the library, they get to see the distinction (ie `Settings.ApplicationTab.<application tab specific function>` instead of `Settings.<application tab specific function>`. So in that sense, I do think all of their eventual methods do make sense in one file. 
On the other hand, multiple classes in one file is definitely not the standard. 

I'm curious if anyone agrees with my first point or if these should just be split into multiple files.

src/UITests/UILibrary/Settings.cs
```diff
@@ -0,0 +1,40 @@
...
+        ApplicationTab ApplicationTab { get; }
+        AboutTab AboutTab { get; }
+
+        public Settings(WindowsDriver<WindowsElement> session)
+        {
+            Session = session;
+            AboutTab = new AboutTab(session);
+            ApplicationTab = new ApplicationTab(session);
+        }
+        public bool SaveAndClose() => false;
```
jalkire: There are a couple of currently unused methods in these AI-Win page driver files for demonstration purposes. Again, there are meant to be updated/removed when we actually use the file later in this feature. I'm happy to clean everything currently unused up for this PR if that's seen as important.

src/UITests/UILibrary/Settings.cs
DaveTryon: I generally prefer to have the class name and the file name be identical, but this is an unusual case. Maybe we can put them in a single file for now, then refactor as needed later? As an aside, since "ApplicationTab" is explicitly in the context of settings, should we consider naming the class SettingsApplicationTab? We could also use a namespace, but that feels a little heavy

---
In reply to: [299116031](https://github.com/microsoft/accessibility-insights-windows/pull/408#discussion_r299116031) [](ancestors = 299116031)

DaveTryon:
Looks like a good start to me :)

src/UITests/UILibrary/Settings.cs
jalkire: Sounds good. We can revisit the ApplicationTab name when we add the Settings tests.

PR #411, opened 2019-07-02
Switch UITests to have one TestMethod per class + improve LoadTestFile test

jalkire:
#### Describe the change
Previously, we believed that we use the ClassInitialize and ClassCleanup attributes to reset the state of ai-win during our UITests. In reality, ai-win was never reset. This PR switches to using one TestMethod per class combined with the TestInitialize and TestCleanup attributes to give us actual control over resetting ai-win. 
It also expands the coverage provided in the LoadTestFile UITest to include some navigation and axe-windows scans. To do this, AutomationIDs are selectively introduced to ai-win.

#### PR checklist

- [ ] Includes UI changes?


src/AccessibilityInsights.SharedUx/Controls/TestTabs/AutomatedChecksControl.xaml
```diff
@@ -61,7 +61,11 @@
                 </Grid.ColumnDefinitions>
```
jalkire: All changes to the AccessibilityInsights and SharedUx projects are scoped to adding AutomationIDs.

src/UITests/GettingStartedPage.cs
```diff
@@ -9,29 +9,33 @@ public class GettingStartedPage : AIWinSession
     {
         [TestMethod]
         [TestCategory("NoStrongName")]
-        public void VerifyGettingStartedTitle()
+        public void GettingStartedPageTests()
```
DaveTryon: >GettingStartedPageTests [](start = 20, length = 23)

Name is wrong if this is intended to be a ctor

DaveTryon: >GettingStartedPageTests [](start = 20, length = 23)

At first I thought this was a ctor. Can we make the name clearer?

src/UITests/LoadTestFile.cs
```diff
@@ -10,35 +12,45 @@ namespace UITests
     [TestClass]
     public class LoadTestFile : AIWinSession
     {
+        static readonly string TestFileName = "WildlifeManagerTest.a11ytest";
+        static readonly string TestFilePath = $"{Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location)}\\TestFiles";
+
         [TestMethod]
         [TestCategory("NoStrongName")]
-        public void VerifyResultsInUIATreePage()
+        public void LoadTestFileTests()
```
DaveTryon: >LoadTestFileTests [](start = 20, length = 17)

How can we make clearer what this method is? Better name and/or comments?

src/UITests/UILibrary/LiveMode.cs
```diff
@@ -1,17 +1,46 @@
...
 using OpenQA.Selenium.Appium.Windows;
+using UITests.Utilities;
 
 namespace UITests.UILibrary
 {
     public class LiveMode
     {
         WindowsDriver<WindowsElement> Session;
 
+        const string OpenFileToolBarAutomationID = "1001";
```
DaveTryon: >1001 [](start = 52, length = 4)

Can we please document how these numbers were computed, in case they need to be recomputed in the future?

src/UITests/LoadTestFile.cs
jalkire: I'm open to suggestions for the name. I went with <name of test class>Tests because that seemed like the most succinct name for the method that executes the test class's tests. 

A comment could be helpful--something like "The method that executes all of this test class's tests"? Or do we want to add an explanation of why we're taking this approach: "In order to better control order of test execution, this method is the only TestMethod in this class and calls methods which run the actual test arrange/act/assert code"?

src/UITests/AIWinSession.cs
```diff
@@ -18,13 +18,14 @@ public class AIWinSession
...
 
-        private static WindowsDriver<WindowsElement> session;
-        protected static AIWinDriver driver;
-        protected static int testAppProcessId;
+        private WindowsDriver<WindowsElement> session;
+        protected AIWinDriver driver;
+        protected int testAppProcessId;
+        public TestContext TestContext { get; set; }
 
         protected static IList<EventLogEntry> Events { get; } = new List<EventLogEntry>();
```
karanbirsingh: we should un-static this too

src/UITests/GettingStartedPage.cs
```diff
@@ -9,29 +9,33 @@ public class GettingStartedPage : AIWinSession
...
             Assert.AreEqual("Accessibility Insights for Windows - ", driver.Title);
         }
 
-        [TestMethod]
-        [TestCategory("NoStrongName")]
-        public void VerifyAccessibility()
+        private void VerifyAccessibility()
         {
-            var result = driver.ScanAIWin();
+            var result = driver.ScanAIWin(TestContext.ResultsDirectory);
```
karanbirsingh: Out of curiosity, do we know if this is sufficient to percolate the a11ytest file to ADO if there's a failure?

src/UITests/GettingStartedPage.cs
```diff
@@ -9,29 +9,33 @@ public class GettingStartedPage : AIWinSession
...
             Assert.AreEqual(0, result.errors);
         }
 
-        [ClassInitialize]
-        public static void ClassInitialize(TestContext context)
+        [TestInitialize]
+        public void TestInitialize()
         {
-            Setup(context);
+            Setup();
```
karanbirsingh: When the checks pass it would be great to see if the video includes navigation. If not, I wonder if it matters enough to us to put navigation inside the main test method instead

src/UITests/UILibrary/LiveMode.cs
```diff
@@ -1,17 +1,46 @@
...
+                folderTextbox.SendKeys(folder);
+                folderTextbox.SendKeys(OpenQA.Selenium.Keys.Enter);
+
+                var fileTextbox = Session.FindElementByAccessibilityId(OpenFileFileTextBoxAutomationID);
+                fileTextbox.SendKeys(fileName);
+                fileTextbox.SendKeys(OpenQA.Selenium.Keys.Enter);
+
+                return true;
+            }
+            catch
```
karanbirsingh: are there examples of what this is guarding against?

src/UITests/Utilities/ExtenstionMethods.cs
```diff
@@ -0,0 +1,31 @@
...
+        /// <param name="driver">WinAppDriver driver session</param>
+        /// <param name="automationID">AutomationID of element to click</param>
+        /// <returns>True if element was found and clicked without exception; false otherwise</returns>
+        internal static bool FindAndClickByAutomationID(this WindowsDriver<WindowsElement> driver, string automationID)
+        {
+            try
+            {
+                var btn = driver.FindElementByAccessibilityId(automationID);
+
+                btn.Click();
```
karanbirsingh: the PR view makes it look like there are extra blank lines here. If there are, can we remove them?

src/UITests/LoadTestFile.cs
karanbirsingh: maybe a comment / name with the term(s) 'test scenario entry point'?

src/UITests/UILibrary/LiveMode.cs
```diff
@@ -1,17 +1,46 @@
...
+            {
+                return false;
```
DaveTryon: >false [](start = 23, length = 5)

I have mixed feelings on returning bool. Would the code be simpler if we simply let the Exception leak through to the unit test? The unit test would still fail, but without any need of error handling in the unit test itself

src/UITests/UILibrary/LiveMode.cs
```diff
@@ -1,17 +1,46 @@
...
+
         public LiveMode(WindowsDriver<WindowsElement> session)
         {
             Session = session;
         }
-        public bool OpenFile(string folder, string fileName) => false;
+
+        public bool OpenFile(string folder, string fileName)
+        {
+            if (!Session.FindAndClickByAutomationID(AutomationIDs.MainWinLoadButton)) return false;
```
DaveTryon: >FindAndClickByAutomationID [](start = 25, length = 26)

I'd tend to call it a programmer error if these are ever called in a context where they don't exist. Test should fail so that the code owner has to go fix it

src/UITests/UILibrary/LiveMode.cs
jalkire: My thinking with the bools is they gives us control over whether or not we enforce that a navigation event succeeds. The primary scenario I had in mind was the telemetry dialog--we want to dismiss it if it's there, but we don't care if it isn't (when we aren't testing it). With this syntax, you can do `Assert.IsTrue(<navigation>)` if you want to require the event to succeed, and just call it without Assert if the navigation is not really part of the test. 

That being said, I had that idea before we finalized our plan to control the user config in these tests. I can no longer think of a scenario where a navigation event doesn't need to happen successfully, so I think you're right that these don't give us a benefit.

src/UITests/UILibrary/LiveMode.cs
jalkire: See my other comment about the bool returns on the navigation methods. It's not really guarding against something so much as changing exceptions to a bool "has succeeded?" return, but I'm just going to get rid of these and let the methods throw exceptions per @DaveTryon's suggestion.

src/UITests/UILibrary/LiveMode.cs
jalkire: Again, the idea was to make the return value reflect success and leaving it up to the specific test to decide if a failure of this method was a failure of the test. Will change that style per my other comment :)

src/UITests/GettingStartedPage.cs
jalkire: Agreed, let's change this if it doesn't show up in the video.

src/UITests/GettingStartedPage.cs
jalkire: I do not know.

src/UITests/GettingStartedPage.cs
jalkire: Comment added

src/UITests/Utilities/ExtenstionMethods.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
```
karanbirsingh: I didn't see this earlier, but the file name is misspelled

PR #413, opened 2019-07-03
Adding a11ytest file to output of UI tests, clean up some lingering code

karanbirsingh:
This change adds the a11y test file to UI test output, similar to how we add event logs. If a UI test fails because of an accessibility issue, the a11y test file should show up in the ADO attachments for that test.

We assume a 1:1 mapping between our test methods and a11y test files because any failure will stop test execution. If this changes in the future, we will need to ensure we can name each a11y test file differently. We could do this now via renaming / copying, but it would be nicer if Axe.Windows could accept a file name for the result file.

Here is an [example](https://dev.azure.com/accessibility-insights/Accessibility%20Insights/_build/results?buildId=1819&view=ms.vss-test-web.build-test-results-tab) of a failing accessibility issue.

-----

This PR also removes some unnecessary lifecycle code for `session`. Since it is no longer static, we can remove those lines. 

The event serialization was updated to handle a case where WinAppDriver was not running locally - this should be inconclusive, but locally it failed because test cleanup expected a valid event path.

### DIFF
--- a/src/AccessibilityInsights/Modes/StartUpModeControl.xaml
+++ b/src/AccessibilityInsights/Modes/StartUpModeControl.xaml
@@ -156,7 +156,7 @@
             <CheckBox Grid.Column="0" Content="{x:Static properties:Resources.ckbxDontShowContent}" Name="ckbxDontShow" FontSize="12" Foreground="{DynamicResource ResourceKey=DarkGreyTextBrush}" 
                       FontWeight="SemiBold" Style="{StaticResource CkbxRightSide}" VerticalAlignment="Center"
                       FocusVisualStyle="{DynamicResource {x:Static SystemParameters.FocusVisualStyleKey}}"/>
-            <Button Grid.Column="2" x:Name="btnExit"  Click="btnExit_Click" UseLayoutRounding="True" AutomationProperties.Name="Button" AutomationProperties.AutomationId="{x:Static sharedProps:AutomationIDs.StartUpModeExitButton}"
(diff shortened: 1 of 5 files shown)


src/AccessibilityInsights/Modes/StartUpModeControl.xaml
```diff
@@ -156,7 +156,7 @@
             <CheckBox Grid.Column="0" Content="{x:Static properties:Resources.ckbxDontShowContent}" Name="ckbxDontShow" FontSize="12" Foreground="{DynamicResource ResourceKey=DarkGreyTextBrush}" 
                       FontWeight="SemiBold" Style="{StaticResource CkbxRightSide}" VerticalAlignment="Center"
                       FocusVisualStyle="{DynamicResource {x:Static SystemParameters.FocusVisualStyleKey}}"/>
-            <Button Grid.Column="2" x:Name="btnExit"  Click="btnExit_Click" UseLayoutRounding="True" AutomationProperties.AutomationId="{x:Static sharedProps:AutomationIDs.StartUpModeExitButton}"
+            <Button Grid.Column="2" x:Name="btnExit"  Click="btnExit_Click" UseLayoutRounding="True" AutomationProperties.Name="Button" AutomationProperties.AutomationId="{x:Static sharedProps:AutomationIDs.StartUpModeExitButton}"
```
karanbirsingh: for testing purposes only, will remove in future iteration

PR #416, opened 2019-07-03
temporarily skip UI tests in the compliance build

karanbirsingh:
We need to get the Screen Resolution Utility extension in the account where the compliance build runs. Until this happens, the UI tests must be skipped. The extension is available and running with our PR builds.


jalkire:
In axe-windows, we created a static class with test category strings so that we wouldn't have to type out the category each time and risk a typo. Might be worth considering that approach here

PR #418, opened 2019-07-05
Add initial UI tests for loading an event

karanbirsingh:
This PR adds an initial UI test for event-loading. It opens a saved event file, clicks to show certain events, and then verifies the number of rows in the window. Getting more granular (e.g. how many property rows vs event detail rows) was difficult due to WinAppDriver oddities. We may explore that further in a future PR. For now, we validate the number of rows across various views and verify there are no accessibility issues. This PR only touches test code and automation IDs in our UI.

### DIFF
--- a/src/AccessibilityInsights.SharedUx/Controls/EventDetailControl.xaml
+++ b/src/AccessibilityInsights.SharedUx/Controls/EventDetailControl.xaml
@@ -22,7 +22,7 @@
                     <ColumnDefinition Width="{Binding Path=Width, 
            RelativeSource={RelativeSource Mode=FindAncestor, AncestorType=UIElement},Mode=OneWayToSource}"/>
                 </Grid.ColumnDefinitions>
-                <customcontrols:CustomDataGrid x:Name="dgEvents" AutomationProperties.AutomationId="dgEventDetails" GotKeyboardFocus="dgEvents_GotKeyboardFocus"
+                <customcontrols:CustomDataGrid x:Name="dgEvents" GotKeyboardFocus="dgEvents_GotKeyboardFocus"
                   AutomationProperties.Name="{x:Static Properties:Resources.EventDetailControlAutomationPropertiesName}"
                   Style="{StaticResource dgStyle}" CellStyle="{StaticResource dgcStyle}" RowStyle="{StaticResource dgrStyle}" ColumnHeaderStyle="{StaticResource dgchStyle}">
                     <i:Interaction.Behaviors>
--- a/src/AccessibilityInsights.SharedUx/Controls/EventDetailControl.xaml
+++ b/src/AccessibilityInsights.SharedUx/Controls/EventDetailControl.xaml
@@ -22,7 +22,7 @@
                     <ColumnDefinition Width="{Binding Path=Width, 
            RelativeSource={RelativeSource Mode=FindAncestor, AncestorType=UIElement},Mode=OneWayToSource}"/>
                 </Grid.ColumnDefinitions>
-                <customcontrols:CustomDataGrid x:Name="dgEvents" GotKeyboardFocus="dgEvents_GotKeyboardFocus"
+                <customcontrols:CustomDataGrid x:Name="dgEvents" AutomationProperties.AutomationId="dgEventDetails" GotKeyboardFocus="dgEvents_GotKeyboardFocus"
                   AutomationProperties.Name="{x:Static Properties:Resources.EventDetailControlAutomationPropertiesName}"
                   Style="{StaticResource dgStyle}" CellStyle="{StaticResource dgcStyle}" RowStyle="{StaticResource dgrStyle}" ColumnHeaderStyle="{StaticResource dgchStyle}">
                     <i:Interaction.Behaviors>
--- a/src/UITests/LoadEventFile.cs
+++ b/src/UITests/LoadEventFile.cs
@@ -12,8 +12,8 @@ namespace UITests
     [TestClass]
     public class LoadEventFile : AIWinSession
     {
-        static readonly string EventFileName = "WildlifeManagerTest.a11yevent";
-        static readonly string TestFilePath = $"{Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location)}\\TestFiles";
+        const string EventFileName = "WildlifeManagerTest.a11yevent";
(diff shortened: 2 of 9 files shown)


src/UITests/LoadEventFile.cs
```diff
@@ -0,0 +1,109 @@
...
+
+        /// <summary>
+        /// Validate the event details pane in the bottom right
+        /// </summary>
+        private void CheckEventDetails()
+        {
+            var detailGrid = driver.FindElementByAccessibilityId("dgEventDetails");
+            var rows = detailGrid.FindElementsByClassName("DataGridRow");
+
+            // rows.Count is 0, need to understand why
```
karanbirsingh: @jalkire let's chat about these when you're back in

src/UITests/LoadEventFile.cs
```diff
@@ -0,0 +1,109 @@
...
+            // patternItems.Count is 0, need to understand why
+            // Assert.AreEqual(3, patternItems.Count);
+        }
+
+        [TestInitialize]
+        public void TestInitialize()
+        {
+            Setup();
+
+            driver.GettingStarted.DismissTelemetry();
```
karanbirsingh: @jalkire flagging to chat about this & whether/where to put a `OpenFileFromStart` method that includes these four lines

src/UITests/LoadEventFile.cs
jalkire: Maybe you already found this, but if you scope to the entire event mode control (AutomationId=ctrlEventMode), you find these 3 rows as well as the other 10. If you scope to the left side (AutomationId=ctrlEvents), you find the other 10. If you scope to the right side (AutomationId=ctrlTabs), you find 0. So the 3 rows we care about here disappear as soon as you scope to a parent below the entire event mode. If we can't figure out why this is happening, we can at least use that higher scoping to verify the count is 10 + 3...

src/UITests/LoadEventFile.cs
jalkire: I could see something like `driver.EnterLiveModeNoHighlighter` in the main driver class to get through the first 3 lines. Not sure I think we should put the 4th line inside another method.

src/UITests/LoadEventFile.cs
```diff
@@ -0,0 +1,109 @@
...
+        /// <summary>
+        /// The entry point for this test scenario. Every TestMethod  will restart ai-win, so
+        /// we want to use them sparingly.
+        /// </summary>
+        [TestMethod]
+        [TestCategory("NoStrongName")]
+        [TestCategory("UITest")]
+        public void LoadEventFileTests()
+        {
+            driver.Maximize();
```
jalkire: Do we need this to make sure clicking happens correctly across different screens?

karanbirsingh: This is just to make it easier to see everything because the screen is kind of cramped - helps with the video

src/AccessibilityInsights.SharedUx/Controls/EventDetailControl.xaml
```diff
@@ -22,7 +22,7 @@
                     <ColumnDefinition Width="{Binding Path=Width, 
            RelativeSource={RelativeSource Mode=FindAncestor, AncestorType=UIElement},Mode=OneWayToSource}"/>
                 </Grid.ColumnDefinitions>
-                <customcontrols:CustomDataGrid x:Name="dgEvents" GotKeyboardFocus="dgEvents_GotKeyboardFocus"
+                <customcontrols:CustomDataGrid x:Name="dgEvents" AutomationProperties.AutomationId="dgEventDetails" GotKeyboardFocus="dgEvents_GotKeyboardFocus"
```
jalkire: This AutomationId should be in `AutomationIDs`

src/UITests/LoadEventFile.cs
```diff
@@ -0,0 +1,109 @@
...
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System.IO;
+using System.Reflection;
+
+namespace UITests
+{
+    /// <summary>
+    /// Start the app, open an A11yEvent file, spot check loaded data, exit app
+    /// </summary>
+    [TestClass]
```
RobGallo: @jalkire , Should these tests use the "Integration" category you introduced?

src/UITests/LoadEventFile.cs
```diff
@@ -0,0 +1,109 @@
...
+    public class LoadEventFile : AIWinSession
+    {
+        static readonly string EventFileName = "WildlifeManagerTest.a11yevent";
```
RobGallo:
```suggestion
        const string EventFileName = "WildlifeManagerTest.a11yEvent";
```

src/UITests/LoadEventFile.cs
```diff
@@ -0,0 +1,109 @@
...
+        static readonly string TestFilePath = $"{Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location)}\\TestFiles";
```
RobGallo: Instead of formatting the path like this, look at System.IO.Path.Combine(...)

src/UITests/LoadEventFile.cs
```diff
@@ -0,0 +1,109 @@
...
+        {
+            Setup();
+
+            driver.GettingStarted.DismissTelemetry();
+            driver.GettingStarted.DismissStartupPage();
+            driver.ToggleHighlighter();
+            driver.LiveMode.OpenFile(TestFilePath, EventFileName);
+        }
+
+        [TestCleanup]
```
RobGallo: Is this method necessary? Seems like you could put the same attribute on the Teardown method.

src/UITests/LoadEventFile.cs
jalkire: Yes, except that the Integration category is part of axe-windows. I suggested adding TestCategories in the same style to ai-win in a previous PR, but we haven't gotten around to it yet

src/UITests/LoadEventFile.cs
karanbirsingh: I'm hesitant to add the attribute to TearDown in case there are other E2E tests that may need additional prework before calling the parent TearDown method

src/UITests/LoadEventFile.cs
karanbirsingh: Using the totals suggestion

PR #419, opened 2019-07-08
Change WinAppDriver connection to retry instead of having a long delay

jalkire:
#### Describe the change
We gave the UITests a loongg delay so that they would run on the ado build. This PR replaces that delay with a repeated retry so that quicker test machines don't have to wait as long to run ui tests. 

#### PR checklist

- [ ] Includes UI changes?

### DIFF
--- a/src/UITests/AIWinSession.cs
+++ b/src/UITests/AIWinSession.cs
@@ -60,18 +60,46 @@ namespace UITests
             var configPathArgument = $"--ConfigFolder \"{Path.Combine(TestContext.TestResultsDirectory, TestContext.TestName)}\"";
 
             process = Process.Start(exePath, configPathArgument);
-            process.WaitForInputIdle();
 
-            // small buffer between splash screen disappearing 
-            // and main window initializing; otherwise in rare
-            // cases splash screen can be picked up as main window
-            Thread.Sleep(30000);
+            const int attempts = 15; // this number should give us enough retries for the build to work
+            RetryAttemptToStartNewSession(attempts);
 
+            driver = new AIWinDriver(session, process.Id);
+        }
+
+        /// <summary>
+        /// We can't start a WinAppDriver session until ai-win is past its splash screen and fully loaded.
+        /// This takes a variable amount of time depending on the executing machine. It is particularly slow
+        /// in our build pipeline. Rather than set a long delay for the worst case scenario, we instead attempt
+        /// to start a new session repeatedly until ai-win is ready.
+        /// </summary>
+        /// <param name="attempts">Number of times to retry starting a new session</param>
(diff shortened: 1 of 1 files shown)


src/UITests/AIWinSession.cs
```diff
@@ -60,20 +60,49 @@ private void LaunchApplicationAndAttach()
...
+        {
+            while (attempts > 0)
+            {
+                attempts--;
+
+                StartNewSession();
+                Thread.Sleep(3000);
+
+                // if the session and its title are present, ai-win is ready for testing.
+                if (!string.IsNullOrEmpty(session?.Title))
```
karanbirsingh: How about moving this condition into the while loop initialization as well?

src/UITests/AIWinSession.cs
```diff
@@ -60,20 +60,49 @@ private void LaunchApplicationAndAttach()
...
-            // small buffer between splash screen disappearing 
-            // and main window initializing; otherwise in rare
-            // cases splash screen can be picked up as main window
-            Thread.Sleep(30000);
-
-            DesiredCapabilities appCapabilities = new DesiredCapabilities();
-            appCapabilities.SetCapability("appTopLevelWindow", process.MainWindowHandle.ToString("x"));
-            session = new WindowsDriver<WindowsElement>(new Uri(WindowsApplicationDriverUrl), appCapabilities);
+            const int attempts = 10; // this number should give us enough retries for the build to work
+            RetryAttemptToStartNewSession(attempts);
```
flyingUnderTheRadar: I wonder if it would make sense to just call it `StartNewSession` with retrying being abstracted into that.  🔕

jalkire: Renamed

src/UITests/AIWinSession.cs
jalkire: Moved

src/UITests/AIWinSession.cs
jalkire: Renamed

src/UITests/AIWinSession.cs
```diff
@@ -60,20 +60,44 @@ private void LaunchApplicationAndAttach()
...
+                StartNewSession();
+                Thread.Sleep(3000);
+            }
+        }
+
+        private void StartNewSession()
+        {
+            try
+            {
+                process.Refresh(); // updates process.MainWindowHandle
```
RobGallo: Food for thought: I tend to use the '_' char to connote private fields within a class. So, "_process" for example. This helps distinguish between local and member variables. I think it's especially useful when the member variable is used in several places throughout a class.

jalkire: Underscore adopted

PR #420, opened 2019-07-09
Add a UI test for the settings page

karanbirsingh:
This UI test includes the following:
- spot-checking the application settings tab
- accessibility testing of the application settings tab
- spot-checking the about tab
- accessibility testing of the about tab
- changing a keyboard shortcut and validating it works
- validating the save and close button becomes enabled with changed options

Here's a video: https://dev.azure.com/accessibility-insights/Accessibility%20Insights/_build/results?buildId=1880&view=ms.vss-test-web.build-test-results-tab&runId=5176&resultId=100147&paneView=attachments

### DIFF
--- a/src/AccessibilityInsights.SharedUx/Properties/AutomationIDs.cs
+++ b/src/AccessibilityInsights.SharedUx/Properties/AutomationIDs.cs
@@ -13,5 +13,15 @@ namespace AccessibilityInsights.SharedUx.Properties
         public static string TelemetryDialogExitButton { get; } = nameof(TelemetryDialogExitButton);
         public static string EventsDataGrid { get; } = nameof(EventsDataGrid);
         public static string EventModeControl { get; } = nameof(EventModeControl);
+        public static string SettingsButton { get; } = nameof(SettingsButton);
+        public static string SettingsApplicationTabItem { get; } = nameof(SettingsApplicationTabItem);
+        public static string SettingsAboutTabItem { get; } = nameof(SettingsAboutTabItem);
+        public static string ChannelComboBox { get; } = nameof(ChannelComboBox);
+        public static string SaveAndCloseButton { get; } = nameof(SaveAndCloseButton);
+        public static string PauseButton { get; } = nameof(PauseButton);
+        public static string HotkeyPauseButton { get; } = nameof(HotkeyPauseButton);
+        public static string HotkeyGrabDialog { get; } = nameof(HotkeyGrabDialog);
+        public static string MainWindow { get; } = nameof(MainWindow);
+        public static string ThirdPartyNoticesHyperlink { get; } = nameof(ThirdPartyNoticesHyperlink);
     }
 }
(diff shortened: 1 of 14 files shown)


src/UITests/UILibrary/AIWinDriver.cs
```diff
@@ -50,7 +50,9 @@ public int ScanAIWin(TestContext context)
             var result = scanner.Scan();
             if (result.ErrorCount > 0)
             {
-                context.AddResultFile(result.OutputFile.A11yTest);
+                var newPath = Path.Combine(outputPath, $"{fileName}.a11ytest");
```
DaveTryon: FxCop may complain about this. The fix is:
```suggestion
                var newPath = Path.Combine(outputPath, Invariant($"{fileName}.a11ytest"));
```

and to add the following at the top of the file (line 9, maybe):

using static System.FormattableString;

src/UITests/UILibrary/AIWinDriver.cs
```diff
@@ -62,6 +64,9 @@ public int ScanAIWin(TestContext context)
 
         public void ToggleHighlighter() => Session.FindElementByAccessibilityId(AutomationIDs.MainWinHighlightButton).Click();
 
+        public void GoToSettings() => Session.FindElementByAccessibilityId(AutomationIDs.SettingsButton).Click();
+
         public void Maximize() => Session.Manage().Window.Maximize();
+
```
DaveTryon: Is this empty line intentional?

src/UITests/UILibrary/AIWinDriver.cs
karanbirsingh: updated, thanks for the suggestion

src/UITests/UILibrary/AIWinDriver.cs
karanbirsingh: mistake - I've removed it

src/AccessibilityInsights.SharedUx/Properties/AutomationIDs.cs
```diff
@@ -13,5 +13,15 @@ public static class AutomationIDs
         public static string TelemetryDialogExitButton { get; } = nameof(TelemetryDialogExitButton);
```
karanbirsingh: John and I will continue with this AutomationIDs strategy and reassess later in the feature whether we like it, want to group better, etc

src/AccessibilityInsights.SharedUx/Properties/AutomationIDs.cs
```diff
@@ -13,5 +13,15 @@ public static class AutomationIDs
...
         public static string EventsDataGrid { get; } = nameof(EventsDataGrid);
         public static string EventModeControl { get; } = nameof(EventModeControl);
+        public static string SettingsButton { get; } = nameof(SettingsButton);
+        public static string SettingsApplicationTabItem { get; } = nameof(SettingsApplicationTabItem);
+        public static string SettingsAboutTabItem { get; } = nameof(SettingsAboutTabItem);
+        public static string ChannelComboBox { get; } = nameof(ChannelComboBox);
```
jalkire: I've been prefixing these with (at least part of) the name of the control/class which contains them. As you mention, we probably want better organization in the future, so consider it a suggestion to use that naming style for now🙂

PR #421, opened 2019-07-09
Significantly improve LoadTestFile UITest coverage

jalkire:
#### Describe the change
Expands coverage of LoadTestFile UITest to:
- Verify the count of results in listview and textblock above listview
- Expand results and click through to the results in UIA tree view for a specific element
- Spot check UIA tree
- Verify the correct number of results show up and that there is some how to fix text
- Spot check properties and patterns
- Switch to a different element in the UIA tree
- Repeat some validation for newly selected element

### DIFF
--- a/src/AccessibilityInsights.SharedUx/Properties/AutomationIDs.cs
+++ b/src/AccessibilityInsights.SharedUx/Properties/AutomationIDs.cs
@@ -4,11 +4,19 @@ namespace AccessibilityInsights.SharedUx.Properties
 {
     public static class AutomationIDs
     {
+        public static string AutomatedChecksExpandAllButton{ get; } = nameof(AutomatedChecksExpandAllButton);
+        public static string AutomatedChecksResultsListView { get; } = nameof(AutomatedChecksResultsListView);
+        public static string AutomatedChecksResultsTextBlock { get; } = nameof(AutomatedChecksResultsTextBlock);
         public static string AutomatedChecksUIATreeButton { get; } = nameof(AutomatedChecksUIATreeButton);
+        public static string HierarchyControlUIATreeView { get; } = nameof(HierarchyControlUIATreeView);
         public static string MainWinLoadButton { get; } = nameof(MainWinLoadButton);
         public static string MainWinHighlightButton { get; } = nameof(MainWinHighlightButton);
         public static string MainWinBreadCrumbOneButton { get; } = nameof(MainWinBreadCrumbOneButton);
         public static string MainWinBreadCrumbTwoButton { get; } = nameof(MainWinBreadCrumbTwoButton);
+        public static string ScannerResultsDetailsListView { get; } = nameof(ScannerResultsDetailsListView);
+        public static string ScannerResultsFixFollowingTextBox { get; } = nameof(ScannerResultsFixFollowingTextBox);
+        public static string ScannerResultsShowAllButton { get; } = nameof(ScannerResultsShowAllButton);
+        public static string SnapshotModeControl { get; } = nameof(SnapshotModeControl);
         public static string StartUpModeExitButton { get; } = nameof(StartUpModeExitButton);
         public static string TelemetryDialogExitButton { get; } = nameof(TelemetryDialogExitButton);
         public static string EventsDataGrid { get; } = nameof(EventsDataGrid);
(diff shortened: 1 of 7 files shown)


src/AccessibilityInsights.SharedUx/Properties/AutomationIDs.cs
```diff
@@ -4,11 +4,19 @@ namespace AccessibilityInsights.SharedUx.Properties
 {
     public static class AutomationIDs
     {
+        public static string AutomatedChecksExpandAllButton{ get; } = nameof(AutomatedChecksExpandAllButton);
```
DaveTryon: >AutomatedChecksExpandAllButton [](start = 29, length = 30)

Should we try to standardize both the names of the IDs ("AutomatedChecksExpandAllButton" has some context but "PauseButton" doesn't). Also, some (but not all) of these are sorted alphabetically. It might make it easier in the long run if we had a single approach (not necessarily part of this PR)

src/UITests/LoadTestFile.cs
```diff
@@ -12,20 +18,22 @@ namespace UITests
     [TestClass]
     public class LoadTestFile : AIWinSession
     {
-        static readonly string TestFileName = "WildlifeManagerTest.a11ytest";
-        static readonly string TestFilePath = $"{Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location)}\\TestFiles";
+        readonly string TestFileName = "WildlifeManagerTest.a11ytest";
```
DaveTryon:
```suggestion
        const string TestFileName = "WildlifeManagerTest.a11ytest";
```

src/UITests/LoadTestFile.cs
```diff
@@ -12,20 +18,22 @@ namespace UITests
...
+        readonly string TestFilePath = $"{Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location)}\\TestFiles";
```
DaveTryon: const instead of readonly?

src/UITests/LoadTestFile.cs
```diff
@@ -12,20 +18,22 @@ namespace UITests
...
```
karanbirsingh: `Path.Combine` was suggested in another PR to me for this TestFilePath line

src/AccessibilityInsights.SharedUx/Properties/AutomationIDs.cs
jalkire: Agreed. We are planning to do standardize/improve this in a future PR.

src/UITests/LoadTestFile.cs
jalkire: Changed to const where possible

src/UITests/LoadTestFile.cs
```diff
@@ -44,6 +52,120 @@ private void ScanAutomatedChecks()
...
+            var tree = driver.FindElementByAccessibilityId(AutomationIDs.HierarchyControlUIATreeView);
+            var nodes = tree.FindElementsByClassName("TreeViewItem");
+            var patterns = GetPatternsNodes(AutomationIDs.SnapshotModeControl, nodes);
+
+            Assert.AreEqual("Name, Desktop 1", text);
+            Assert.AreEqual(10, patterns.Count());
+            Assert.AreEqual("LegacyIAccessiblePattern", patterns.First().Text);
+            Assert.AreEqual(10, props.Count);
+
+            patterns.Last().SendKeys(Keys.Control + Keys.Tab);
```
karanbirsingh: Having trouble understanding what this does since it's at the end of the method. If it is required for a different test, can we move it into the beginning of the other test?

jalkire: This switches the tab on the right side of ai-win. I could do it at the start of the other test (`ValidateRootResults`), but either way we are making an assumption about the current state of the tabs (currently, we assume the details tab is selected. With the suggested change, we would instead assume the results tab is selected). The benefit to the current way is that we don't have to go and get an element to ctrl + tab on since we already have one here.

karanbirsingh: That makes sense, thanks. Can you add a comment above the line in this or a future PR? I think it might help

jalkire: Comment added!

PR #426, opened 2019-07-10
Add basic live mode UITest

jalkire:
#### Describe the change
This UITest for live mode launches WildlifeManager, runs ai-win tests, and does basic validation and axe-windows scans. If the PR build succeeds, this will be a good starting point off of which to build.

Find a screencap [here](https://dev.azure.com/accessibility-insights/Accessibility%20Insights/_apis/test/Runs/5300/Results/100017/Attachments/3)

### DIFF
--- a/src/AccessibilityInsights.SharedUx/Controls/InspectTabsControl.xaml
+++ b/src/AccessibilityInsights.SharedUx/Controls/InspectTabsControl.xaml
@@ -55,8 +55,8 @@
                            FontWeight="Bold" 
                            TextWrapping="Wrap"
                            TextTrimming="CharacterEllipsis"
-                           Height="30" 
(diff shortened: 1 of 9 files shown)


src/AccessibilityInsights.SharedUx/Controls/InspectTabsControl.xaml
```diff
@@ -55,8 +55,8 @@
                            FontWeight="Bold" 
                            TextWrapping="Wrap"
                            TextTrimming="CharacterEllipsis"
-                           Height="30" 
-                           AutomationProperties.Name="{x:Static Properties:Resources.tbElementAutomationPropertiesName}"/>
```
jalkire: As far as I can tell, setting AutomationProperties.Name here was a bug. All it did was replace the actual text with "Glimpse" for AT users.

PR #427, opened 2019-07-11
Expand LiveMode UITest Coverage

jalkire:
#### Describe the change
Expand the live mode UI tests to spot check the test results in a similar way to the LoadTestFile test. Extracted some validation code out from the LoadTestFile tests to share with the added tests.

Find screencap [here](https://dev.azure.com/accessibility-insights/Accessibility%20Insights/_apis/test/Runs/5366/Results/100141/Attachments/3)

### DIFF
--- a/src/UITests/LiveMode.cs
+++ b/src/UITests/LiveMode.cs
@@ -4,7 +4,6 @@ using Microsoft.VisualStudio.TestTools.UnitTesting;
 using System;
 using System.Diagnostics;
 using System.IO;
-using System.Threading;
 
 namespace UITests
 {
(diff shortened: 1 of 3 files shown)


src/UITests/LiveMode.cs
```diff
@@ -28,13 +27,29 @@ public void LiveModeTests()
...
+            ScanAccessibility();
+            ValidateTestModeTitle();
+            driver.TestMode.AutomatedChecks.ValidateAutomatedChecks(12);
+            driver.TestMode.AutomatedChecks.GoToAutomatedChecksElementDetails(4);
+            ValidateResultsInUIATree();
+        }
+
+        private void ValidateResultsInUIATree()
+        {
+            driver.TestMode.ResultsInUIATree.ValidateResults(false, 2, 11);
```
DaveTryon: >driver [](start = 12, length = 6)

Code smell: all of the lines in this method start "driver.TestMode"--that suggests that perhaps the ValidateResultsInUIATree method belongs in the TestMode class.

jalkire: The parameters passed to the methods called in ValidateResultsInUIATree are specific to this test class and I don’t expect to reuse that exact set of calls often. ValidateResultsInUIATree only exists because I thought having the grouping makes TestTestResults read a little better. 
Does that explanation change your mind, or would you still say this makes more sense in TestMode?

DaveTryon: I'm fine with it either way.

karanbirsingh: I have merged this as-is because John is out until next week - this potential change can come in separately if we'd like

PR #428, opened 2019-07-12
Adding initial ui-test-job template

karanbirsingh:
We're experimenting with creating separate jobs for our UI tests that can run in parallel. We're using YAML templates in order to accomplish this. This PR adds a template for the ui-test-job and executes it from the prebuild file. If it works, we have opportunities to refactor some of our other build scripts via templates.

### DIFF
--- a/build/prbuild.yml
+++ b/build/prbuild.yml
@@ -8,7 +8,8 @@ variables:
   SignAppForRelease: 'false'
 
 jobs:
-- job: Build and unit tests Release
+- job: ReleaseBuildUnitTests
+  displayName: Build and unit tests (release)
   pool:
     vmImage: 'windows-2019'
   variables:
@@ -60,7 +61,8 @@ jobs:
   - task: ms.vss-governance-buildtask.governance-build-task-component-detection.ComponentGovernanceComponentDetection@0
     displayName: 'Component Detection'
 
-- job: Build and unit tests Debug
+- job: DebugBuildUnitTests
+  displayName: Build and unit tests (debug)
   pool:
     vmImage: 'windows-2019'
   variables:
--- a/build/prbuild.yml
+++ b/build/prbuild.yml
@@ -8,7 +8,7 @@ variables:
   SignAppForRelease: 'false'
 
 jobs:
-- job: Build and unit tests (Release)
+- job: Build and unit tests Release
   pool:
     vmImage: 'windows-2019'
   variables:
@@ -60,7 +60,7 @@ jobs:
   - task: ms.vss-governance-buildtask.governance-build-task-component-detection.ComponentGovernanceComponentDetection@0
     displayName: 'Component Detection'
 
-- job: Build and unit tests (Debug)
+- job: Build and unit tests Debug
   pool:
     vmImage: 'windows-2019'
   variables:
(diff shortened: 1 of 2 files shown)


build/prbuild.yml
```diff
@@ -8,6 +8,14 @@ variables:
   SignAppForRelease: 'false'
 
 jobs:
+- template: ui-test-job.yml
```
karanbirsingh: We could put the UI tests after the unit tests instead if that makes more sense

DaveTryon: They'll all run in parallel, but the proposed ordering seems simpler from a "mental organization" perspective. Can we also flag the unit test jobs as unit tests, instead of just Debug and Release?

---
In reply to: [303156978](https://github.com/microsoft/accessibility-insights-windows/pull/428#discussion_r303156978) [](ancestors = 303156978)

build/ui-test-job.yml
```diff
@@ -0,0 +1,48 @@
...
+    displayName: 'Test Assemblies **\release\*test*.dll;-:**\obj\**'
+    inputs:
+      testAssemblyVer2: |
+        **\*test*.dll
+        !**\obj\**
+      vsTestVersion: 16.0
+      codeCoverageEnabled: true
+      platform: '$(BuildPlatform)'
+      configuration: ${{ parameters.configuration }}
+      rerunFailedTests: true
```
DaveTryon: >rerunFailedTests: true [](start = 6, length = 22)

Retries were added because some unit test timeouts will sporadically be exceeded. Do you think we need them for UI Tests?

build/ui-test-job.yml
```diff
@@ -0,0 +1,48 @@
...
+      AgentResolution: 1080p
+
+  - task: VSTest@2
+    displayName: 'Test Assemblies **\release\*test*.dll;-:**\obj\**'
+    inputs:
+      testAssemblyVer2: |
+        **\*test*.dll
+        !**\obj\**
+      vsTestVersion: 16.0
+      codeCoverageEnabled: true
```
DaveTryon: >codeCoverageEnabled: true [](start = 6, length = 25)

Is code coverage intentional or sort of accidental? I'm not sure it makes sense here...

build/ui-test-job.yml
```diff
@@ -0,0 +1,48 @@
...
+    displayName: 'Build Solution **\*.sln'
+    inputs:
+      vsVersion: 16.0
+      platform: '$(BuildPlatform)'
+      configuration: ${{ parameters.configuration }}
+
+  - task: WinAppDriver.winappdriver-pipelines-task.winappdriver-pipelines-task.Windows Application Driver@0
+    displayName: 'Start - WinAppDriver'
+    inputs:
+      AgentResolution: 1080p
```
DaveTryon: >1080p [](start = 23, length = 5)

I like how this replaces the Set Screen Resolution task

build/prbuild.yml
karanbirsingh: pending

build/ui-test-job.yml
karanbirsingh: accidental, came over in the copy - pending

PR #429, opened 2019-07-12
Simplify YAML structure of PR builds

karanbirsingh:
This PR attempts to simplify the PR YAML build by creating and using a template for the build & unit test job. This template is called once with debug and again with release. There are template variables for CG and license header checks. There should be no functional change after this PR is merged - it is intended to simply rearrange how the tasks are defined.

### DIFF
--- a/build/build-and-unit-tests.yml
+++ b/build/build-and-unit-tests.yml
@@ -1,0 +1,63 @@
+# This template contains jobs to build and run unit tests
+
+parameters:
+  configuration: ''
+  checkLicenseHeaders: ''
+  runComponentGovernance: ''
+
+jobs:
+- job: BuildAndUnitTests${{ parameters.configuration }}
+  displayName: Build and run unit tests - ${{ parameters.configuration }}
+  condition: succeeded()
+  pool:
+    vmImage: windows-2019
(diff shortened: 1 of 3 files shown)


build/build-and-unit-tests.yml
```diff
@@ -0,0 +1,63 @@
+# This template contains jobs to build and run unit tests
+
+parameters:
+  configuration: ''
+  checkLicenseHeaders: ''
+  runComponentGovernance: ''
+
+jobs:
+- job: BuildAndUnitTests${{ parameters.configuration }}
```
DaveTryon: >BuildAndUnitTests$ [](start = 7, length = 18)

Should this file reflect that it's only for PR builds? The signed build will need a similar template, but has additional TSA-related steps

karanbirsingh: I agree with you and wonder if we can defer the change until the signed build refactoring. Another reason I mention this is that we can also define sequences of steps (without defining an entire job) that can be reused. This means we could compose files used in both the PR and signed build.

DaveTryon: Works for me :)

PR #431, opened 2019-07-15
Modify signed yaml to include UI tests

karanbirsingh:
This YAML change affects the signed build so that we now run the UI tests as well. This was unblocked with the installation of the WinAppDriver extension into our signed build organization.

The compliance YAML was deleted because the steps are duplicated in the signed build.


build/signedbuild.yml
```diff
@@ -11,6 +11,14 @@ variables:
   FAKES_SUPPORTED: 1
 
 jobs:
```
jalkire: I think the SignedRelease job needs these jobs under `dependsOn`

karanbirsingh: Good catch, updated

build/signedbuild.yml
```diff
@@ -184,6 +192,8 @@ jobs:
   dependsOn: 
   - ComplianceRelease
   - ComplianceDebug
+  - UITestsdebug
```
karanbirsingh: from https://github.com/microsoft/accessibility-insights-windows/blob/master/build/ui-test-job.yml

PR #433, opened 2019-07-15
Switch to the offical Appium NuGet package

karanbirsingh:
#### Describe the change
In the UI tests project, we were using a prerelease/temp version of the package from the WinAppDriver team. This code should be pulled from the official Appium package. This PR switches packages. This led to updates of Newtonsoft across the project because Appium requires >=12 and most of our code used 10.0.3. The versions were consolidated to >=12

#### PR checklist

- (saving/loading) Run through of all [test scenarios](https://github.com/Microsoft/accessibility-insights-windows/blob/master/docs/Scenarios.md) completed?
- (no) Does this address an existing issue? If yes, Issue# - 
- (no) Includes UI changes?


src/AccessibilityInsights/MainWindow.xaml.cs
```diff
@@ -159,6 +159,9 @@ public MainWindow()
 
             SetHighContrastTheme();
 
+            ///in case we need to do any debugging with elevated app
+            SupportDebugging();
```
jalkire: Why was this moved? Curious

karanbirsingh: I sometimes hit errors in `PopulateConfigurations` that occurred too early. This way we can step through PopulateConfigurations

PR #439, opened 2019-07-16
Add UITest docs

jalkire:
#### Describe the change
Added UITest docs

### DIFF
--- a/src/UITests/README.md
+++ b/src/UITests/README.md
@@ -1,0 +1,30 @@
+## UI Tests
+* UI tests reside in the `UITests` project.
+* [WinAppDriver](https://github.com/Microsoft/WinAppDriver) is used for navigation and [Axe.Windows](https://github.com/microsoft/axe-windows) is used for accessibility scanning.
+* Test classes inherit from `AIWinSession`.
+* Each test class covers an AI-Win usage scenario. For example, the `LoadTestFile` test covers the scenario of loading a test file and inspecting indiviudal test results. Each test class has at least the following methods:
+  * A `TestMethod` named `NameOfTestClassTests` that performs the desired validation.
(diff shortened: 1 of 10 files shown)


src/UITests/README.md
```diff
@@ -0,0 +1,30 @@
+## UI Tests
+* UI tests reside in the `UITests` project.
+* [WinAppDriver](https://github.com/Microsoft/WinAppDriver) is used for navigation and [Axe.Windows](https://github.com/microsoft/axe-windows) is used for accessibility scanning.
+* Test classes inherit from `AIWinSession`.
+* Each test class covers an AI-Win usage scenario. For example, the `LoadTestFile` test covers the scenario of loading a test file and inspecting indiviudal test results. Each test class has at least the following methods:
```
karanbirsingh: spelling of individual
```suggestion
* Each test class covers an AI-Win usage scenario. For example, the `LoadTestFile` test covers the scenario of loading a test file and inspecting individual test results. Each test class has at least the following methods:
```

src/UITests/README.md
```diff
@@ -0,0 +1,30 @@
+## UI Tests
```
karanbirsingh: I think we should add info about how to run the test (e.g. running winappdriver)

src/UITests/README.md
```diff
@@ -0,0 +1,30 @@
...
+* UI tests reside in the `UITests` project.
```
karanbirsingh: Can we also add brief information about possible artifacts in ADO?
- a11ytest file
- events file
- screen recording

src/UITests/README.md
jalkire: Info added

src/UITests/README.md
jalkire: Info added

PR #443, opened 2019-07-17
Adding assert text messages in UI tests

karanbirsingh:
This PR only modifies the UI Tests. We add messages to the asserts so that errors are more actionable. This is particularly important because a test method includes a full scenario with multiple screens and checks.

### DIFF
--- a/src/UITests/UILibrary/AIWinDriver.cs
+++ b/src/UITests/UILibrary/AIWinDriver.cs
@@ -62,7 +62,7 @@ namespace UITests.UILibrary
         public void VerifyAccessibility(TestContext context, string fileName, int expectedIssueCount)
         {
             var issueCount = this.ScanAIWin(context, fileName);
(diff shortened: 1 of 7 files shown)


src/UITests/UILibrary/AIWinDriver.cs
```diff
@@ -59,6 +59,12 @@ public int ScanAIWin(TestContext context, string fileName)
             return result.ErrorCount;
         }
 
+        public void VerifyAccessibility(TestContext context, string fileName, int expectedIssueCount)
+        {
+            var issueCount = this.ScanAIWin(context, fileName);
+            Assert.AreEqual(expectedIssueCount, issueCount, "axe.windows found accessibility issues, check a11ytest file in test artifacts");
```
DaveTryon: >a11ytest file [](start = 107, length = 13)

Can we include the file name?

karanbirsingh: updated

Issue #425, opened 2019-07-10
[BUG] When the Always on top setting is on, even if windows is locked, the application still stays active


piyushsaxena2910:
**Describe the bug**
Installed the msi on my windows 10 machine. When I lock my machine, the "Accessibility Insights for Windows" always stays on top, even with the locked screen. This only happens when the "Always on top" setting is selected.

**To Reproduce**
1. Install the "Accessibility Insights for Windows" MSI
2. Go to settings and turn on setting "Always on top"
3. Lock windows by pressing Win+L
4. The "Accessibility Insights for Windows" still shows up and it fully accessible

**Expected behavior**
When I lock my machine, the application should not be available.

**Actual behavior**
When I lock my machine, the application is fully available and I can play around with it.

**Screenshots or .GIF**
![image](https://user-images.githubusercontent.com/26318272/60934407-ebf62e00-a27a-11e9-89fe-e20de7ad9e0e.png)

**Desktop (please complete the following information):**
 - OS: [Windows 10 OS Build 17763.557]
 - Accessibility Insights for Windows Version: 1.1.883.1

Sirisha-Kommushetty:
@piyushsaxena2910 - This is "By Design" with Always on Top. Please close the app or turn off the "always on top" setting before locking your device.

piyushsaxena2910:
@Sirisha-Kommushetty Thank you for your response!

PR #444, opened 2019-07-18
Disable UIAccess by default, allow user enabling

DaveTryon:
#### Describe the change
These changes do the following:
1. Make the manifest file for AccessibilityInsights.exe external to the binary so we can manipulate it within the installation.
2. Rename the manifest files
    - app_dev.manifest becomes UIAccess_Disabled.manifest
    - app_UIAccess.manifest becomes UIAccess_Enabled.manifest
3. When building and installing, copy both manifest files to the target folder. Also copy UIAccess_Disabled.manifest to AccessibilityInsights.exe.manifest. UIAccess is disabled by default
4. Add a script (UIAccess.cmd) to allow the user to swap between UIAccess_Enabled and UIAccess_Disabled manifests. Implementation note: The script intentionally rewrites the file, as just copying is unreliable due to an OS manifest caching mechanism.
5. Update the WXS file to copy the 4 new files
6. Update VersionSwitcher.exe to determine the UIAccess state from the manifest files, then carry that state over to the new installation before relaunching the app.

#### PR checklist

- [x] Does this address an existing issue? If yes, Issue# - #425 (the old behavior was by design, but we're altering the design based on feedback)
- [ ] Includes UI changes?


src/AccessibilityInsights.VersionSwitcher/InstallationEngine.cs
```diff
@@ -47,40 +50,87 @@ internal void PerformInstallation()
                 }
             }
             UpdateConfigWithNewChannel(options.NewChannel);
+            SetManifestForUIAccess(options.EnableUIAccess);
             LaunchPostInstallApp();
             EventLogger.WriteInformationalMessage("Completed Installation");
         }
 
         /// <summary>
-        /// Create a CommandLineParameters object based on available input
```
DaveTryon: >CommandLineParameters [](start = 21, length = 21)

Opportunistic correction

src/AccessibilityInsights.VersionSwitcher/InstallationEngine.cs
```diff
@@ -47,40 +50,87 @@ internal void PerformInstallation()
...
+        /// Create an InstallationOptions object based on available input
         /// </summary>
         /// <returns>The populated InstallationOptions object</returns>
         private static InstallationOptions GetInstallationOptions()
         {
-            // Temporary implementation for testing--still need to finalize actual command line
```
DaveTryon: >Temporary [](start = 15, length = 9)

Should have been removed previously

src/AccessibilityInsights.VersionSwitcher/InstallationEngine.cs
```diff
@@ -173,15 +223,15 @@ private void InstallWithinTransaction(InstallationOptions options, Transaction t
         /// <param name="newChannel">The new channel to use</param>
         private static void UpdateConfigWithNewChannel(string newChannel)
         {
-            if (newChannel != null)
-            {
-                var defaultConfigPaths = FixedConfigSettingsProvider.CreateDefaultSettingsProvider();
-                var configFile = Path.Combine(defaultConfigPaths.ConfigurationFolderPath, Constants.AppConfigFileName);
+            if (newChannel == null)
```
DaveTryon: >newChannel [](start = 16, length = 10)

Small opportunistic refactor

src/AccessibilityInsights.VersionSwitcher/InstallationOptions.cs
```diff
@@ -31,17 +31,24 @@ internal class InstallationOptions
...
+        /// Whether or not UIAccess should be enabled post-install
+        /// </summary>
+        internal bool EnableUIAccess { get; }
+
         /// <summary>
         /// ctor
         /// </summary>
         /// <param name="msiPath">String identifying the installer location</param>
         /// <param name="newChannel">String indicating the value for NewChannel</param>
-        internal InstallationOptions(string msiPath, string newChannel = null)
```
DaveTryon: >= null [](start = 71, length = 6)

We never actually used the default parameter

src/AccessibilityInsights/Scripts/UIAccess.cmd
```diff
@@ -0,0 +1,40 @@
...
+ 
+ :Usage
+ echo.
+ echo UIAccess.cmd
+ echo.
+ echo This script will set the UIAccess status of Accessibilty Insights for Windows.
+ echo UIAccess support is disabled by default. This script must be run from an
+ echo administrative command prompt.
+ echo.
+ echo Usage: UIAccess.cmd ^<action^>
```
DaveTryon: >^ [](start = 26, length = 1)

Syntax note: the caret here is an escape sequence.
^< becomes <
^> becomes >

src/AccessibilityInsights/Scripts/UIAccess.cmd
```diff
@@ -0,0 +1,40 @@
...
+ echo Where ^<action^> is one of:
+ echo   ENABLE  (to enable UIAccess support)
+ echo   DISABLE (to disable UIAccess support)
+ echo.
+ goto :eof
+ 
+ :Verify
+ echo.
+ echo CAUTION! You should only change the UIAccess state if you understand the
+ echo          potential security implications!
```
DaveTryon: >potential security implications [](start = 15, length = 31)

This would be better pointing to a web page, but since we don't have one yet, there's no link

src/AccessibilityInsights/Scripts/UIAccess.cmd
```diff
@@ -0,0 +1,40 @@
...
+ choice /m "Are you sure you want to %1 UIAccess support?"
+ if ErrorLevel 1 goto UpdateManifest
+ 
+ echo.
+ echo No change was made
+ echo.
+ goto :eof
+ 
+ :UpdateManifest
+ type UIAccess_%1d.manifest > AccessibilityInsights.exe.manifest
```
DaveTryon: >type [](start = 1, length = 4)

I'll add a comment about why this uses type and not just copy

src/AccessibilityInsights/Scripts/UIAccess.cmd
```diff
@@ -0,0 +1,42 @@
...
+ echo.
+ echo Operation was canceled.
+ echo.
+ goto :eof
+ 
+ :UpdateManifest
+ rem DO NOT use copy here, as we also need to update the file's last write time
+ type UIAccess_%1d.manifest > AccessibilityInsights.exe.manifest
+ echo.
+ echo Update complete. Changes will take effect when Accessibilty Insights for
```
jalkire: I don't know if we care enough about this script to fix this case, but if the type command fails we still get this success message. I hit it because I attempted to run the script without CDing into the folder:
image

DaveTryon:
One test is stuck. Closing and reopening to restart the validation

src/AccessibilityInsights/Scripts/UIAccess.cmd
flyingUnderTheRadar: Same with non admin command prompt window. It does however also say that `Access is denied`
image

DaveTryon: @flyingUnderTheRadar, I've updated the script to cover both the case where it's run from a different folder, and the case where it's run without administrative access. I also fixed the script to not update if the user says "No" at the confirm dialog.

---
In reply to: [305437587](https://github.com/microsoft/accessibility-insights-windows/pull/444#discussion_r305437587) [](ancestors = 305437587)

DaveTryon: I've fixed this

---
In reply to: [305415488](https://github.com/microsoft/accessibility-insights-windows/pull/444#discussion_r305415488) [](ancestors = 305415488)

src/AccessibilityInsights/Scripts/UIAccess.cmd
```diff
@@ -0,0 +1,49 @@
...
+:UserCanceled
+echo.
+echo Operation was canceled.
+goto :eof
+
+:UpdateManifest
+rem DO NOT use copy here, as we also need to update the file's last write time.
+rem An "Access is denied" failure will not set errorlevel without help. The
+rem extra echo statements provide that help
+type UIAccess_%1d.manifest > AccessibilityInsights.exe.manifest && echo. || echo.
```
DaveTryon: >&& echo. || echo. [](start = 64, length = 17)

This was based on [https://stackoverflow.com/questions/10354016/file-redirection-in-windows-and-errorlevel](https://stackoverflow.com/questions/10354016/file-redirection-in-windows-and-errorlevel). It might be possible to make it a little simpler, but the comments left enough doubt in my mind that I stuck with what should always work.

src/AccessibilityInsights/Scripts/UIAccess.cmd
flyingUnderTheRadar: LGTM ! I tried running it from a different folder and with non admin. Works as it is supposed to

PR #508, opened 2019-08-28
Keep textbox placeholder text on keyboard focus

jalkire:
#### Describe the change
Currently, we have some textboxes with placeholder text. This text disappears when the textbox receives keyboard focus. This PR updates that behavior such that the placeholder text instead disappears only when the textbox is not empty, regardless of keyboard focus. A new `PlaceholderTextBox` class extending `TextBox` is introduced to accomplish this.

Example of new behavior:
placeholder

#### PR checklist

- [x] Includes UI changes?


jalkire:
> It makes sense, although I'm curious whether we can encapsulate the new behavior into a new control to make this change easier across the app if needed. What are your thoughts?

I originally thought a custom control sounded overkill for this, but you inspired me to go for it. Updated to use a `PlaceholderTextBox` extending `TextBox`

PR #519, opened 2019-09-04
Set dummy HelpText value on some buttons

jalkire:
#### Describe the change
According to [the docs](https://docs.microsoft.com/en-us/dotnet/api/system.windows.automation.automationproperties.helptext?view=netframework-4.8), AutomationProperties.HelpText is generally the same text that is provided in the tooltip for the control. WPF takes this to heart and automatically sets a button's helptext to its tooltip value. Frustratingly, if the tooltip is set as a text-containing control such as a `Run`, the helptext is set to be that control's class name. For ai-win's icon buttons, the tooltip's content is roughly the same as the AutomationProperties.Name's content. To avoid repetitive screen reader output, we don't want to set the helptext with the tooltip at all (much less with the tooltip's class name). This PR sets dummy (blank space) values to the helptext in the situations where we want to stop the default tooltip-as-helptext behavior. 

#### Before:
image
#### After:
image

#### PR checklist

- [ ] Includes UI changes?

### DIFF
--- a/src/AccessibilityInsights/MainWindow.xaml
+++ b/src/AccessibilityInsights/MainWindow.xaml
@@ -309,7 +309,7 @@
                                     <behaviors:KeyboardToolTipButtonBehavior/>
                                 </i:Interaction.Behaviors>
                                 <Button.Style>
-                                    <Style TargetType="Button" BasedOn="{StaticResource BtnStandard}">
+                                    <Style TargetType="Button" BasedOn="{StaticResource BtnNoAutoHelpText}">
                                         <Setter Property="FocusVisualStyle" Value="{DynamicResource {x:Static SystemParameters.FocusVisualStyleKey}}"/>
                                         <Style.Triggers>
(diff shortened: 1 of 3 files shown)


src/AccessibilityInsights/MainWindow.xaml
```diff
@@ -363,7 +363,7 @@
                                 <fabric:FabricIconControl GlyphName="Refresh" FontSize="16" Foreground="{DynamicResource ResourceKey=IconBrush}"/>
                             </Button>
                             <Button x:Name="btnSave"
-                                    Style="{StaticResource BtnStandard}"
+                                    Style="{StaticResource BtnNoAutoHelpText}" AutomationProperties.HelpText=""
```
karanbirsingh: can we remove this helptext setter?

jalkire:
```suggestion
                                    Style="{StaticResource BtnNoAutoHelpText}"
```

PR #522, opened 2019-09-05
Move fabric icons into control view

karanbirsingh:
#### Describe the change

This change updates the UIA properties of our fabric icons across the app. After this change the icons should appear in the control view, have the control type "Image", the localized control type "icon", and have no children. Because the control type is now "image", the icons will show up in scan mode. This also means that a screen reader will say "search icon" instead of just "search" - beforehand, the text child of the icon (implementation detail) would show up in scan mode but not the icon itself. This PR improves the representation so that the text child is not shown and instead the user is presented with an icon.

#### PR checklist

- [x] Includes UI changes?

### DIFF
--- a/src/AccessibilityInsights.CommonUxComponents/Controls/CommonAutomationPeerCreator.cs
+++ b/src/AccessibilityInsights.CommonUxComponents/Controls/CommonAutomationPeerCreator.cs
@@ -1,5 +1,5 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
-// Licensed under the MIT license. See LICENSE file in the project root for full license information.using System.Windows.Automation.Peers;
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
 using System.Windows.Automation.Peers;
 using System.Windows.Controls;
 
--- a/src/AccessibilityInsights.CommonUxComponents/Controls/CommonAutomationPeerCreator.cs
+++ b/src/AccessibilityInsights.CommonUxComponents/Controls/CommonAutomationPeerCreator.cs
@@ -18,7 +18,7 @@ namespace AccessibilityInsights.CommonUxComponents.Controls
                 owner,
                 localizedControl: "icon",
                 isContentElement: false,
-                isControlElement: false,
+                isControlElement: true,
                 hideChildren: true,
                 controlType: AutomationControlType.Image);
         }
--- a/src/AccessibilityInsights.CommonUxComponents/Controls/CommonAutomationPeerCreator.cs
+++ b/src/AccessibilityInsights.CommonUxComponents/Controls/CommonAutomationPeerCreator.cs
@@ -1,4 +1,5 @@
-﻿using System.Windows.Automation.Peers;
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.using System.Windows.Automation.Peers;
 using System.Windows.Controls;
 
 namespace AccessibilityInsights.CommonUxComponents.Controls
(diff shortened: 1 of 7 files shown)


src/AccessibilityInsights.CommonUxComponents/Controls/CommonAutomationPeerCreator.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using System.Windows.Automation.Peers;
+using System.Windows.Controls;
+
+namespace AccessibilityInsights.CommonUxComponents.Controls
+{
+    class CommonAutomationPeerCreator
+    {
+        public static AutomationPeer CreateIconAutomationPeer(UserControl owner)
```
dbjorge: I'm not sure it's necessarily the case that *all* icons everywhere should be considered decorative/non-content; maybe this method name should specifically mention `DecorativeIcon` or `PresentationalIcon` or `NonContentIcon`?

src/AccessibilityInsights.CommonUxComponents/Controls/CustomControlOverridingAutomationPeer.cs
```diff
@@ -26,43 +28,53 @@ public class CustomControlOverridingAutomationPeer: UserControlAutomationPeer
         /// </summary>
         private string LocalizedControlType;
 
+        private AutomationControlType ControlType;
+
+        /// <summary>
+        /// Whether to hide all children automation peers
```
dbjorge: This comment adds no information over the name of the property; why include it? (same for the other existing doc comments on the private fields above it)

src/AccessibilityInsights.CommonUxComponents/Controls/CommonAutomationPeerCreator.cs
karanbirsingh: I like `NonContentIcon` or maybe even `ControlViewIcon`

src/AccessibilityInsights.CommonUxComponents/Controls/CustomControlOverridingAutomationPeer.cs
```diff
@@ -26,43 +28,53 @@ public class CustomControlOverridingAutomationPeer: UserControlAutomationPeer
...
+            bool isControlElement = true, 
+            bool isContentElement = false, 
+            bool hideChildren = false, 
+            AutomationControlType controlType = AutomationControlType.Custom)
         : base(owner)
         {
-            this.LocalizedControlType = localizedcontrol;
+            if (isContentElement && !isControlElement)
+            {
+                throw new ArgumentException("The content tree is a subset of the control tree");
```
dbjorge: Should we make this a compile error rather than a runtime error? (eg, by replacinhg the 2 bool parameters with an enum of Content | ControlOnly | Decorative)

jalkire: I like the enum idea. The one (arguable) benefit I see to leaving the two bools is to keep this class as lightweight and aligned with the underlying UIAutomation classes as possible.

karanbirsingh: The enum alternative occurred to John/me but we weren't sure if it'd be better to stick closer to the UIA surface. Now that you've also mentioned it, I'll take a closer look. We also have an existing `TreeViewMode` enum for `Raw`, `Control`, and `Content` (although `ControlOnly` might be more understandable for this one)

dbjorge: That's fair enough. I think it's unlikely you'd miss this particular error in any sort of testing, so I think it's fine as a runtime check

PR #526, opened 2019-09-09
Opt into accessibility features available in .NET framework 4.8

karanbirsingh:
This PR opts into accessibility features as described by these [docs](https://docs.microsoft.com/en-us/dotnet/framework/whats-new/whats-new-in-accessibility#accessibility-switches):

> You can configure your app to opt into accessibility features if it targets .NET Framework 4.7 or an earlier version but is running on .NET Framework 4.7.1 or later.

This results in some of the following improvements if the app is running on a machine with .NET framework 4.8 installed:

- > In .NET Framework 4.7.2 and earlier versions, tooltips display only when the user hovers the mouse cursor over a control. In .NET Framework 4.8, tooltips also display on keyboard focus, as well as via a keyboard shortcut.
- > Windows 10 introduced two new UIAutomation properties, SizeOfSet and PositionInSet, which are used by applications to describe the count of items in a set. UIAutomation client applications such as screen readers can then query an application for these properties and announce an accurate representation of the application’s UI...In addition, items in ItemsControl instances provide a value for these properties automatically without additional action from the developer.

The switch also removes some UI elements from the control view if they are not visible. One of our UI tests depended on this behavior to check some text so I have removed that line for now.

#### PR checklist

- [ ] Includes UI changes?

### DIFF
--- a/src/AccessibilityInsights/App.config
+++ b/src/AccessibilityInsights/App.config
@@ -9,6 +9,7 @@
       <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.7.1" />
   </startup>
   <runtime>
+    <AppContextSwitchOverrides value="Switch.UseLegacyAccessibilityFeatures=false;Switch.UseLegacyAccessibilityFeatures.2=false;Switch.UseLegacyAccessibilityFeatures.3=false;Switch.UseLegacyToolTipDisplay=false" />
     <assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">
       <dependentAssembly>
         <assemblyIdentity name="Newtonsoft.Json" publicKeyToken="30ad4fe6b2a6aeed" culture="neutral" />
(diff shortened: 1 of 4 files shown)


src/AccessibilityInsights/App.config
```diff
@@ -9,6 +9,7 @@
       <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.7.1" />
   </startup>
   <runtime>
+    <AppContextSwitchOverrides value="Switch.UseLegacyAccessibilityFeatures=false;Switch.UseLegacyAccessibilityFeatures.2=false;Switch.UseLegacyAccessibilityFeatures.3=false;Switch.UseLegacyToolTipDisplay=false" />
```
karanbirsingh: You can find this in "Tooltips on keyboard access" at this link: https://docs.microsoft.com/en-us/dotnet/framework/whats-new/whats-new-in-accessibility#wpf48

PR #527, opened 2019-09-09
Add telemetry of installed .NET Framework version

DaveTryon:
#### Describe the change
Add a telemetry property to allow us to better understand what .NET Framework version is installed. This is to enable us to better evaluate how users will be impacted by potential future adoptions of newer versions of the .NET framework.

The lack of unit tests for the new code is intentional. We don't want to add any more Fake-based tests, and there's little return in testing the behavior of .NET's Registry.GetValue method. Testing was ensuring that the return value is correct for my dev machine, as well as confirming that `GetInstalledDotNetFrameworkVersion` returns null if the value doesn't exist in the registry.

#### PR checklist

- [ ] Includes UI changes?

### DIFF
--- a/src/AccessibilityInsights.SharedUx/Telemetry/TelemetryProperty.cs
+++ b/src/AccessibilityInsights.SharedUx/Telemetry/TelemetryProperty.cs
@@ -37,5 +37,6 @@ namespace AccessibilityInsights.SharedUx.Telemetry
         ReleaseChannel,
         ReleaseChannelConsidered,
         IssueReporter,
+        InstalledDotNetFrameworkVersion,
     }
 }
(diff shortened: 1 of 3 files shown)


src/AccessibilityInsights.SharedUx/Telemetry/TelemetryProperty.cs
```diff
@@ -37,5 +37,6 @@ public enum TelemetryProperty
         ReleaseChannel,
         ReleaseChannelConsidered,
         IssueReporter,
+        InstalledDotNetFrameworkVersion,
```
DaveTryon: I'll update the telemetry docs once this is merged

PR #528, opened 2019-09-10
Improve LiveMode UITest

jalkire:
#### Describe the change
The LiveMode UITest currently has a few issues:
1. Selecting WildlifeManager to run tests against is unreliable
2. A recent axe-windows update changed the number of test results we get, causing an assert to fail #523 
3. The recent .NET 4.8 switch update changes the text we get from TreeViewItems #526. I filed [an issue about this](https://github.com/microsoft/WinAppDriver/issues/859) with WinAppDriver, but we should fix the test in the meantime.

This PR addresses 2 and 3 by updating our asserts. It addresses 1 by setting focus to ai-win and then WildlifeManager before attempted to run tests, which seems to improve reliability. It does not fully fix the issue, though.


build/ui-test-job.yml
```diff
@@ -38,7 +38,7 @@ jobs:
       platform: '$(BuildPlatform)'
       configuration: ${{ parameters.configuration }}
       runSettingsFile: $(Build.SourcesDirectory)/src/UITests//UITests.runsettings
-      testFiltercriteria: 'TestCategory=Integration&TestCategory!=LiveMode'
+      testFiltercriteria: 'TestCategory=Integration'
```
jalkire: Re-enable live mode UI test

jalkire: Nevermind

src/AccessibilityInsights.Win32/Properties/AssemblyInfo.cs
```diff
@@ -37,6 +37,7 @@
 [assembly: InternalsVisibleTo("AccessibilityInsights.SharedUx")]
 [assembly: InternalsVisibleTo("AccessibilityInsights.CommonUxComponents")]
 [assembly: InternalsVisibleTo("AccessibilityInsights.Extensions")]
+[assembly: InternalsVisibleTo("UITests")]
```
DaveTryon: My first thought was "Why aren't we doing this in the signed build?" I figured it out after I thought about it, but adding a comment here might clarify it for future readers

src/UITests/LiveMode.cs
```diff
@@ -37,7 +38,7 @@ private void TestTestResults()
 
         private void ValidateResultsInUIATree()
         {
-            driver.TestMode.ResultsInUIATree.ValidateResults(2, 11);
```
DaveTryon: >11 [](start = 64, length = 2)

Is there a way to avoid potentially changing these values any time we pick up a new version of Axe.Windows?

jalkire: Per our offline discussion, I've added a comment explaining when we expect this value to change

src/AccessibilityInsights.Win32/Properties/AssemblyInfo.cs
jalkire: Comment added

PR #531, opened 2019-09-11
fix: enable Narrator to read the element description in the inspect details pane

dbjorge:
#### Describe the change

This change fixes an issue where narrator wasn't including the `tbElement` text (the text description of the element above the properties in the inspect details pane) in reading order. It does so by:

* Collapsing the original nested grid structure of the details tabitem content into a flatter grid
* Wrapping that flatter grid in an unnamed container which appears in the UIAutomation tree (I used an unnamed NamedPane, but if there's a more appropriate control for this someone has a suggestion for, I'd be happy to use it)

Both of these changes were required to get the component to appear in reading order in Narrator. It's not clear to my why.

There is no visual change. I'm struggling to verify the change in NVDA/JAWS since I can't get either to read the text content in question in either production or a local build, will get help from Rob to figure out what I'm missing about their usage :)

side-by-side screenshots of insiders build and build with change, with no apparent differences

#### PR checklist

- [x] Does this address an existing issue? If yes, Issue# 1583849
- [x] Includes UI changes?

### DIFF
--- a/src/AccessibilityInsights.SharedUx/Controls/InspectTabsControl.xaml
+++ b/src/AccessibilityInsights.SharedUx/Controls/InspectTabsControl.xaml
@@ -37,34 +37,31 @@
                             Key="V"
                             Modifiers="Alt"/>
             </TabItem.InputBindings>
-            <Grid>
-                <Grid.RowDefinitions>
-                    <RowDefinition Height="Auto"/>
-                    <RowDefinition Height="*"/>
-                </Grid.RowDefinitions>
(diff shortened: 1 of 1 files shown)


src/AccessibilityInsights.SharedUx/Controls/InspectTabsControl.xaml
```diff
@@ -37,34 +37,31 @@
                             Key="V"
                             Modifiers="Alt"/>
             </TabItem.InputBindings>
-            <Grid>
```
jalkire: Happy to see some superfluous grids going away 😃

PR #532, opened 2019-09-11
Improve n-of-m numbering by restructuring tree view walker menu

karanbirsingh:
#### Describe the change
The SizeOfSet and PositionInSet properties of the tree view menu items were incorrect (2 of 7, 3 of 7, etc) for the following reasons:
- We used a non-actionable menu item as a visual header
- Separators are included in the SizeOfSet count for menu items
- Hidden menu items are included in the SizeOfSet count for menu items

We are unable to override the SizeOfSet or PositionInSet properties unless we explicitly target .NET framework 4.8. In the meantime, this PR restructures the menu to improve accessibility. Now the SizeOfSet and PositionInSet properties are as expected and the visual-only header and separator have gone away.

**before**
"before" screenshot of menu

**after**
"after" screenshot of menu

#### PR checklist

- [x] Includes UI changes?
-- It looks like both in production and this PR, we flag popups created by context menus because they have no name. We may want to look into this in more detail to see if axe-windows needs to be updated

### DIFF
--- a/src/AccessibilityInsights.SharedUx/Controls/HierarchyControl.xaml.cs
+++ b/src/AccessibilityInsights.SharedUx/Controls/HierarchyControl.xaml.cs
@@ -494,7 +494,7 @@ namespace AccessibilityInsights.SharedUx.Controls
                 this.rbRaw.IsChecked = this.ElementContext.Element.TreeWalkerMode == TreeViewMode.Raw;
             }
             var suffix = this.rbRaw.IsChecked.HasValue && this.rbRaw.IsChecked.Value ? "Checked" : "";
-            this.mniRaw.SetValue(AutomationProperties.NameProperty, $"Walk tree in Raw view {suffix}");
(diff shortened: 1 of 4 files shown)


karanbirsingh:
@cheade here is what came out of our conversation. I'm oof tmrw but @jalkire or @dbjorge have some context

src/AccessibilityInsights.SharedUx/Controls/HierarchyControl.xaml.cs
```diff
@@ -131,6 +131,12 @@ public void SetElement(ElementContext ec, bool expandall = false)
                 }
 
                 UpdateButtonVisibility();
+
+                cmHierarchySettings.Items.Remove(mniShowUncertain);
```
karanbirsingh: realized I'm not sure if this will fail upon rescan - e.g. what if mniShowUncertain is not in the list? A contains check first might work better

Issue #545, opened 2019-09-16
[BUG] Release script is not resilient to changes in Octokit version


DaveTryon:
Please check whether the bug has [already been filed](https://github.com/Microsoft/accessibility-insights-windows/issues).

**Describe the bug**
The ValidateReleaseInfo.ps1 script uses a hardcoded version for Octokit, but the NuGet reference always picks up the latest version. As such, the script breaks when a new version is released.

Note: the Canary and Production release pipelines have the same sort of problem 😞 

**Additional context**
Priority requested -

PR #548, opened 2019-09-16
Add Octokit version as an overridable variable

DaveTryon:
#### Describe the change
Update ValidateReleaseInfo.ps1 with an environment variable to control the release of Octokit to use when validating the releases. This variable can be updated by changing the build definition.

This is the code-side fix for #545. The other changes are all in build definitions (already in place and validated)


tools/scripts/ValidateReleaseInfo.ps1
```diff
@@ -74,8 +74,15 @@ function ValidateChannelInfo($channel)
 	return $isCurrentVersionValid -and $isMinimumVersionValid -and $isVersionRelationshipValid -and $isInstallerUrlValid -and $isReleaseNotesUrlValid
 }
 
+# Determine our octokit version
+$octokitVersion = $Env:OctokitVersion
```
DaveTryon: This is set to 0.33.0 in the build definition for the PR build. Having the default value here is mainly for local use

PR #549, opened 2019-09-16
Improve n-of-m numbering by restructuring event config settings menu 

karanbirsingh:
#### Describe the change

As in #532 - the current structure of the context menu for event recording is not accessible:
- the separator is included in the n-of-m count
- the disabled header menu item is not easily accessible via screen readers
- the n-of-m counts are inaccurate because they include the non-focusable header

This PR introduces a submenu so that the n-of-m counts are correct and the header is more accessible.

#### PR checklist

- [x] Includes UI changes?

Screenshot:
image

### DIFF
--- a/src/AccessibilityInsights.SharedUx/Controls/EventRecordControl.xaml
+++ b/src/AccessibilityInsights.SharedUx/Controls/EventRecordControl.xaml
@@ -117,7 +117,7 @@
                     <fabric:FabricIconControl GlyphName="Settings" GlyphSize="Custom" FontSize="12" HorizontalAlignment="Center" VerticalAlignment="Center" Foreground="{DynamicResource ResourceKey=IconBrush}"/>
                     <Button.ContextMenu>
                         <ContextMenu FlowDirection="LeftToRight" Loaded="ContextMenu_Loaded" Unloaded="ContextMenu_Unloaded">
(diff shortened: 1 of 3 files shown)


src/AccessibilityInsights.SharedUx/Controls/EventRecordControl.xaml
```diff
@@ -117,26 +117,26 @@
                     <fabric:FabricIconControl GlyphName="Settings" GlyphSize="Custom" FontSize="12" HorizontalAlignment="Center" VerticalAlignment="Center" Foreground="{DynamicResource ResourceKey=IconBrush}"/>
                     <Button.ContextMenu>
                         <ContextMenu FlowDirection="LeftToRight" Loaded="ContextMenu_Loaded" Unloaded="ContextMenu_Unloaded">
-                            <MenuItem Header="Event recording scope" FontWeight="Bold" IsEnabled="False" IsCheckable="False"/>
-                            <MenuItem x:Name="mniRaw" IsCheckable="False" Click="mniRB_Click"
+                            <MenuItem Header="Event recording scope">
```
jalkire: This should probably be in resources

jalkire: I realize this was already not in resources, so if you want I can let it through

Issue #544, opened 2019-09-16
[BUG] File issue button link is cut off


cheade:
**Describe the bug**
![image](https://user-images.githubusercontent.com/40274682/64978630-d6991380-d86a-11e9-8b7a-08be62459c31.png)

See "File issue" button link is cut off. The issue column is unable to be adjusted to accommodate full width of "file issue".

**To Reproduce**
1. Open Accessibility Insights
2. Run automated checks on an app that will produce Accessibility errors
3. View any error result in the UIA tree view.
4. "Test results" box contains the cut-off error, under the "Issue" column

**Expected behavior**
Issue column should be default wide enough to contain all of "File issue" without getting cut off. 

**Actual behavior**
"File issue" is cut off, and the issue column cannot be adjusted. 

**Desktop (please complete the following information):**
 - OS: Windows 10 version 1809 (OS Build 17763.437)
 - Accessibility Insights for Windows Version: Prerelease Build (1.1.988.1)
 - Target Application: Wildlife manager
 - Target Application Version: N/A

**Additional context**
Priority requested - low

jalkire:
Context: this is happening because the width is a hardcoded value (due to WPF ListView behavior not lining up with what we wanted) set based on our original font size, which is no longer the default font size. It is not cut off when you set your font to small in settings: 
![image](https://user-images.githubusercontent.com/4615491/64983852-5d9fb900-d876-11e9-95d7-8afcb76c5b4a.png)

This easy to fix.

DaveTryon:
@cheade, the fix for this will be in build 1.1.0990.01. This build should be in the Canary channel on 9/18. Please regress it against an appropriate build, then close it out. Thanks!

DaveTryon:
@cheade, Build 1.1.990.1 (which contains this fix) just release to the Production channel. Please close at your convenience

cheade:
Thanks, team!

PR #551, opened 2019-09-16
Issue #544: Don't cut off "File Issue" text in standard font size

DaveTryon:
#### Describe the change
Make the "File Issue" column width responsive to the configured font size.

#### PR checklist

- [ ] Does this address an existing issue? If yes, Issue# - #544 
- [ ] Includes UI changes?

image

### DIFF
--- a/src/AccessibilityInsights.SharedUx/Utilities/HelperMethods.cs
+++ b/src/AccessibilityInsights.SharedUx/Utilities/HelperMethods.cs
@@ -51,6 +51,21 @@ namespace AccessibilityInsights.SharedUx.Utilities
         }
 
         /// <summary>
+        /// Get a "File Issue" column width that is appropriate for the currently configured font size
+        /// </summary>
+        public static double FileIssueColumnWidth
+        {
+            get
+            {
+                switch (ConfigurationManager.GetDefaultInstance().AppConfig.FontSize)
+                {
+                    case Enums.FontSize.Small: return 80.0;
+                    default: return 90.0;
+                }
+            }
+        }
+
+        /// <summary>
         /// Provides bindable property for ProgressRingControls
         /// </summary>
         public static bool PlayScanningSound => ConfigurationManager.GetDefaultInstance().AppConfig.PlayScanningSound;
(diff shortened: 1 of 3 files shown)


src/AccessibilityInsights.SharedUx/Utilities/HelperMethods.cs
```diff
@@ -50,6 +50,21 @@ public static Visibility GeneralFileBugVisibility
...
+        /// Get a "File Issue" column width that is appropriate for the currently configured font size
+        /// </summary>
+        public static double FileIssueColumnWidth
+        {
+            get
+            {
+                switch (ConfigurationManager.GetDefaultInstance().AppConfig.FontSize)
+                {
+                    case Enums.FontSize.Small: return 80.0;
+                    default: return 90.0;
```
DaveTryon: >default [](start = 20, length = 7)

I used a switch instead of an if as defensive programming about future text sizes being added. Happy to switch it if preferred

src/AccessibilityInsights.SharedUx/Utilities/HelperMethods.cs
```diff
@@ -50,6 +50,21 @@ public static Visibility GeneralFileBugVisibility
...
         }
 
+        /// <summary>
+        /// Get a "File Issue" column width that is appropriate for the currently configured font size
+        /// </summary>
+        public static double FileIssueColumnWidth
+        {
+            get
+            {
+                switch (ConfigurationManager.GetDefaultInstance().AppConfig.FontSize)
```
jalkire: Do we want this method to include the 0 width option seen in `MainWindow`'s `BugColumnWidth` property? I think (but am not sure) it is a bug that `ScannerResultControl`'s `BugColumnWidth` is not currently 0 when there is no issue filing. This isn't a regression though, so this PR doesn't need to fix it.

DaveTryon: I was thinking that it made sense to keep the visibility and the size decoupled. Leaving it as is for now, but if/when it's a problem it could make sense to combine them.

PR #552, opened 2019-09-17
Update hotkey dialog to have TextBox + block saving tab key

jalkire:
#### Describe the change
We were seeing a few accessibility/usability issues with the set hotkey dialog. This PR makes two key changes:
1. A TextBox is added to display the new hotkey selection. What is displayed does not change. The key purpose of the TextBox is to provide an actual focusable element on the window, which there previously was not.
2. The Tab key can no longer be added to a shortcut.

Minor visual impact. We could make it look the same, but then the TextBox would not look like a TextBox, which seems bad to me.
Before:
image
After:
image

#### PR checklist

- [x] Includes UI changes?

### DIFF
--- a/src/AccessibilityInsights.SharedUx/Dialogs/HotkeyGrabDialog.xaml
+++ b/src/AccessibilityInsights.SharedUx/Dialogs/HotkeyGrabDialog.xaml
@@ -5,14 +5,13 @@
         xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
         xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
         xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
-        xmlns:local="clr-namespace:AccessibilityInsights.SharedUx.Dialogs"
         xmlns:Properties="clr-namespace:AccessibilityInsights.SharedUx.Properties"
         mc:Ignorable="d"
         ResizeMode="NoResize"
         Title="{x:Static Properties:Resources.HotkeyGrabDialogWindowTitle}" Height="162" Width="319" 
         KeyDown="Window_KeyDown" KeyUp="Window_KeyUp" 
         WindowStartupLocation="CenterOwner"
-        ShowInTaskbar="False" Topmost="True"
+        ShowInTaskbar="False" Topmost="True" FocusManager.FocusedElement="{Binding ElementName=tbHotkey}"
         AutomationProperties.AutomationId="{x:Static Properties:AutomationIDs.SettingsHotkeyGrabDialog}">
     <Window.Resources>
         <ResourceDictionary Source="pack://application:,,,/AccessibilityInsights.SharedUx;component/Resources/Styles.xaml"/>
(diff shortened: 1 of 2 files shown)


src/AccessibilityInsights.SharedUx/Dialogs/HotkeyGrabDialog.xaml
```diff
@@ -5,14 +5,13 @@
...
         xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
-        xmlns:local="clr-namespace:AccessibilityInsights.SharedUx.Dialogs"
         xmlns:Properties="clr-namespace:AccessibilityInsights.SharedUx.Properties"
         mc:Ignorable="d"
         ResizeMode="NoResize"
         Title="{x:Static Properties:Resources.HotkeyGrabDialogWindowTitle}" Height="162" Width="319" 
         KeyDown="Window_KeyDown" KeyUp="Window_KeyUp" 
         WindowStartupLocation="CenterOwner"
-        ShowInTaskbar="False" Topmost="True"
+        ShowInTaskbar="False" Topmost="True" FocusManager.FocusedElement="{Binding ElementName=tbHotkey}"
```
DaveTryon: >True [](start = 39, length = 4)

Possibly out of scope for this PR, but it seems like Topmost should inherit the Topmost flag of the app. Current code allows this window to float if the app is non-Topmost then you switch to another app

jalkire: I want to say we intentionally made this dialog always topmost because you're not really supposed to leave it open (it doesn't even have a close button)

src/AccessibilityInsights.SharedUx/Dialogs/HotkeyGrabDialog.xaml.cs
```diff
@@ -50,6 +40,7 @@ private void Window_KeyDown(object sender, KeyEventArgs e)
             switch (tempKey)
             {
                 case Key.Escape:
+                case Key.Tab:
```
karanbirsingh: Should we exclude shift as well? I think it might be worthwhile - VS IDE seems to exclude it and it might allow users to shift-tab without us capturing

src/AccessibilityInsights.SharedUx/Dialogs/HotkeyGrabDialog.xaml
jalkire: I've been convinced. Change made.

jalkire:
> In the current dialog, Narrator reads the instructions when you press CapsLock+RightArrow. In the new dialog, CapsLock+RightArrow takes you to the "Press Escape to cancel", and you need to go backwards (CapsLock+LeftArrow) to get to the instructions. That seems incorrect.
> 
> Also, should we have any sort of auditory feedback when the text in the textbox changes. Using Narrator, I'm getting no audio when the modifier changes, and if there's any audio from the main key it gets lost when returning to the settings window.

The instructions are set to the HelpText, but I'm changing the reading order as suggested too.

For the auditory feedback, I believe that is the same as the current behavior, except that you should now hear your new hotkey read to you when the hotkey dialog closes and focus returns to the settings button.

src/AccessibilityInsights.SharedUx/Dialogs/HotkeyGrabDialog.xaml
```diff
@@ -28,19 +27,20 @@
             <RowDefinition Height="Auto"/>
         </Grid.RowDefinitions>
         <Label Content="{x:Static Properties:Resources.HotkeyGrabDialogWindowTitle}" HorizontalAlignment="Left" Margin="8,8,0,0" Grid.Row="0" VerticalAlignment="Top" Style="{StaticResource LblH5}" Padding="0"/>
-        <TextBlock HorizontalAlignment="Left" Margin="8,10,0,0" TextWrapping="Wrap" 
+        <TextBox Grid.Row="2" Margin="0,10" Name="tbHotkey"  
+                 Height="24" Width="300" HorizontalContentAlignment="Center"
+                 VerticalContentAlignment="Center" Style="{StaticResource StandardTextBox}"
+                 AutomationProperties.Name="{x:Static Properties:Resources.HotkeyGrabDialogWindowTitle}"
+                 AutomationProperties.HelpText="{x:Static Properties:Resources.RunTextToChangeKeyBoard}"
+                 KeyDown="Window_KeyDown" KeyUp="Window_KeyUp"
```
dbjorge: Should the names of these handlers be updated now that they aren't just Window's handlers?

src/AccessibilityInsights.SharedUx/Dialogs/HotkeyGrabDialog.xaml.cs
```diff
@@ -117,18 +102,23 @@ private string GetCleanKeyString()
...
                 this.Close();
-            } 
+            }
+        }
+
+        private void TbHotkey_PreviewTextInput(object sender, TextCompositionEventArgs e) => e.Handled = true;
+
+        private void TbHotkey_PreviewExecuted(object sender, ExecutedRoutedEventArgs e)
+        {
+            if (e.Command == ApplicationCommands.Paste)
```
dbjorge: Should this also cover other types of ApplicationCommands that would normally modify the element's Text? (looking at the list of them, Cut and Delete come to mind)

src/AccessibilityInsights.SharedUx/Dialogs/HotkeyGrabDialog.xaml.cs
```diff
@@ -12,29 +12,20 @@ namespace AccessibilityInsights.SharedUx.Dialogs
...
         /// <summary>
         /// Button to update
         /// </summary>
         Button ButtonParent;
 
-        /// <summary>
-        /// Constructor
-        /// </summary>
-        /// <param name="tb">Button which triggered dialog</param>
         public HotkeyGrabDialog(Button tb)
```
dbjorge: When I was testing this, I was surprised by the modifier key presses being "sticky" by default; I'd have expected the sequence "ctrl down, ctrl up, shift down, x" to result in the shortcut being "shift + x" but it is actually "ctrl + shift +x".

I realize this behavior is not different from the original, so I think it's appropriate for this PR to avoid changing it, but I wonder if it's just me that was surprised by it/whether we should file an issue to consider changing it to be more in line with other programs (the first two I tried were VS and VS Code, both of which were non-sticky)

PR #555, opened 2019-09-17
Stop hotkeys from acting while settings pane is open

jalkire:
#### Describe the change
Currently, hotkeys attempt to function at all times and states in ai-win. This includes while the user is trying to change their hotkeys. This PR stops hotkeys from acting when the settings pane is open to avoid potentially confusing or difficult scenarios. We currently only have one hotkey which could possibly be useful in settings, the show/hide window hotkey. In my opinion, the purpose of this hotkey relates to actually testing and inspecting apps, not changing one's settings. I feel comfortable disabling it here. If anyone disagrees with that (or the change as a whole), feel free to voice those concerns.

#### PR checklist

- [-] Does this address an existing issue? If yes, Issue# - 
- [-] Includes UI changes?
  - [-] Run the production version of Accessibility Insights for Windows against a version with changes.
  - [-] Attach any screenshots / GIF's that are applicable.

### DIFF
--- a/src/AccessibilityInsights.SharedUx/KeyboardHelpers/HotKey.cs
+++ b/src/AccessibilityInsights.SharedUx/KeyboardHelpers/HotKey.cs
@@ -20,7 +20,7 @@ namespace AccessibilityInsights.SharedUx.KeyboardHelpers
         public Keys Key { get; internal set; }
         public HotkeyModifier Modifier { get; internal set; }
 
-        public Action Action { get; set; }
+        public Action Action { get; private set; }
 
         public uint ErrorCode { get; private set; }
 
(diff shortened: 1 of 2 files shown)


karanbirsingh:
I think this is a good change and wonder if it's feasible to scope it further to when the hotkey dialog is open; that would allow folks to show/hide or handle any future hotkeys. If not, minimizing can be accomplished via Windows hotkeys as well.

jalkire:
> I think this is a good change and wonder if it's feasible to scope it further to when the hotkey dialog is open; that would allow folks to show/hide or handle any future hotkeys. If not, minimizing can be accomplished via Windows hotkeys as well.

It is definitely feasible/possible. My main thought is that our hotkeys aren't really meant for the settings page conceptually--in my opinion, the fact that any work in the settings page is more of a side-effect of their implementation than an intended behavior. It seems like a hotkey should deliver some experience/functionality, and in no current or planned case does that intended experience involve the settings page. The implementation in this PR is meant to leave the door open for a settings-applicable hotkey if we actually wanted to make one in the future. Maybe I'm overthinking this, though...

PR #583, opened 2019-09-30
Fix screen reader behavior in AutomatedChecks ListView

jalkire:
#### Describe the change
A recent change to enable .NET 4.8 accessibility features had an unintended side effect of breaking our Automated Checks `ListView`'s screen reader user behavior. Specifically, once a group item had been expanded at least once, screen readers would no longer speak upon that item receiving keyboard focus. Per @karanbirsingh's investigation, this appeared to be caused by a change to the `GroupItemAutomationPeer` which made its `HasKeyboardFocus` UIA property set based on its child expander's `HasKeyboardFocus` property. Because of our customization, that expander never receives keyboard focus. This PR addresses the problem by overriding the child expander's `AutomationPeer` and setting its `HasKeyboardFocus` value based on its parent GroupItem's `IsFocused` value. This does not impact non-.NET-4.8 users because it the previous `GroupItemAutomationPeer` is not dependent upon the expander's automation peer in this way.

#### PR checklist

- [-] Does this address an existing issue? If yes, Issue# - 
- [-] Includes UI changes?
  - [-] Run the production version of Accessibility Insights for Windows against a version with changes.
  - [-] Attach any screenshots / GIF's that are applicable.

### DIFF
--- a/src/AccessibilityInsights.SharedUx/Controls/CustomControls/CustomExpander.cs
+++ b/src/AccessibilityInsights.SharedUx/Controls/CustomControls/CustomExpander.cs
@@ -23,7 +23,7 @@ namespace AccessibilityInsights.SharedUx.Controls.CustomControls
         /// In .NET 4.8 grouped ListViews, the GroupItem's HasKeyboardFocus value is set based on its child expander's
         /// HasKeyboardFocus value, not its own IsFocused value. In our AutomatedChecks use case, this breaks ATs'
         /// readings of the ListView group headers. By setting the child expander's HasKeyboardFocus value here,
-        /// we get the desired value in its parent GroupItem's value, and ATs fuction properly again.
+        /// we get the desired value in its parent GroupItem's value, and ATs function properly again.
         /// </summary>
         override protected bool HasKeyboardFocusCore() => parentGroupItem?.IsFocused ?? base.HasKeyboardFocusCore();
     }
--- a/src/AccessibilityInsights.SharedUx/Controls/CustomControls/CustomExpander.cs
+++ b/src/AccessibilityInsights.SharedUx/Controls/CustomControls/CustomExpander.cs
@@ -1,0 +1,21 @@
+using System.Windows.Automation.Peers;
+using System.Windows.Controls;
+
+namespace AccessibilityInsights.SharedUx.Controls.CustomControls
+{
+    public class CustomExpander : Expander
+    {
+        protected override AutomationPeer OnCreateAutomationPeer()
+        {
+            return new CustomExpanderAutomationPeer(this);
+        }
+    }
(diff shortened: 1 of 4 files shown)


src/AccessibilityInsights.SharedUx/Controls/CustomControls/CustomExpander.cs
```diff
@@ -0,0 +1,30 @@
...
+    {
+        GroupItem parentGroupItem;
+
+        public CustomExpanderAutomationPeer(Expander owner) : base(owner) => parentGroupItem = owner?.TemplatedParent as GroupItem;
+
+        /// <summary>
+        /// In .NET 4.8 grouped ListViews, the GroupItem's HasKeyboardFocus value is set based on its child expander's
+        /// HasKeyboardFocus value, not its own IsFocused value. In our AutomatedChecks use case, this breaks ATs'
+        /// readings of the ListView group headers. By setting the child expander's HasKeyboardFocus value here,
+        /// we get the desired value in its parent GroupItem's value, and ATs fuction properly again.
```
karanbirsingh:
```suggestion
        /// we get the desired value in its parent GroupItem's value, and ATs function properly again.
```

karanbirsingh:
we found the source here that inspired this change: https://github.com/dotnet/wpf/blob/8b20a88a69d268168569deb6284193f8e5c56c6a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Automation/Peers/GroupItemAutomationPeer.cs#L352

src/AccessibilityInsights.SharedUx/Controls/CustomControls/CustomExpander.cs
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    public class CustomExpanderAutomationPeer : ExpanderAutomationPeer
+    {
+        GroupItem parentGroupItem;
+
+        public CustomExpanderAutomationPeer(Expander owner) : base(owner) => parentGroupItem = owner?.TemplatedParent as GroupItem;
+
+        /// <summary>
+        /// In .NET 4.8 grouped ListViews, the GroupItem's HasKeyboardFocus value is set based on its child expander's
```
DaveTryon: Thanks for the explanation. What's our risk if the .NET team fix their behavior in the future?

jalkire: I see a low risk that this change negatively impacts future behavior. There is more of a risk that something is changed which renders this obsolete/ineffectual--I have no way of judging the likelihood of that scenario. We'd just want to pay special attention to this ListView in future .NET updates, which is already the case.

PR #617, opened 2019-10-22
Limit scan result telemetry to just once per scan, and never on opening a file

DaveTryon:
#### Describe the change
We currently call PublishScanResults every time a user enters test or hierarchy view, even if no new scan has occurred, and even if we're opening a file from disk. This is artificially inflating the scan information showing up in telemetry, and is causing "Indecisive" to display as a RuleId in the telemetry stream whenever an older A11yTest file is opened.

This change explicitly prevents calls to PublishScanResults when opening a file from disk, and only allows the first attempt to PublishScanResults to succeed.

#### PR checklist

- [ ] Includes UI changes?

### DIFF
--- a/src/AccessibilityInsights/Modes/SnapshotModeControl.xaml.cs
+++ b/src/AccessibilityInsights/Modes/SnapshotModeControl.xaml.cs
@@ -133,14 +133,18 @@ namespace AccessibilityInsights.Modes
             {
                 this.ctrlHierarchy.IsEnabled = false;
                 ElementContext ec = null;
+                bool publishScanResults = MainWin.PublishScanResultsOneShot;
                 await Task.Run(() =>
                 {
                     CaptureAction.SetTestModeDataContext(ecId, this.DataContextMode, Configuration.TreeViewMode);
                     ec = GetDataAction.GetElementContext(ecId);
 
-                    // send telemetry of scan results. 
-                    var dc = GetDataAction.GetElementDataContext(ecId);
-                    dc.PublishScanResults();
+                    if (publishScanResults)
+                    {
+                        // send telemetry of scan results. 
+                        var dc = GetDataAction.GetElementDataContext(ecId);
+                        dc.PublishScanResults();
+                    }
                 }).ConfigureAwait(false);
 
                 Application.Current.Dispatcher.Invoke(() =>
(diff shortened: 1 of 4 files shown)


src/AccessibilityInsights/Modes/TestModeControl.xaml.cs
```diff
@@ -161,7 +162,10 @@ public async Task SetElement(Guid ecId)
                                 dc.ElementCounter.UpperBound);
                             IsSaveEnabled = false;
                         }
-                        dc.PublishScanResults();
+                        if (publishScanResults)
```
jalkire: `CaptureAction.SetTestModeDataContext` returns a bool that tells us whether the DataContext was updated (in the case, whether or not a scan was run). We also already have `DataContextMode`, which tells us if a file was loaded. We could leverage these two values and avoid manually tracking `PublishScanResultsOneShot`

src/AccessibilityInsights/Modes/SnapshotModeControl.xaml.cs
```diff
@@ -133,14 +133,18 @@ public async Task SetElement(Guid ecId)
...
+                bool publishScanResults = MainWin.PublishScanResultsOneShot;
                 await Task.Run(() =>
                 {
                     CaptureAction.SetTestModeDataContext(ecId, this.DataContextMode, Configuration.TreeViewMode);
                     ec = GetDataAction.GetElementContext(ecId);
 
-                    // send telemetry of scan results. 
-                    var dc = GetDataAction.GetElementDataContext(ecId);
-                    dc.PublishScanResults();
+                    if (publishScanResults)
```
jalkire: Same suggesting as in `TestModeControl`

src/AccessibilityInsights/Modes/TestModeControl.xaml.cs
DaveTryon: Thanks--that worked!

src/AccessibilityInsights/Modes/SnapshotModeControl.xaml.cs
DaveTryon: Thanks--that worked!

PR #620, opened 2019-10-29
Fix settings.targets to include keys on SignedRelease

jalkire:
#### Describe the change
In #536- Fix signed build, I claimed to fix our signed build. That may have been true for digital signing, but the PR happened to break strong name signing. The cause: [the conditional here](https://github.com/microsoft/accessibility-insights-windows/blob/3d46005fc1c46861b0975bd5a4594592275a9299/build/settings.targets#L10) wasn't updated to check for the new SignedRelease configuration. A signed build ran without errors related to strong name keys. 

#### PR checklist

- [ ] Includes UI changes?

### DIFF
--- a/build/settings.targets
+++ b/build/settings.targets
@@ -7,7 +7,7 @@
   <Import Project="delaysign.targets" />
 
   <Choose>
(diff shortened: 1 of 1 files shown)


jalkire:
> Approving contingent on signed build succeeding without errors

Build completed without errors.

PR #622, opened 2019-10-30
Gate signing only with build variables

jalkire:
#### Describe the change
PRs #536, #620, and #621 failed to get us digitally and strong name signed. Hopefully this PR will. Here, we stop gating signing based on the Configuration and leave it entirely up to the build variables. We also tweak the variables set in signedbuild.yml to make signing opt-in instead of opt-out. A signed build ran and produced a digitally and strong name signed artifact.

### DIFF
--- a/src/AccessibilityInsights.sln
+++ b/src/AccessibilityInsights.sln
@@ -471,9 +471,7 @@ Global
 		{7A7C04EA-8279-4D69-959D-E804DD81BC70}.Release|x86.Build.0 = Release|Any CPU
 		{7A7C04EA-8279-4D69-959D-E804DD81BC70}.SignedRelease|Any CPU.ActiveCfg = Release|Any CPU
 		{7A7C04EA-8279-4D69-959D-E804DD81BC70}.SignedRelease|x64.ActiveCfg = Release|Any CPU
-		{7A7C04EA-8279-4D69-959D-E804DD81BC70}.SignedRelease|x64.Build.0 = Release|Any CPU
 		{7A7C04EA-8279-4D69-959D-E804DD81BC70}.SignedRelease|x86.ActiveCfg = Release|Any CPU
-		{7A7C04EA-8279-4D69-959D-E804DD81BC70}.SignedRelease|x86.Build.0 = Release|Any CPU
 	EndGlobalSection
 	GlobalSection(SolutionProperties) = preSolution
 		HideSolutionNode = FALSE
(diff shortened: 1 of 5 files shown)


src/AccessibilityInsights.sln
```diff
@@ -471,9 +471,7 @@ Global
 		{7A7C04EA-8279-4D69-959D-E804DD81BC70}.Release|x86.Build.0 = Release|Any CPU
 		{7A7C04EA-8279-4D69-959D-E804DD81BC70}.SignedRelease|Any CPU.ActiveCfg = Release|Any CPU
 		{7A7C04EA-8279-4D69-959D-E804DD81BC70}.SignedRelease|x64.ActiveCfg = Release|Any CPU
-		{7A7C04EA-8279-4D69-959D-E804DD81BC70}.SignedRelease|x64.Build.0 = Release|Any CPU
```
jalkire: Lines 474 and 476 should have been removed when SignedRelease was initially introduced.

src/AccessibilityInsights.sln
```diff
@@ -471,9 +471,7 @@ Global
 		{7A7C04EA-8279-4D69-959D-E804DD81BC70}.Release|x86.Build.0 = Release|Any CPU
 		{7A7C04EA-8279-4D69-959D-E804DD81BC70}.SignedRelease|Any CPU.ActiveCfg = Release|Any CPU
 		{7A7C04EA-8279-4D69-959D-E804DD81BC70}.SignedRelease|x64.ActiveCfg = Release|Any CPU
```
DaveTryon: Should these lines also go away?

src/AccessibilityInsights.sln
```diff
@@ -471,9 +471,7 @@ Global
...
-		{7A7C04EA-8279-4D69-959D-E804DD81BC70}.SignedRelease|x64.Build.0 = Release|Any CPU
 		{7A7C04EA-8279-4D69-959D-E804DD81BC70}.SignedRelease|x86.ActiveCfg = Release|Any CPU
```
DaveTryon: Should this line also go away?

src/AccessibilityInsights.sln
jalkire: Discussed offline--VS auto generates this and will restore these lines as well as the undesired silently when the SignedRelease configuration is selected

src/AccessibilityInsights.sln
jalkire: Same as other comment

PR #630, opened 2019-11-07
Add private versions of Axe.Windows methods that aren't used by Axe.Windows

DaveTryon:
#### Describe the change
We identified some Axe.Windows methods that are used only in AIWin. We're adding these methods to AIWin and removing them from Axe.Windows. This is the complement of https://github.com/microsoft/axe-windows/pull/214

#### PR checklist

- [ ] Includes UI changes?

### DIFF
--- a/src/AccessibilityInsights.SharedUx/Utilities/ExtensionMethods.cs
+++ b/src/AccessibilityInsights.SharedUx/Utilities/ExtensionMethods.cs
@@ -267,7 +267,7 @@ namespace AccessibilityInsights.SharedUx.Utilities
             }
         }
 
-        public static System.Windows.Media.Color ToMediaColor(this System.Drawing.Color color)
+        internal static System.Windows.Media.Color ToMediaColor(this System.Drawing.Color color)
         {
             return System.Windows.Media.Color.FromArgb(color.A, color.R, color.G, color.B);
         }
@@ -277,7 +277,7 @@ namespace AccessibilityInsights.SharedUx.Utilities
         /// </summary>
         /// <param name="bitmap"></param>
         /// <returns></returns>
-        public static BitmapSource ConvertToSource(this Bitmap bitmap)
+        internal static BitmapSource ConvertToSource(this Bitmap bitmap)
         {
             if (bitmap == null) throw new ArgumentNullException(nameof(bitmap));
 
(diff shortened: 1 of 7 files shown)


src/AccessibilityInsights.Win32/NativeMethods.cs
```diff
@@ -11,6 +12,17 @@ namespace AccessibilityInsights.Win32
...
+        [DllImport("gdi32.dll", CharSet = CharSet.Auto, SetLastError = true, ExactSpelling = true)]
+        internal static extern uint GetDeviceCaps(IntPtr hDC, int nIndex);
+
+        //https://msdn.microsoft.com/en-us/library/windows/desktop/dd145062(v=vs.85).aspx
+        [DllImport("User32.dll")]
+        internal static extern IntPtr MonitorFromPoint([In]Point pt, [In]uint dwFlags);
+
+        //https://msdn.microsoft.com/en-us/library/windows/desktop/dn280510(v=vs.85).aspx
+        [DllImport("Shcore.dll")]
+        internal static extern IntPtr GetDpiForMonitor([In]IntPtr hmonitor, [In]DpiType dpiType, [Out]out uint dpiX, [Out]out uint dpiY);
```
dbjorge: The underlying `GetDpiForMonitor` returns an HRESULT, not a pointer

src/AccessibilityInsights.SharedUx/Utilities/ExtensionMethods.cs
```diff
@@ -181,9 +185,9 @@ public static List<RuleResultViewModel> GetRuleResultsViewModelList(this Element
         /// </summary>
         /// <param name="window"></param>
         /// <returns></returns>
-        public static Point GetTopLeftPoint(this Window window)
+        public static System.Windows.Point GetTopLeftPoint(this Window window)
         {
-            Point topLeft = new Point();
+            System.Windows.Point topLeft = new System.Windows.Point();
```
RobGallo: Could we add some using or alias statements so that this code doesn't need to include the namespaces? I find the namespaces pretty distracting.

DaveTryon: The problem is that we now have Point in 2 namespaces, so we need to be explicit about which one we're referencing. We could use an alias, but that might be even more distracting

src/AccessibilityInsights.Win32/NativeMethods.cs
DaveTryon: Good catch. Based on http://www.pinvoke.net/default.aspx/Enums/HRESULT.html, I'll use uint instead of IntPtr

Issue #546, opened 2019-09-16
Issue filing UX/UI styling and behavior[BUG] 


cheade:
**Describe the bug**
Button UX behavior and styling in the issue filing for Azure Boards section. 
Text box behavior in GitHub Issue filing has primary text color instead of placeholder text (secondary-text), making it look like text is already filled in. 

**To Reproduce**
1. Run automated checks on an app with errors
2. Click file issue (make sure you don't have a service connected, this should be a first-time setup)
3. Button to connect Azure Boards account is blue before any text has been entered.
4. Text box in GitHub has stand-in text that is primary colored, making it look like there is already text in the box. 

**Expected behavior**
Azure Boards Issue filing setup: 
Button should be disabled styling until a user enters in text to submit. 
GitHub textbox: placeholder text should be in secondary-text color (see placeholder text in the Azure Boards section)

**Screenshots or .GIF**
Azure Boards: 
![image](https://user-images.githubusercontent.com/40274682/64979923-a010c800-d86d-11e9-9bcd-a74d9f393460.png)
GitHub: 
![image](https://user-images.githubusercontent.com/40274682/64979944-ab63f380-d86d-11e9-877f-4a3b455b993c.png)

**Desktop (please complete the following information):**
- OS: Windows 10 version 1809 (OS Build 17763.437)
- Accessibility Insights for Windows Version: Prerelease Build (1.1.988.1)
- Target Application: Wildlife manager
- Target Application Version: N/A

**Additional context**
Priority requested - low
[Link to Figma file](https://www.figma.com/file/MxXdGGjWOjE7KaL3rfJNed5C/Accessibility-Insights-for-Desktop?node-id=442%3A3)

ferBonnin:
Marking as help wanted, this should be a straightforward fix.

DaveTryon:
@cheade, please regress this against a current canary build and close if it's fixed to your satisfaction. Thanks!

cheade:
Thanks, @DaveTryon ! It looks like the GItHub Text color has been resolved but the button for Connecting the Azure DevOps account is still in its primary color. It should be disabled or not present until a user inputs the data into the text field.

ferBonnin:
assigning Dave and marking as Active since per Chelsea, the button is still not using expected behaviour

risanaka:
[...]

PR #641, opened 2019-11-21
Update metadata for releasing 1.1.1048.01 to insider

karanbirsingh:
Creating a new insider release

https://github.com/microsoft/accessibility-insights-windows/blob/be6568c0909b7ef66a3ff45fd0efddda76fc9375/Channels/Insider/release_notes.md

### DIFF
--- a/Channels/Insider/release_info.json
+++ b/Channels/Insider/release_info.json
@@ -1,8 +1,8 @@
 {
   "default": {
-    "current_version": "1.1.0990.01",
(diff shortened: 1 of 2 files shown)


DaveTryon:
Changes look good to me, but @cheade hasn't yet closed #546, which is called out as a bug fix in this release. Seems like we ought to close all fixed issues before we release. I've sent an email to try to nudge this along

Issue #299, opened 2019-04-15
[Feature Request] What's new content hasn't been updated since initial release


DaveTryon:
**Describe the bug**
The content of the What's new dialog needs to be updated to reflect features added since the initial release

**To Reproduce**
1. Go to the What's new dialog
2. Look at the content

**Expected behavior**
The content should have reference to GitHub issue filing and release channels

**Actual behavior**
The content is unchanged from the initial release

**Screenshots or .GIF**
![image](https://user-images.githubusercontent.com/45672944/56150059-cfd83e80-5f62-11e9-906a-d533183e7d6f.png)

**Desktop (please complete the following information):**
 - OS: Windows 10 1903
 - Accessibility Insights for Windows Version: 1.1.0834.01

neocln:
@adrianaandreeva we need a way to display some contents of the "What's new" page. 
could you assign a PM on this and define a Ux on it? once that's done, we need a feature to implement it.

Sirisha-Kommushetty:
@cheade - We have some thoughts on how this page can be revamped. Would be great to chat next week and close on the next steps.

Sirisha-Kommushetty:
Had a discussion with @cheade and @DaveTryon on the next steps here.

Sirisha-Kommushetty:
@cheade - Any updates on this?

cheade:
Last we all talked we had decided that we will be getting rid of the "What's new" tab. Here is the [Figma mock](https://www.figma.com/file/MxXdGGjWOjE7KaL3rfJNed5C/Accessibility-Insights-for-Desktop?node-id=2325%3A23545), but I don't believe we have any major changes.

DaveTryon:
Welcome screen is finally updated in the current canary build (see https://github.com/microsoft/accessibility-insights-windows/releases/tag/v1.1.1230.01). Closing this out.

PR #676, opened 2020-01-01
Remove what's new section + rearrange welcome page

jalkire:
#### Describe the change
We haven't updated the What's New section for a long time. Per issue #299, this PR finishes removing the What's New section of the Welcome screen and rearranges the screen per design. There are still a few more visual tweaks to do before the issue is complete.

image

#### PR checklist

- [x] Does this address an existing issue? If yes, Issue# - 
- [x] Includes UI changes?


src/AccessibilityInsights.SharedUx/Dialogs/StartUpModeControl.xaml.cs
```diff
@@ -61,7 +61,7 @@ public StartUpModeControl()
             Uri releaseNotesUri = MsiUtilities.GetReleaseNotesUri(new SetupExceptionReporter());
             if (releaseNotesUri != null)
             {
-                hlLink.NavigateUri = releaseNotesUri;
+                //hlLink.NavigateUri = releaseNotesUri;
```
smoralesd: should the `if` be remove now that the `then` block is empty? or maybe the commented line was not intentional?

jalkire: Good catch! Removed this now unused code.

src/AccessibilityInsights.SharedUx/Dialogs/StartUpModeControl.xaml.cs
```diff
@@ -86,7 +78,7 @@ public void UpdateHotkeyLabels()
         /// <param name="e"></param>
         private void btnExit_Click(object sender, RoutedEventArgs e)
         {
```
RobGallo: The previous version is correct, no? If the "don't show" checkbox is checked, ShowWelcomeScreenOnLaunch is set to false.

jalkire: The checkbox's text has changed so that it is now a "do show" checkbox, so this needed to be changed as well

jalkire: Updated name

PR #687, opened 2020-01-10
Port IsVisibleLocation from Axe.Windows to AIWin

DaveTryon:
#### Describe the change
The IsVisibleLocation extension method in Axe.Windows is used only by AIWin, so we want to migrate it into AIWin. This change adds it to AIWin. A separate change (as part of the port) will remove it from Axe.Windows.

#### PR checklist

- [ ] Includes UI changes?

### DIFF
--- a/src/AccessibilityInsights.SharedUx/Utilities/ExtensionMethods.cs
+++ b/src/AccessibilityInsights.SharedUx/Utilities/ExtensionMethods.cs
@@ -289,5 +289,19 @@ namespace AccessibilityInsights.SharedUx.Utilities
             NativeMethods.DeleteObject(hbmp);
             return result;
         }
+
+        /// <summary>
+        /// Check whether p is fully visible based on current screen size
+        /// </summary>
+        /// <param name="p"></param>
+        /// <param name="c"></param>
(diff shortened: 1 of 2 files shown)


src/AccessibilityInsights.SharedUx/Utilities/ExtensionMethods.cs
```diff
@@ -289,5 +289,19 @@ internal static BitmapSource ConvertToSource(this Bitmap bitmap)
...
         }
+
+        /// <summary>
+        /// Check whether p is fully visible based on current screen size
+        /// </summary>
+        /// <param name="p"></param>
+        /// <param name="c"></param>
+        /// <param name="margin"></param>
+        /// <returns></returns>
+        internal static bool IsVisibleLocation(this Rectangle p)
```
DaveTryon: This method was copied from lines 110-122 of https://github.com/microsoft/axe-windows/blob/master/src/Desktop/Utility/ExtensionMethods.cs. I intentionally didn't add a "using System.Windows.Forms" to the file, since it introduces class ambiguity that would then need to be addressed.

PR #698, opened 2020-02-05
Eat ArgumentNullException where we already eat NullReferenceException

DaveTryon:
#### Describe the change
There are a few places in the code that explicitly catch `NullReferenceException`, presumably based on observation of previous behavior that we can't directly control. Null parameter checks were recently added to several Axe.Windows methods, meaning that they now throw `ArgumentNullException` instead of `NullReferenceException`. The underlying fact that the data is null hasn't changed, but the changed Exception type is leaving the `ArgumentNullException` uncaught, and the app crashes as a result.

This change identifies all places where we silently ignore `NullReferenceException` and treats `ArgumentNullException` in exactly the same way. It's not pretty, but it's intended to prevent the crash in the same way that we used to prevent it. We've begun conversations on options of notifiying the user that something unexpected occurred, but there's no clear time frame on when we'll have a design, much less implementation.

Note: This doesn't change `ClearOverlayDriver.BringMainWindowsOfPOIElementToFront`, since that code already handles `ArgumentException` (a base class for `ArgumentNullException`).

#### PR checklist

- [ ] Includes UI changes?

### DIFF
--- a/src/AccessibilityInsights.SharedUx/ActionViews/ReturnA11yElementsView.xaml.cs
+++ b/src/AccessibilityInsights.SharedUx/ActionViews/ReturnA11yElementsView.xaml.cs
@@ -361,9 +361,13 @@ namespace AccessibilityInsights.SharedUx.ActionViews
             {
                 MarkElement(((A11yElement)this.cbElements.SelectedItem));
             }
-            catch(NullReferenceException ex)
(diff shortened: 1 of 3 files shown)


DaveTryon:
> Looks good apart from the `e` variable name related build failure!

duh. fixed

PR #710, opened 2020-02-20
Add checks for existence of .NET Core runtime 3.0 or higher

DaveTryon:
#### Describe the change
The latest version of Axe.Windows requires the .NET Core 3.0 runtime, which isn't pre-installed on older versions of Windows. This change adds a check for the presence and version of the .NET core runtime and blocks installation if an appropriate version isn't found. It uses the same check that we use for the Axe.Windows CLI (see https://github.com/microsoft/axe-windows/pull/308 for reference). The text of the blocking message reads "Accessibility Insights for Windows v1.1 requires .NET Core Runtime 3.0 or newer. Please visit https://dotnet.microsoft.com/download/dotnet-core". Once the user clicks the OK button, they're taken to the existing final dialog that explains that installation was unsuccessful.

We should make a point of calling out this dependency in the release notes when this goes to production, since someone using AIWin on an older Windows installation will probably encounter failures during upgrade.

#### PR checklist

- [] Run through of all [test scenarios](https://github.com/Microsoft/accessibility-insights-windows/blob/master/docs/Scenarios.md) completed?
- [] Does this address an existing issue? If yes, Issue# - 
- [x] Includes UI changes?
  - [] Run the production version of Accessibility Insights for Windows against a version with changes.

image

### DIFF
--- a/src/MSI/Product.wxs
+++ b/src/MSI/Product.wxs
@@ -25,6 +25,18 @@ Licensed under the MIT license. See LICENSE file in the project root for full li
     <WixVariable Id="WixUIBannerBmp" Value="Resources\WixDialogBanner.png" />
     <WixVariable Id="WixUILicenseRtf" Value="..\AccessibilityInsights\bin\Release\eula.rtf" />
 
+    <Property Id="NETCORERUNTIMEFOUNDX86">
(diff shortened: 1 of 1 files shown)


DaveTryon:
Upon further testing, this change isn't necessary (tested on Win7)

PR #719, opened 2020-03-09
Dark mode (initial step to get early feedback)

DaveTryon:
#### Describe the change
This is the first step in enabling dark mode in AIWin. It's by no means complete, but hopefully far enough along to start to elicit some feedback. Settings dialogs haven't been updated, so there are some known contrast issues there.

Dark mode is gated behind a flag that needs to be added to the config file. To enable dark mode, add the following to %LocalAppData%\AccessibilityInsights\V1\Configurations\Configuration.json (add it somewhere in the middle to you don't invalidate the json)-- **_Make a backup copy of configuration.json if you're not sure about how to edit a json file_** :

```
  "EnableDarkMode": true,
```

you'll also need to enable dark mode in Windows. The next time you start AIWin, it will come up in dark mode.

#### PR checklist

- [x] Includes UI changes?
  - [ ] Attach any screenshots / GIF's that are applicable.  Too many to attach 😞 

### DIFF
--- a/src/AccessibilityInsights.CommonUxComponents/Controls/DualFabricIconControl.xaml
+++ b/src/AccessibilityInsights.CommonUxComponents/Controls/DualFabricIconControl.xaml
@@ -20,7 +20,7 @@
                     <Setter Property="Foreground" Value="{DynamicResource ResourceKey=WarningBrush}"/>
                 </Trigger>
                 <Trigger Property="GlyphName" Value="DOM">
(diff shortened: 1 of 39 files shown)


jalkire:
This looks like a great start to me! I have one concern with pushing it in master though. Since we lock this to the system theme and don't provide a separate toggle, this means Canary users who use the system dark theme (like myself) will encounter a few minor but difficult contrast issues that require a system theme change to fix. Are we ok with this experience on Canary for the time being?

PR #726, opened 2020-03-14
Update dialog: Better support in dark mode

DaveTryon:
#### Describe the change
The update dialog doesn't work well in dark mode

#### PR checklist

- [x] Includes UI changes?

Update dialog showing a mandatory update in light mode (current is on top, revised is on the bottom--note how the current version makes the disabled "Later" button disappear completely, even though it's still in the UIA tree. I checked with @jalkire, and the "disappearing text" seems wrong to him. I checked with @ferBonnin about color contrast requirements for disabled text, and she confirmed that disabled buttons have no contrast requirement, so the current changes will be OK. She mentioned that we might want to simply remove the permanently disabled button as a usability improvement. I'll bring this up with @adarshrema, and if he approves the removal, that will be a separate PR. 

image

Update dialog showing a mandatory update in dark mode (current is on the top, revised is on the bottom):

image


jalkire:
Have you verified these new colors/brushes in HC modes?

DaveTryon:
> Have you verified these new colors/brushes in HC modes?

HC modes provide no visual distinction until mouseover, which is less than ideal. We're going to collapse the permanently disabled buttons, so it will be a non-issue in this case. There are other cases where we'll need to find a better story for disabled buttons in HC (but not in this change)

PR #728, opened 2020-03-17
Use .NET 4.7.2 instead of .NET 4.7.1

DaveTryon:
#### Describe the change
As a side effect of the conversion of Axe.Windows to .NET Core, the compiler is now ignoring version 4.3.4 of System.Net.Http.dll (specified via NuGet). As a result, we're now shipping version 4.3.3 of System.Net.Http.dll instead of the security review suggested version 4.3.4. The recommended way to get to the newer version is to use .NET 4.7.2 instead of 4.7.1, but that's a significant change. This PR updated to .NET 4.7.2, and we'll need to evaluate thoroughly before we merge.

#### PR checklist

- [ ] Includes UI changes?


DaveTryon:
Closing this PR out and will create a new PR based on the branch in the main repo (instead of the one in my fork)

PR #730, opened 2020-03-18
Improve border area at zoom

DaveTryon:
#### Describe the change
The caption and command bars get overwritten as the window shrinks and/or the text size grows. This PR adjusts the positioning and backgrounds of the controls. It does _not_ address the problem of the items on the right of the command bar becoming unavailable as the window becomes smaller.

#### PR checklist

- [x] Includes UI changes?

Before the change:
Before

After the change:
After

### DIFF
--- a/src/AccessibilityInsights/MainWindow.xaml
+++ b/src/AccessibilityInsights/MainWindow.xaml
@@ -289,11 +289,11 @@
                             <RowDefinition Height="Auto"/>
                             <RowDefinition Height="Auto"/>
                         </Grid.RowDefinitions>
-                        <DockPanel Height="Auto">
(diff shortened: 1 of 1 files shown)


src/AccessibilityInsights/MainWindow.xaml
```diff
@@ -234,7 +234,7 @@
                             </ItemsControl.ItemsPanel>
                             <Button Width="35" Height="28"
                                     Style="{StaticResource BtnStandard}"
-                                    x:Name="btnHelp" Focusable="False" IsTabStop="False"
```
DaveTryon: This change is a lot easier if you hide whitespace changes (accessible via the gear icon):

image

PR #266, opened 2019-04-08
Release channel config ux

jalkire:
#### Describe the change
Adds ux to application settings to pick a release channel. Saving a new channel selection will trigger a reinstall of AI-Win to that channel. Includes changes to version switcher logic, compliments of @DaveTryon.

Note that there will be a confirmation dialog to avoid unwanted app reinstalls added in a future PR.

#### PR checklist

- [ ] Run through of all [test scenarios](https://github.com/Microsoft/accessibility-insights-windows/blob/master/docs/Scenarios.md) completed?
- [ ] Does this address an existing issue? If yes, Issue# - 
- [x] Includes UI changes?
  - [x] Run the production version of Accessibility Insights for Windows against a version with changes.
  - [x] Attach any screenshots / GIF's that are applicable.

image

image

> Note: After the PR has been created, certain checks will be kicked off. All of these checks must pass before a merge.

PR #732, opened 2020-03-19
Ensure copied VersionSwitcher files are identical to their source files

DaveTryon:
#### Describe the change
Our existing upgrade code copies the VersionSwitcher app from ProgramFiles to Temp, then executes from Temp. We lock each file against writing immediately after copying it, but there's a tiny time window where a malicious app could theoretically modify the copied file before it is locked. This change waits until after the file is protected against writing, then compares its contents to the source file. There's a potential for a perf hit here, but since both source and destination files are probably still in the disk cache, it should be minimal.

This PR includes one other tiny change. #266 bypassed the method that validates that the source folder exists. This change seemed unintentional since it created an unused method. I noticed this unused method only because of the color coloring when I had this file open in the IDE. I've restored the code path that checks for the existence of the source folder.

Tested by building the MSI with my clock set back a couple of weeks (this creates a version stamp older than what is in Canary), resetting my clock, uninstalling my Canary AIWin installation, then installing the MSI that I built. When AIWin started, it took me through the upgrade process, which successfully upgraded to the official Canary build.

#### PR checklist

- [ ] Includes UI changes?

### DIFF
--- a/src/AccessibilityInsights.SetupLibrary/VersionSwitcherWrapper.cs
+++ b/src/AccessibilityInsights.SetupLibrary/VersionSwitcherWrapper.cs
@@ -4,6 +4,7 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
+using System.Linq;
 
 namespace AccessibilityInsights.SetupLibrary
 {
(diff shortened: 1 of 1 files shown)


src/AccessibilityInsights.SetupLibrary/VersionSwitcherWrapper.cs
```diff
@@ -118,6 +118,30 @@ private static void RecursiveTreeCopy(string source, string dest, List<FileStrea
...
+                if (file1Bytes[index] != file2Bytes[index])
+                {
+                    matches = false;
+                }
+            }
+
+            if (!matches)
+            {
+#pragma warning disable CA1303 // Do not pass literals as localized parameters
+                throw new IOException("File " + file1 + " does not match " + file2);
```
DaveTryon: Exception type here is somewhat arbitrary--please let me know if you think a different class would be more appropriate

src/AccessibilityInsights.SetupLibrary/VersionSwitcherWrapper.cs
```diff
@@ -118,6 +118,30 @@ private static void RecursiveTreeCopy(string source, string dest, List<FileStrea
...
+                EnsureFilesAreIdentical(file, destFile);
+            }
+        }
+
+        private static void EnsureFilesAreIdentical(string file1, string file2)
+        {
+            byte[] file1Bytes = File.ReadAllBytes(file1);
+            byte[] file2Bytes = File.ReadAllBytes(file2);
+
+            bool matches = file1Bytes.Length == file2Bytes.Length;
```
jalkire: You could use
```suggestion
            bool matches = file1Bytes.SequenceEqual(file2Bytes);
```
here and then skip the manual iteration/equality checking below. Presumably doesn't make much of a difference

DaveTryon: Always happy to have less code 😄

PR #746, opened 2020-03-31
Use direct project references to extension projects

DaveTryon:
#### Describe the change
When extensions were first added, we opted for a model that would allow third parties to "superimpose" their extensions into an installation of AIWin. It seemed to make sense at the time, but it meant that there were no compiler-enforced links between the EXE and the NuGet packages that are consumed by the extensions. As a result, we ran the risk of version mismatches in the assembly packages.

After internal discussion, we've decided to let the compiler manage the assembly versions for us. We still use MEF to find and load extensions, but we can simplify both the code and the file hierarchy. As an added bonus, ExtensionsTests no longer needs to use fakes.

Tested by building the local MSI and ensuring that upgrades, ADO, GitHub issues, and telemetry all work.

#### PR checklist

- [ ] Includes UI changes?


src/AccessibilityInsights.ExtensionsTests/ContainerUnitTests.cs
```diff
@@ -3,49 +3,95 @@
...
+using System;
+using System.Reflection;
+using System.IO;
 
 namespace AccessibilityInsights.ExtensionsTests
 {
     [TestClass]
     public class ContainerUnitTests
     {
+        const string NoMatchOverride = "this_file_does_not_exist.dll";
```
RobGallo: These two variable names aren't so informative. How about something indicating that one path is valid and one isn't?

Issue #754, opened 2020-04-08
[ENG] Exception when starting under latest Visual Studio


DaveTryon:
Please check whether the bug has [already been filed](https://github.com/Microsoft/accessibility-insights-windows/issues).

**Describe the bug**
Our `ConfigurationManager` initialization code depends on a specific initialization order that can fail at app startup. Our previous workaround was to allow an Exception to be thrown, then silently eat it. This works, but the Exception is getting flagged when running in the latest version of Version Studio.

**To Reproduce**
1. Build the debug flavor of AIWin in the latest version of Visual Studio (16.5.3)
2. Start it in the debugger

**Expected behavior**
App should start up cleanly

**Actual behavior**
Debugger stops on a NullReferenceException in `ConfigurationManager.PopulateMainConfiguration`

PR #755, opened 2020-04-08
Refactor ConfigurationManager.GetDefaultInstance

DaveTryon:
Refactor ConfigurationManager.GetDefaultInstance to prevent NullReferenceException (instead of ignoring it)

#### Describe the change
`ConfigurationManager.GetDefaultInstance` contains some hackiness to allow for the fact that we sometimes ask for the configuration before we can fully answer the question. The old code caught and ate a NullReferenceException during initialization, but the current version of Visual Studio treats this as an unhandled Exception and stops in the debugger. This change prevents the Exception rather than eating it. In the offhand chance that an Exception still occurs, we catch it and log it. It makes no attempt to fix the ugliness inherent to how we're creating the singleton.

#### PR checklist

- [x] Does this address an existing issue? If yes, Issue# - #754 
- [ ] Includes UI changes?

### DIFF
--- a/src/AccessibilityInsights.SharedUx/Settings/ConfigurationManager.cs
+++ b/src/AccessibilityInsights.SharedUx/Settings/ConfigurationManager.cs
@@ -164,18 +164,9 @@ namespace AccessibilityInsights.SharedUx.Settings
         /// <returns></returns>
         public static ConfigurationManager GetDefaultInstance(FixedConfigSettingsProvider provider = null)
         {
-            if(sDefaultInstance == null)
+            if (sDefaultInstance == null && provider != null)
             {
-                try
-                {
-                    sDefaultInstance = new ConfigurationManager(provider);
-                }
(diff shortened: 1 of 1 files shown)


src/AccessibilityInsights.SharedUx/Settings/ConfigurationManager.cs
```diff
@@ -164,18 +164,9 @@ public void SaveEventsConfiguration()
         /// <returns></returns>
         public static ConfigurationManager GetDefaultInstance(FixedConfigSettingsProvider provider = null)
         {
-            if(sDefaultInstance == null)
+            if (sDefaultInstance == null && provider != null)
```
jalkire, line 167: Was `provider == null` the only situation we expect an exception here? I feel a little nervous removing the try...catch. Though I see that most of the work under the ConfigurationManager constructor is try/caught already.

Issue #756, opened 2020-04-09
[BUG] Disappearing icons in HC White mode


DaveTryon:
Please check whether the bug has [already been filed](https://github.com/Microsoft/accessibility-insights-windows/issues).

**Describe the bug**
The dark mode changes have caused some unintended changes in HC modes, at least one of which is a regression that impacts usability.

**To Reproduce**
1. Install the canary build
2. Enable HC White mode
3. Look at the icons on the command bar

**Expected behavior**
Icons should be visible

**Actual behavior**
Icons are not visible. This is a regression from the current production build.

**Screenshots or .GIF**
Current production build before dark mode changes (1.1.1152.01):
![image](https://user-images.githubusercontent.com/45672944/78846071-795bdf80-79bf-11ea-89cc-587e4184fa31.png)

Current canary build with incremental dark mode changes (1.1.1187.01):
![image](https://user-images.githubusercontent.com/45672944/78846080-8973bf00-79bf-11ea-94ad-a09d02d791af.png)

Observations:
1. Icons in HC White are white on white, so they disappear
2. Current HC White hover colors are black on purple, which fails contrast test
3. In HC modes, the command bar used to have the same background color as the caption bar. It now has the same background color as the main window area.
4. In HC White, the selection in the command bar disappear--this is probably related to the change in the background color.

The breadcrumb links in HC are now also rendered in the system link color--this is intentional since it helps draw attention to their link functionality.

**Desktop (please complete the following information):**
 - OS: Windows 10 1909
 - Accessibility Insights for Windows Version: See details

**Additional context**
Priority requested -

PR #757, opened 2020-04-09
Fix high contrast regression

DaveTryon:
#### Describe the change
Issue #756 calls out a regression that (as a result of the dark mode changes) several icons disappear in HC White mode. This change creates 4 new colors in the palette and leverages these color to ensure that HC modes (especially HC White) are working properly. It restore pre-dark mode parity, plus it fixes the fact that command bar icons were kind of broken on mouse hover in HC White mode. There are still some mouse hover issues in HC White mode, but those will get fixed in a separate PR.

Also, #756 only called out issues in Live mode. This also fixes a couple of icons in Test mode that had the same problem (no screenshots for these)

#### PR checklist

- [ x Does this address an existing issue? If yes, Issue# - #756 
- [x] Includes UI changes?

Production (before dark mode changes):
image

Canary (with dark mode changes):
image

Version with these changes:
image

Things to note:
- Light and dark remain unchanged
- Command bar backgrounds are back to what they are in Production
- Icons in non-hover HC White are visible again (same as Production)
- Icons in hover HC White have sufficient contrast (fixes an issue in Production)

### DIFF
--- a/src/AccessibilityInsights.SharedUx/Resources/Styles.xaml
+++ b/src/AccessibilityInsights.SharedUx/Resources/Styles.xaml
@@ -1473,4 +1473,14 @@
             </Setter.Value>
         </Setter>
     </Style>
+    <Style TargetType="fabric:FabricIconControl" x:Key="hoverAwareFabricIcon">
+        <Style.Triggers>
+            <Trigger Property="IsMouseOver" Value="True">
+                <Setter Property="Foreground" Value="{DynamicResource ResourceKey=IconHoverFGBrush}"/>
+            </Trigger>
+            <Trigger Property="IsMouseOver" Value="False">
+                <Setter Property="Foreground" Value="{DynamicResource ResourceKey=PrimaryFGBrush}"/>
+            </Trigger>
+        </Style.Triggers>
+    </Style>
 </ResourceDictionary>
--- a/src/AccessibilityInsights.SharedUx/Resources/Styles.xaml
+++ b/src/AccessibilityInsights.SharedUx/Resources/Styles.xaml
@@ -16,7 +16,7 @@
             <Setter.Value>
                 <ControlTemplate>
                     <Border BorderThickness="2" Margin="-2"
-                      BorderBrush="{DynamicResource ResourceKey=IconBrush}"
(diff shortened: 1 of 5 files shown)


src/AccessibilityInsights/MainWindow.xaml
```diff
@@ -345,15 +345,21 @@
                                                     <DataTrigger Binding="{Binding Path=vmHilighter.State, Mode=OneWay, RelativeSource={RelativeSource Mode=FindAncestor, AncestorType=local:MainWindow}}" Value="Off">
                                                         <Setter Property="GlyphName" Value="UnSetColor"/>
                                                     </DataTrigger>
+                                                    <Trigger Property="IsMouseOver" Value="True">
```
DaveTryon, line 348: @jalkire, I'd  much prefer to use the `hoverAwareFabricIcon` style here (instead of duplicating the triggers), but I wasn't able to find a neat way to get both the icon binding and the color binding. I'd welcome any suggestions you may have on this

src/AccessibilityInsights.SharedUx/Resources/Styles.xaml
```diff
@@ -16,7 +16,7 @@
             <Setter.Value>
                 <ControlTemplate>
                     <Border BorderThickness="2" Margin="-2"
-                      BorderBrush="{DynamicResource ResourceKey=IconBrush}"
+                      BorderBrush="{DynamicResource ResourceKey=FocusBorderBrush}"
```
DaveTryon, line 19: This is to decouple the focus highlight from the icon brush color (was needed only for HC White)

src/AccessibilityInsights/MainWindow.xaml
jalkire: You could try setting the [`BasedOn` property](https://docs.microsoft.com/en-us/dotnet/api/system.windows.style.basedon?view=netframework-4.8) of this style up on line 340 to point at `hoverAwareFabricIcon`. Unfortunately XAML's inheritance isn't super flexible so you might end up losing the parent triggers if you set new ones here, but hopefully not

PR #759, opened 2020-04-09
Improve highlighter contrast in HC modes

DaveTryon:
#### Describe the change
Highlighter and snapshot button contrast is a little odd in high contrast modes. It's something of an inexact science, since it depends on the pixels around the element being highlighted, but this change makes it a little better. We use the system-defined "Hot tracking color", which is defined [here](https://docs.microsoft.com/en-us/dotnet/api/system.windows.systemcolors.hottrackcolor?view=netframework-4.8) as "The color used to designate a hot-tracked item".

The Snapshot button background is drawn in the same color, and the snapshot button icon is drawn in a system color that is always black in HC Black themes, and white in the HC White theme.

#### PR checklist

- [x] Does this address an existing issue? If yes, Issue# - 1661682. Note that the bug is fairly vague, but the bug has been interpreted in triage as referring to the highlighter color in HC modes.
- [x] Includes UI changes?

Light and dark mode haven't changed, and we've made no attempt to respect HC settings in the tooltip. These screen grabs are all taken using calc.exe:
image

### DIFF
--- a/src/AccessibilityInsights/App.xaml.cs
+++ b/src/AccessibilityInsights/App.xaml.cs
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 using AccessibilityInsights.SharedUx.Enums;
+using AccessibilityInsights.SharedUx.Highlighting;
 using System;
 using System.Collections.Generic;
 using System.Windows;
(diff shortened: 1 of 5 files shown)


src/AccessibilityInsights/App.xaml.cs
```diff
@@ -85,6 +86,7 @@ public void SetColorTheme(Theme theme)
             Resources.MergedDictionaries.Remove(this.themeResourceDictionaty);
             this.themeResourceDictionaty = new ResourceDictionary() { Source = Brushes[theme] };
             Resources.MergedDictionaries.Add(this.themeResourceDictionaty);
+            HollowHighlightDriver.ClearAllHighlighters();
```
DaveTryon, line 89: This is needed to flush the cached brushes if color change

jalkire:
I believe tab stops and text patterns have different highlight colors. Might be worth updating those too, but doesn't need to be in this PR.

PR #792, opened 2020-05-15
Remove fakes from Extensions.TelemetryTests project

DaveTryon:
#### Describe the change
This removes the Fakes dependency from the `Extensions.TelemetryTests` project. It required some refactoring of the `Extensions.Telemetry` project to introduce test seams:

- `TelemetryClientWrapper` simply holds a `TelemetryClient` object, which is a sealed class and therefore incompatible with Moq
- `TelemetryClientFactory` just creates the `TelemetryClient` object. Having separate factory let me separate the test responsibilities
- In `OSHelper.GetVersion', I added an optional provider method to allow the calls to the registry to get replaced by tests. Since this is a static class, it seemed simplest to just make it an optional parameter to the method. That might change if/when more methods are added to this class.

On the test side, there was a similar breakout of classes, as well as the addition of the `TestRegistry` class. I thought about making `TestRegistry` a contained class in `OSHelpersUnitTests`, but decided to make it easier to reuse should we need it down the road. Code coverage of the `Exensions.Telemetry` project is now over 90% even when fakes aren't available. Gaps include the production ctor of `AITelemetry` (intentional) and the calls inside `TelemetryClientWrapper`, since I couldn't find a good way to validate that the call had been made. I'm open to ideas if anyone has a neat trick here.

Validate that telemetry is still flowing by forcing my installation ID to a known value, then querying all data sent by that installation ID in the last hour. Data matches my activity.

After this PR is merged, we'll have only 2 projects (`SharedUxTests` and `Extensions.AzureDevOpsTests`) that still rely on fakes

#### PR checklist

- [ ] Includes UI changes?


src/AccessibilityInsights.Extensions.Telemetry/AITelemetry.cs
```diff
@@ -81,11 +80,11 @@ private static void MergeIntoProperties(EventTelemetry aiEvent, IReadOnlyDiction
         }
 
         /// <summary>
-        /// Process an event asynchronously
+        /// Process an event
         /// </summary>
-        private async void ProcessEvent(EventTelemetry aiEvent)
+        private void ProcessEvent(EventTelemetry aiEvent)
         {
-            await Task.Run(new Action(() => TClient?.TrackEvent(aiEvent))).ConfigureAwait(false);
```
DaveTryon, line 88: The async behavior has been moved into the TelemetryClientWrapper to simplify testing

src/AccessibilityInsights.Extensions.Telemetry/TelemetryClientWrapper.cs
```diff
@@ -0,0 +1,33 @@
...
+        }
+
+        async void ITelemetryClientWrapper.TrackEvent(EventTelemetry data)
+        {
+            await Task.Run(new Action(() => _client.TrackEvent(data))).ConfigureAwait(false);
+        }
+
+        void ITelemetryClientWrapper.TrackException(Exception e, Dictionary<string, string> contextProperties)
+        {
+            _client.TrackException(e, contextProperties);
```
DaveTryon, line 30: This is retaining the synchronous behavior of TrackException. @karanbirsingh, what are your thoughts on making it asynchronous, like the events are?

src/AccessibilityInsights.Extensions.TelemetryTests/OSHelpersUnitTests.cs
```diff
@@ -1,57 +1,82 @@
...
+            int minorVersion = int.Parse(pieces[1]);
+            int build = int.Parse(pieces[2]);
 
-                Assert.AreEqual("99.88.77", OSHelpers.GetVersion());
-                Assert.AreEqual(2, index);
-            }
+            // Validation here assumes a range of Windows OS versions.
+            // Our lower bound is 6.1.7601  (Win 7 SP1)
+            // Our upper bound is 6.3.99999
+            // At current Win10 cadence, that will last until about 2040
```
DaveTryon, line 76: I wanted to strike a balance between unexpected changes in how Windows writes the version and creating a maintenance headache. I'll be astonished if this code is still running in 2020.

src/AccessibilityInsights.Extensions.Telemetry/TelemetryClientWrapper.cs
karanbirsingh: I had assumed both were fire & forget, so I'd be OK with that

src/AccessibilityInsights.Extensions.TelemetryTests/AITelemetryTests.cs
```diff
@@ -1,69 +1,151 @@
...
+            wrapperMock.VerifyAll();
         }
 
         [TestMethod]
         [Timeout(1000)]
-        public void ContextPropertiesRemainBetweenEventsTest()
+        public void ReportException_ExceptionIsNull_DoesNotCallWrapper()
         {
-            using (ShimsContext.Create())
+            Mock<ITelemetryClientWrapper> wrapperMock = new Mock<ITelemetryClientWrapper>(MockBehavior.Strict);
```
karanbirsingh, line 96: Does moq let us verify a non-call more explicitly?

src/AccessibilityInsights.Extensions.TelemetryTests/TelemetryClientFactoryUnitTests.cs
```diff
@@ -0,0 +1,26 @@
...
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace AccessibilityInsights.Extensions.TelemetryTests
+{
+    [TestClass]
+    public class TelemetryClientFactoryUnitTests
+    {
+        [TestMethod]
+        [Timeout(1000)]
+        public void Ctor_CreatesObjectWithExcpectedProperties()
```
karanbirsingh, line 14: Expected* small typo

src/AccessibilityInsights.Extensions.TelemetryTests/AITelemetryTests.cs
DaveTryon: With `MockBehavior.Strict`, the mock will throw an Exception on any attempt to call a method that doesn't have a corresponding `Setup` call.

src/AccessibilityInsights.Extensions.TelemetryTests/TelemetryClientFactoryUnitTests.cs
DaveTryon: Corrected

src/AccessibilityInsights.Extensions.Telemetry/TelemetryClientWrapper.cs
DaveTryon: Made them both fire & forget

PR #793, opened 2020-05-18
Remove fakes from the Extensions.AzureDevOpsTests project

DaveTryon:
#### Describe the change
Refactor the `Extensions.AzureDevOps` project in a way that allows us to write unit tests without using Fakes. This allowed for a simplification and clarification of the relationship between the various classes.

General points:
- `IDevOpsIntegration` is a new interface for `AzureDevOpsIntegration` so that we can abstract away the concrete implementation
- `AzureDevOpsIntegration` is no longer a singleton. All code inside the `AzureDevOpsIntegration` class uses instanced data to simplify testing.
- Only 1 instance of `AzureDevOpsIntegration` is created in production code. It is owned by the `AzureBoardsIssueReporting` class, which is what gets created by MEF
- The `FileIssueHelpers` class is no longer static. It accepts an `IDevOpsIntegration` object via its ctor, so it is now easier to test.
- The `FileIssueHelpers` class had multiple methods that were nothing more than wrappers around object creation or calling methods in `AzureDevOpsIntegration`. These were originally added as a way to create a seam for testing via Fakes, but without Fakes they add only complexity. They've been removed.
- Unit tests were disorganized--the `AzureBoardsIssueReportingTests` class was exclusively testing code in `FileIssueHelpers`, and there was some duplication of coverage between test classes. Unit tests are now more aligned to the classes being tested. Unit test coverage has increased slightly (and could increase much more by adding future tests, which is somewhat beyond the scope of removing fakes)

I tried to commit after each significant change, so it might be simpler to review one commit at a time.

#### PR checklist

Validated that AzureDevOps still works--connecting, disconnecting, loading and saving config, project/team enumeration, and issue filing. There should be no user-perceivable difference with these changes.

- [ ] Includes UI changes?


src/AccessibilityInsights.Extensions.AzureDevOps/AzureBoardsIssueReporting.cs
```diff
@@ -15,17 +15,51 @@ namespace AccessibilityInsights.Extensions.AzureDevOps
     [Export(typeof(IIssueReporting))]
     public class AzureBoardsIssueReporting : IIssueReporting
     {
-        private static AzureDevOpsIntegration AzureDevOps => AzureDevOpsIntegration.GetCurrentInstance();
+        private readonly FileIssueHelpers _fileIssueHelpers;
+        private readonly IDevOpsIntegration _devOpsIntegration;
+
+        /// <summary>
+        /// The one and only IDevOpsIntegration object used in production. It's static
```
DaveTryon, line 22: I'm not fully satisfied with this code. I'd much prefer to have a static setter on ConfigurationControl, but for reasons that elude me, it didn't want to resolve. I considered an event with a listener, but that also seemed like the wrong approach.

src/AccessibilityInsights.Extensions.AzureDevOps/AzureBoardsIssueReporting.cs
```diff
@@ -15,17 +15,51 @@ namespace AccessibilityInsights.Extensions.AzureDevOps
...
+        /// so that the ConfigurationControl code can read it. Note that code in this
+        /// class reads only the instanced version, not the static version.
+        /// </summary>
+        internal static IDevOpsIntegration DevOpsIntegration { get; private set; }
```
DaveTryon, line 26: I'm not fully satisfied with this code. I'd much prefer to have a static setter on ConfigurationControl, but for reasons that elude me, it didn't want to resolve. I considered an event with a listener, but that also seemed like the wrong approach. Actively soliciting idea on this one...

src/AccessibilityInsights.Extensions.AzureDevOps/FileIssue/FileIssueHelpers.cs
```diff
@@ -274,57 +288,7 @@ private static string GetTempFileName(string extension)
...
-        public static ConnectionInfo CreateConnectionInfo(Uri serverUri, TeamProject project, AdoTeam team)
-        {
-            return new ConnectionInfo(serverUri, project, team);
-        }
-
-        public static ConnectionInfo CreateConnectionInfo(string configString)
-        {
-            try
-            {
-                return new ConnectionInfo(configString);
```
DaveTryon, line 306: This ctor is only used in unit tests. I'll probably remove it, but likely in a separate PR

PR #836, opened 2020-07-07
Convert to new csproj file format

DaveTryon:
#### Describe the change
I apologize up front for the size of this PR. It replaces the old-format `.csproj` files (which dependabot doesn't understand) with their new formats. It follows the same pattern set in `Axe.Windows`, where many of the common settings are bundled into `.targets` files. Hopefully this will allow future dependabot changes to go through without manual help. It also greatly simplifies each individual `.csproj` file and centralizes settings that apply to all projects. These changes should be invisible to customers and should impact just the engineering process.

There are several elements that all have to land at the same time to avoid build breaks:
1. All of the `.csproj` files move to the new format. The signed build made it difficult to do these one at a time
2. We use the `PackageReferences` style of referencing projects. This does a better job of handling transitive dependencies
3. All `packages.config` files are deleted (they're not necessary when using `PackageReferences`)
4. Almost all version slides have been removed (they're dynamically generated when using `PackageReferences`)
5. Build loops include a `dotnet restore` task (needed for `PackageReferences`)
6. The analyzers complained about using `default(foo)` if `foo` was a reference type. Replaced these with `null`
7. Added some suppressions on warnings relating to `ImportingConstructor` in the extensions
8. Deleted the `AssemblyInfo.cs` files for the test projects
9. Replaces `System.Windows.Interactivity` with `Microsoft.Xaml.Behaviors` (the old one isn't supported with the new build tools)
10. Updated directories to reflect the extra layer (net472) in the build tree
11. Updated the file list in `Product.wxs` to reflect the new collection of assemblies.
12. Did some cleanup of app.config files (some of which went away completely)

Signed builds succeed (see 2020-07-07.04), and testing with that MSI hasn't yet revealed any unexpected changes.

I've temporarily disabled static analysis during the compliance build--it was behaving slightly differently (different counts of issues, possibly because of using `Microsoft.Xaml.Behaviors` instead of `System.Windows.Interactivity`), and given the size of this PR, I wanted to get the main change in, then circle back for the analysis differences.

#### PR checklist

- [n/a] Includes UI changes?


src/AccessibilityInsights.Extensions.AzureDevOps/FileIssue/FileIssueHelpers.cs
```diff
@@ -224,7 +225,7 @@ internal static string RemoveInternalHTML(string inputHTML, string keyText)
             {
                 foreach (IHTMLDOMNode divnode in divnodes)
                 {
-                    foreach (IHTMLDOMNode child in divnode.childNodes)
```
DaveTryon, line 227: devnode.childNodes no longer exposes GetEnumerator, so I had to rework it for enumeration

src/AccessibilityInsights.Extensions.AzureDevOps/FileIssue/FileIssueHelpers.cs
```diff
@@ -240,15 +241,23 @@ internal static string RemoveInternalHTML(string inputHTML, string keyText)
 
             if (node != null)
             {
-                foreach (var n in node.childNodes)
+                foreach (IHTMLDOMNode child in GetChildren(node).ToList())
                 {
-                    node.removeChild(n);
```
DaveTryon, line 245: The old code was modifying a list while traversing it, which can lead to problems. The new code traverses a copy of the list so there's no interference

src/AccessibilityInsights.SharedUx/Controls/CustomControls/CustomGridViewColumn.xaml.cs
```diff
@@ -73,7 +73,7 @@ internal string ContentPath
         #endregion ContentPath
         #endregion DependencyProperties
 
-        internal CustomGridViewColumn() => InitializeComponent();
+        public CustomGridViewColumn() => InitializeComponent();
```
DaveTryon, line 76: This needs to be public for XAML parsing with the new libraries

src/AccessibilityInsights.SharedUx/Resources/Styles.xaml
```diff
@@ -4,11 +4,11 @@
                     xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
                     xmlns:po="http://schemas.microsoft.com/winfx/2006/xaml/presentation/options"
                     xmlns:fabric="clr-namespace:AccessibilityInsights.CommonUxComponents.Controls;assembly=AccessibilityInsights.CommonUxComponents"
-                    xmlns:animations="clr-namespace:AccessibilityInsights.SharedUx.Animations;assembly=AccessibilityInsights.SharedUx"
```
DaveTryon, line 7: More changes related to the new XAML assemblies

src/AccessibilityInsights.SetupLibraryUnitTests/SetupLibraryUnitTests.csproj
```diff
@@ -1,85 +1,29 @@
...
-    <TargetFrameworkProfile />
-  </PropertyGroup>
-  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
-    <DebugSymbols>true</DebugSymbols>
-    <DebugType>full</DebugType>
-    <Optimize>false</Optimize>
-    <OutputPath>bin\Debug\</OutputPath>
-    <DefineConstants>DEBUG;TRACE</DefineConstants>
-    <ErrorReport>prompt</ErrorReport>
-    <WarningLevel>4</WarningLevel>
```
jalkire, line 32: I don't see anything analogous to WarningLevel 4 in the new setup. Has it been renamed or made default?

DaveTryon: Interesting. VS shows everything as Warning level 4 even though it's not explicitly set. I explicitly added it to build/NetFrameworkAll.targets, just to be sure :)

PR #851, opened 2020-07-22
Copy files to signing validation directory & scan it

jalkire:
#### Describe the change
Currently, the auto injected signing validation task scans our entire build directory, which results in hundreds of false positives. This PR updates our signed build to copy the assemblies and executables that we ship to a particular directory and directs the signing validation task to scan only that directory. You can find an example of the files copied to the SigningValidation directory [here](https://dev.azure.com/accessibility-insights/Accessibility%20Insights/_build/results?buildId=11813&view=artifacts&type=publishedArtifacts).

#### PR checklist

- [n/a] Includes UI changes?

### DIFF
--- a/build/signedbuild.yml
+++ b/build/signedbuild.yml
@@ -201,6 +201,7 @@ jobs:
   variables:
     runCodesignValidationInjection: 'true'
     SignAppForRelease: 'true'
+    GDN_CODESIGN_TARGETDIRECTORY: '$(Build.ArtifactStagingDirectory)\SigningValidation'
 
   steps:
   - task: ms-vseng.MicroBuildTasks.a0262b21-fb8f-46f8-bb9a-60ed560d4a87.MicroBuildLocalizationPlugin@1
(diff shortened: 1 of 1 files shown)


build/signedbuild.yml
```diff
@@ -310,6 +311,16 @@ jobs:
       uploadRoslyn: false
       uploadTSLint: false
 
+  - task: CopyFiles@2
+    displayName: 'Copy Files for Signing Validation'
+    inputs:
+      SourceFolder: '$(Build.SourcesDirectory)\src'
+      Contents: |
+       AccessibilityInsights\bin\Release\net472\**\?(*.exe|*.dll)
+       AccessibilityInsights.VersionSwitcher\bin\Release\net472\**\?(*.exe|AccessibilityInsights.*.dll)
```
DaveTryon, line 320: Any thoughts on ways that we might be able to future-proof this in case VersionSwitcher somehow picked up additional 3rd party assemblies?

jalkire: My only issue with the exclude approach is that if we add a new external dependency in the same manner as the existing dependencies (ie, one that is signed in the AccessibilityInsights project), we would then have to update the exclude list. 

That being said, if the situation we want to avoid is one where we inadvertently ship an unsigned dependency, the exclude approach will ensure that the signing validation task still catches the issue. I'll update the PR 🙂

Issue #814, opened 2020-06-08
[BUG] Unable to view the UIA tree once the pane width has been reduced to zero 


guybark:
1. Run the AIWin tool and point it to some UI element to populate AIWin's UIA tree and details panes.
2. Using the mouse, reduce the width of the pane containing the UIA tree to zero. I did this deliberately in order to be able to generate a screenshot on a fairly small screen, where I needed to capture some target app's UI and AIWin's Details pane, but not the UIA tree. When repro'ing this, make sure, the pane width is reduced to zero, not just close to zero.
3. Close AIWin.
4. Open AIWin and try to get the UIA tree pane back with the mouse.

**Expected behavior**
Somehow the mouse can be used to get the UIA tree pane back.

**Actual behavior**
The double-header resizing arrow doesn't appear when trying to resize the now-invisible UIA tree pane, so there seems no way to get the pane back. Uninstalling and re-installing AIWin didn't help. In the end I searched the AIWin code to find where user settings are stored, which led me to the local Layout.json file, and I deleted it. When I next ran AIWin, its UIA had reverted to its default values.

The screenshot below shows the AIWin tool with the UIA tree pane reduced to zero-width.

**Screenshots or .GIF**
![image](https://user-images.githubusercontent.com/14315611/84068966-9fc9d700-a97e-11ea-8dd9-1d84d5d28022.png)

**Desktop (please complete the following information):**
 - OS: Windows 10 1909
 - Accessibility Insights for Windows Version: 1.1.1248.1

DaveTryon:
Plan is to set a minimum width of 1 or 2 pixels on each side of the splitter, so that the splitter can always be retrieved with the mouse. @guybark, the splitter is already always accessible via keyboard (press tab until the splitter is highlighted, then use the arrow keys to reposition).

guybark:
@DaveTryon - Thanks for the note on how to use the keyboard to change the width of the panes, that's really helpful! The splitter's keyboard focus feedback seems to respect colors from the active high contrast theme too. Nice!

DaveTryon:
@guybark , please regress this using the current canary build. Thanks!

PR #852, opened 2020-07-23
Set minimum width in mode control left columns

jalkire:
#### Describe the change
Currently, if a user reduces a left column width (such as the hierarchy tree in live mode) to 0, they can no longer use the mouse to resize that column. This PR addresses that concern by setting the min column width of applicable columns to 2. This ensures the user can always resize the column with their mouse.

New behavior:
column

This PR addresses [BUG] Unable to view the UIA tree once the pane width has been reduced to zero #814

#### PR checklist

- [x] Does this address an existing issue? If yes, Issue #814
- [x] Includes UI changes?

### DIFF
--- a/src/AccessibilityInsights/Modes/EventModeControl.xaml
+++ b/src/AccessibilityInsights/Modes/EventModeControl.xaml
@@ -20,7 +20,7 @@
             <RowDefinition Height="*"/>
         </Grid.RowDefinitions>
         <Grid.ColumnDefinitions>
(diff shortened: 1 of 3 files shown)


DaveTryon:
Do we need to consider a minimum width for the right-hand panes, as well?

jalkire:
> Do we need to consider a minimum width for the right-hand panes, as well?

Probably a good idea. It looks like adding that won't be trivial, so I will follow up with that in another PR.

Issue #809, opened 2020-06-02
[BUG] Azure Boards zoom level isn't being persisted


DaveTryon:
**Describe the bug**
The zoom level for the Azure Boards bug filing window is not getting updated when the user changes it in the control.

**To Reproduce**
1. Scan Wildlife Manager
2. Pick one of the failures
3. File an issue against that issue
4. Change the zoom level in the bug filing window
5. Save the bug, then exit AIWin
6. Restart AIWin
7. Re-scan Wildlife Manager
8. Pick one of the failures
9. File an issue against the issue

**Expected behavior**
The zoom level should be persisted from the previous session

**Actual behavior**
The change to the zoom level didn't persist.

**Screenshots or .GIF**

**Desktop (please complete the following information):**
 - OS: WIndows 10 2004
 - Accessibility Insights for Windows Version: 1.1.1248.1
 - Target Application: Wildlife Manager

**Additional context**
I did some experimentation by manually editing the config file, and the value in the config file *is* being respected--it looks like the problem is that the config file isn't being updated when the zoom level is changed in the bug filing window..

DaveTryon:
I'll regress this in tomorrow's canary build

DaveTryon:
Verified fixed in 1.1.1300.1

PR #853, opened 2020-07-23
fix: Persist zoom level from ADO issue filing

DaveTryon:
#### Describe the change
As called out in #809, we're failing to save user customizations in zoom level. The root cause is that when we make each IIssueReporting responsible to serialize its own settings, we assumed (incorrectly) that settings would only ever change from the settings tab. As such, when the user zoom is customized, it never gets written to the config file. 

The fix is to add a new method to IIssueReporting, which optionally exposes the current serialized object, and then update this after calling FileIssueAsync. If the IIssueReporting implementation provides information, save it to the config file.

Test:
1. Open AIWin
2. Scan an app (Wildlife Manager will do nicely)
3. Start bug filing via ADO
4. Change the zoom level from its original value
5. Exit the bug filing (doesn't matter if you file the bug)
6. Exit AIWin
7. Open AIWin again
8. Scan an app
9. Start bug filing via ADO

Validation: The zoom level you set before should be restored.

#### PR checklist

- [x] Does this address an existing issue? If yes, Issue# ; #809 
- [n/a] Includes UI changes?

### DIFF
--- a/src/AccessibilityInsights.SharedUx/FileIssue/IssueReporterManager.cs
+++ b/src/AccessibilityInsights.SharedUx/FileIssue/IssueReporterManager.cs
@@ -101,6 +101,11 @@ namespace AccessibilityInsights.SharedUx.FileIssue
                 JsonConvert.SerializeObject(configsDictionary);
         }
 
+        /// <summary>
+        /// Call this when settings for an IIssueReporting object may have changed. The
+        /// settings will be fetched and updated in the store.
+        /// </summary>
+        /// <param name="issueReporting">The object whose settings may have changed</param>
         public void UpdateIssueReporterSettings(IIssueReporting issueReporting)
         {
             if (issueReporting == null)
--- a/src/AccessibilityInsights.SharedUx/FileIssue/IssueReporterManager.cs
+++ b/src/AccessibilityInsights.SharedUx/FileIssue/IssueReporterManager.cs
@@ -19,6 +19,8 @@ namespace AccessibilityInsights.SharedUx.FileIssue
         readonly static object _lockObject = new object();
         Dictionary<Guid, IIssueReporting> IssueReportingOptionsDict = new Dictionary<Guid, IIssueReporting>();
         private static IssueReporterManager _defaultInstance = null;
+        private readonly ConfigurationModel _appConfig;
+        private readonly IEnumerable<IIssueReporting> _issueReportingOptions;
 
         public static IssueReporterManager GetInstance()
         {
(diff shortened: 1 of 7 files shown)


src/AccessibilityInsights.SharedUx/FileIssue/IssueReporterManager.cs
```diff
@@ -42,22 +46,28 @@ private IssueReporterManager()
...
-                JsonConvert.DeserializeObject<Dictionary<Guid, string>>(serializedConfigsDict)
-                : new Dictionary<Guid, string>();
+            if (appConfig == null)
+                throw new ArgumentNullException(nameof(appConfig));
+
+            _appConfig = appConfig;
+            _issueReportingOptions = issueReportingOptions ?? Enumerable.Empty<IIssueReporting>();
+        }
+
+        internal void RestorePersistedConfigurations()
```
DaveTryon, line 58: I changed this class to use 2-step initialization instead of doing everything in the ctor. Doesn't impact the overall effect, but it caused changes in the unit tests

src/AccessibilityInsights.SharedUx/FileIssue/IssueReporterManager.cs
```diff
@@ -76,6 +86,39 @@ internal IssueReporterManager(ConfigurationModel configs, IEnumerable<IIssueRepo
             }
         }
 
+        private Dictionary<Guid, string> GetConfigsDictionary()
```
DaveTryon, line 89: This used to be in lines 46-50. Extracted it out since I needed it twice

src/AccessibilityInsights.SharedUxTests/FileIssue/IssueReporterManagerTests.cs
```diff
@@ -143,11 +174,88 @@ public void SetIssueReporter_ReporterNonExistent_ReporterUnset()
...
-            issueReporterMock.Setup(p => p.StableIdentifier).Returns(TestGuid);
-            issueReporterMock.Setup(p => p.RestoreConfigurationAsync(TestReporterConfigs)).Returns(new Task<IIssueResult>(() => { return null; }));
+            ConfigurationModel configs = GetConfigurationModel(null);
+            IssueReporterManager repManager = new IssueReporterManager(configs, Enumerable.Empty<IIssueReporting>());
+            Mock<IIssueReporting> issueReportingMock = GetIssueReporterMock(
+                expectRestoreConfig: false, supportGetSerializedSettings: true);
+
+            repManager.UpdateIssueReporterSettings(issueReportingMock.Object);
+
+            issueReportingMock.VerifyAll();
```
karanbirsingh, line 213: should this test do something to verify the contents of `configs`?

DaveTryon: Added a check

Issue #840, opened 2020-07-16
[BUG] AI for Windows throws an error using Text Pattern Explorer with Edge (both new and legacy)


kbabbitt:
**Describe the bug**
"Failed to retrieve range(s): Object reference not set to an instance of an object." when exploring TextPattern on a text box in Chromium-based Edge or Edge Legacy.

**To Reproduce**
1. Start Accessibility Insights for Windows
2. Start Chromium-based Edge or Edge Legacy
3. Hover over the search box on the new tab page
4. In AI, click on "Text Pattern Explore..."

**Expected behavior**
Text Pattern Explorer dialog appears without errors

**Actual behavior**
Text Pattern Explorer dialog appears with "Failed to retrieve range(s): Object reference not set to an instance of an object" error.

**Desktop (please complete the following information):**
 - Windows 10 1909 (OS Build 18363.959)
 - Accessibility Insights for Windows Version: 1.1.1248.1
 - Target Application: Microsoft Edge 84.0.522.40, Microsoft Edge Legacy 44.18362.449.0

adarshrema:
Needs investigation by the engineering team if this is an issue the Edge team needs to fix or if that has to happen at our end.

DaveTryon:
This has been merged to master and will be in the next canary build. Will resolve once the build is released

DaveTryon:
@kbabbitt, please regress against the latest canary release (https://github.com/microsoft/accessibility-insights-windows/releases/tag/v1.1.1307.01), then close if the fix works for you. Thanks!

kbabbitt:
Looks good to me, thank you for the fix!

PR #881, opened 2020-07-29
Fix null dereference in text patterns

DaveTryon:
#### Describe the change
As called out in #840, there's a null dereference when using the text patterns dialog. The root cause was overlooking the fact that highlighter colors are prefixed with "HL", and attempting to use a color without this prefix. We couldn't find a color, returned null, and caused a null dereference exception. This is a regression of the palette work for dark mode, so this change essentially reverts the change to the highlighter color and adds comments to help avoid similar breaks in the future.

One thing that is _not_ addressed in this change is that since we choose our highlighter color with no awareness of the colors being used by the target app, there's no guarantee of the visual impact that the highlighter will have in the context of the target app. Our highlighter is always yellow, which seems like an OK starting point and we can tune it in future changes. Just for reference, here are samples of the highlighter in the Chromium-based edge in various color modes--note that Edge isn't yet responding to dark mode, so Light mode and Dark mode are identical. The extra highlighter in HC-Black mode is just because I happened to have the mouse hovering when I grabbed the screenshot:

image

I'd love to add unit tests for highlighters, but I'm not really sure how much value there is there since it relies upon the resources that are external to the class. Feels like something that might work in UITests, but we'd have to work out the details. For now, the main focus is to correct the Exception.

#### PR checklist

- [x] Does this address an existing issue? If yes, Issue# - #840 
- [x] Includes UI changes?

### DIFF
--- a/src/AccessibilityInsights.SharedUx/Highlighting/Highlighter.cs
+++ b/src/AccessibilityInsights.SharedUx/Highlighting/Highlighter.cs
@@ -207,7 +207,7 @@ namespace AccessibilityInsights.SharedUx.Highlighting
         /// <returns></returns>
         private static SolidColorBrush GetBrush(HighlighterColor color)
         {
-            return Application.Current.Resources[$"HL{color.ToString()}"] as SolidColorBrush;
+            return Application.Current.Resources[$"HL{color}"] as SolidColorBrush;
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


src/AccessibilityInsights.SharedUx/Highlighting/Highlighter.cs
```diff
@@ -207,7 +207,7 @@ public void SetCallBackForSnapshot(Action callback)
         /// <returns></returns>
         private static SolidColorBrush GetBrush(HighlighterColor color)
         {
-            return Application.Current.Resources[$"HL{color.ToString()}"] as SolidColorBrush;
+            return Application.Current.Resources[$"HL{color}"] as SolidColorBrush;
```
DaveTryon, line 210: This is where the "HL" gets prefixed. VS flagged the `ToString()` call as superfluous, so I removed it and it works as expected.

Issue #815, opened 2020-06-08
[BUG] AIWin Events TreeView expandable nodes should have 3-state checkboxes


guybark:
**Describe the bug**
The Events checkbox UI is unclear as to what events will be reported.

**To Reproduce**
1. Run AIWin and get to its Events treeview in preparation to listening for FocusChanged and LiveRegionChanged events.
2. Check and uncheck individual events' radio buttons until only the checkboxes for the FocusChanged and LiveRegionChanged events are checked.

**Expected behavior**
The Events UI should be unambiguous as to what events are going to be reported. As such, if a parent treeview node has some child nodes with a checked checkbox, and some child nodes with an unchecked checkbox, the parent node's checkbox should have an Indeterminate state.

**Actual behavior**
Depending on the steps taken for checking the events' checkboxes of interest, the checkboxes for the parent treeview nodes may or may not be checked. If that checkbox is unchecked, some customer may feel that all its child nodes are effectively unchecked, despite some of the child nodes' checkbox being checked.

The screenshot below shows a parent node whose checkbox is checked, and another whose checkbox is unchecked. Both nodes have one child node whose checkbox is checked.

**Screenshots or .GIF**
![image](https://user-images.githubusercontent.com/14315611/84070350-dacd0a00-a980-11ea-8705-9d4ef425c830.png)

**Desktop (please complete the following information):**
 - OS: Windows 10 1909
 - Accessibility Insights for Windows Version: 1.1.1248.1

adarshrema:
Confirmed with @guybark that the parent nodes for the event tree are,
* Expected events.. by control type
* My events

DaveTryon:
@guybark, this is fixed in https://github.com/microsoft/accessibility-insights-windows/releases/tag/v1.1.1314.01. Please regress at your convenience

DaveTryon:
@guybark, I'd suggest regressing in https://github.com/microsoft/accessibility-insights-windows/releases/tag/v1.1.1314.02, just because that build has the fix that allows you to successfully switch channels

PR #888, opened 2020-08-06
fix: Support ThreeState checkboxes when selecting events to listen to

DaveTryon:
#### Describe the change
As described in #815, the checkboxes in the event selection dialog don't follow the 3-state hierarchical selection pattern and are potentially confusing. This PR adds the plumbing needed to support the expected 3-state selection pattern. It includes the following changes:

- Expose an `IsThreeState` property to the `EventConfigNodeViewModel` class, and provide a way to set it when the node is constructed.
- Bind the `IsThreeState` property to the 3 key checkboxes in the configuration page
- Provide a notification to a ThreeState parent when any of its children change state or if the set of children is modified
- Teach the ThreeState parent to recalculate its `IsChecked` state in response to a modified child
- The natural order of a ThreeState checkbox is On, Mixed, Off, and that order will be respected in response to user actions. We only want to support On or Off in response to user actions on a ThreeState checkbox, so intentionally remap the Mixed state to Off in response to user actions on the ThreeState checkbox
- To ensure that future changes continue to preserve the parent/child linkage, change the type of the public `EventConfigNodeViewModel.Children` to be an `IReadOnlyCollection`, and make the editable version `private readonly`. I had originally resisted this but found a fairly painless way to incorporate it into this change.

Behavior before these changes:
Checkboxes Before

Behavior after these changes:
Checkboxes After

#### PR checklist

- [x] Does this address an existing issue? If yes, Issue# - #815 
- [x] Includes UI changes?

### DIFF
--- a/src/AccessibilityInsights.SharedUx/Controls/EventConfigTabControl.xaml.cs
+++ b/src/AccessibilityInsights.SharedUx/Controls/EventConfigTabControl.xaml.cs
@@ -147,7 +147,7 @@ namespace AccessibilityInsights.SharedUx.Controls
          
             if (CustomPropertiesNode.Children.Count > 0)
             {
-                CustomNode.Children.Add(CustomPropertiesNode);
+                CustomNode.AddChild(CustomPropertiesNode);
             }
         }
 
--- a/src/AccessibilityInsights.SharedUx/Controls/EventConfigTabControl.xaml.cs
+++ b/src/AccessibilityInsights.SharedUx/Controls/EventConfigTabControl.xaml.cs
@@ -74,7 +74,7 @@ namespace AccessibilityInsights.SharedUx.Controls
             RootNodes = new List<EventConfigNodeViewModel>();
             trviewConfigEvents.ItemsSource = RootNodes;
             var ids = SupportedEvents.GetEventsForControl(el.ControlTypeId, el.Patterns);
-            SuggestedNode = new EventConfigNodeViewModel(string.Format(CultureInfo.InvariantCulture, Properties.Resources.EventConfigTabControl_SetElement_Expected_Events_based_on_the_0_control_type, Axe.Windows.Core.Types.ControlType.GetInstance().GetNameById(el.ControlTypeId))) { IsExpanded = true };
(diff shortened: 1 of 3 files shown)


src/AccessibilityInsights.SharedUx/Controls/EventConfigTabControl.xaml.cs
```diff
@@ -96,7 +96,7 @@ public void SetElement(A11yElement el)
                 RootNodes.Add(SuggestedNode);
             }
 
-            CustomNode = new EventConfigNodeViewModel("My Events");
+            CustomNode = new EventConfigNodeViewModel("My Events", isThreeState: true);
 
             CustomPropertiesNode = new EventConfigNodeViewModel("Properties") { Depth = 1 };
```
jalkire, line 101: I believe this node should be tri state as well, since it has children.
```suggestion
            CustomPropertiesNode = new EventConfigNodeViewModel("Properties", isThreeState: true) { Depth = 1 };
```

DaveTryon: Updated :)

src/AccessibilityInsights.SharedUx/ViewModels/EventConfigNodeViewModel.cs
```diff
@@ -22,47 +22,117 @@ namespace AccessibilityInsights.SharedUx.ViewModels
...
         /// </summary>
-        private bool _ischecked = false;
-        public bool IsChecked
+        private bool? _ischecked = false;
+        public bool? IsChecked
         {
             get
             {
-                return this._ischecked;
+                return this.IsThreeState ? this._ischecked :
```
karanbirsingh, line 49: To me this feels easier to parse w/o the nested ternary, would something like `
return this.isThreeState? this._ischecked : this._ischecked == true` work?

DaveTryon: Missed this before merging. It's in #890

Issue #887, opened 2020-08-04
copy-to-clipboard behavior is not consistent


adarshrema:
Spinning this off #880.

Confirm that copying (Ctrl+C) text from automated checks result is working as expected while,
* Selecting an individual item from automated checks result list to copy failure details
* Selecting different parts of UI Automation Tree view to copy details like properties, description

Once this behavior is confirmed, we will update documentation accordingly.

DaveTryon:
The problem here is that AIWin is copying on KeyUp of Ctrl+C, while the standard for Windows apps is to copy on KeyDown of Ctrl+C. Either will work if the keyboard pattern is Ctrl Down, C Down, C Up, Ctrl Up. However, decades of muscle memory have trained my left hand to do Ctrl Down, C Down, Ctrl Up, C up. When that happens, Ctrl is no longer down when the KeyUp is received for C, and copy doesn't trigger. I've created a PR to trigger copy on KeyDown of C, so we're consistent with the way the OS handles Copy.

DaveTryon:
Confirmed in build 1.1.1320.1

PR #893, opened 2020-08-07
fix: Change Ctrl+C from KeyUp to KeyDown (typical for Windows apps)

DaveTryon:
#### Describe the change
The "standard" for Windows apps (based on notepad, Office, and many others) is to trigger copy on KeyDown of Ctrl+C. AIWin is triggering on KeyUp of Ctrl+C. It probably shouldn't make much difference unless a user's muscle memory (like mine) is to do them quickly enough that the C is released slightly _after_ the Ctrl key. Copying will appear to be broken for users who do this, which is what led to #887. With this change in place, AIWin behaves like other Windows apps and copying feeling robust.

#### PR checklist

- [x] Does this address an existing issue? If yes, Issue# - #887 
- [n/a] Includes UI changes?

### DIFF
--- a/src/AccessibilityInsights.SharedUx/Controls/PropertyInfoControl.xaml.cs
+++ b/src/AccessibilityInsights.SharedUx/Controls/PropertyInfoControl.xaml.cs
@@ -231,7 +231,7 @@ namespace AccessibilityInsights.SharedUx.Controls
         /// </summary>
         /// <param name="sender"></param>
         /// <param name="e"></param>
-        private void lvProperties_KeyUp(object sender, KeyEventArgs e)
+        private void lvProperties_KeyDown(object sender, KeyEventArgs e)
         {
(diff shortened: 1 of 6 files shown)


src/AccessibilityInsights.SharedUx/Controls/PropertyInfoControl.xaml.cs
```diff
@@ -231,7 +231,7 @@ private void UserControl_Loaded(object sender, RoutedEventArgs e)
         /// </summary>
         /// <param name="sender"></param>
         /// <param name="e"></param>
-        private void lvProperties_KeyUp(object sender, KeyEventArgs e)
+        private void lvProperties_KeyDown(object sender, KeyEventArgs e)
```
DaveTryon, line 234: @jalkire, there's currently no KeyUp or KeyDown method in PropertyInfoControl.xaml. Should there be?

PR #901, opened 2020-08-24
Fix ui tests

jalkire:
#### Describe the change
Currently, there are two issues with our UI tests:
1. The live mode UI tests fail because axe-windows has been updated and reports a different number of results and because we changed our directory structure.
2. The automation to load test and event files fails on my machine (but not the remote build).

This PR addresses 1 by updating the expected result count and WildlifeManager executable path and 2 by removing some unnecessary automation complexity.

#### PR checklist

- [n/a] Includes UI changes?

### DIFF
--- a/src/UITests/LiveMode.cs
+++ b/src/UITests/LiveMode.cs
@@ -11,7 +11,7 @@ namespace UITests
     [TestClass]
     public class LiveMode : AIWinSession
     {
-        readonly string TestAppPath = Path.GetFullPath("../../../../tools/WildlifeManager/WildlifeManager.exe");
+        readonly string TestAppPath = Path.GetFullPath("../../../../../tools/WildlifeManager/WildlifeManager.exe");
         Process _wildlifeManager;
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


src/UITests/LiveMode.cs
```diff
@@ -38,7 +38,7 @@ private void TestTestResults()
 
         private void ValidateResultsInUIATree()
         {
-            int expectedResultsCount = 12; // this number should only change if axe-windows or WildlifeManager are modified.
+            int expectedResultsCount = 13; // this number should only change if axe-windows or WildlifeManager are modified.
```
DaveTryon, line 41: This seems to be bound to the new Axe.Windows version, while the rest is more infrastructure. Would it make sense to have a PR for infrastructure, then update the results count as part of the PR to update the Axe.Windows version?

jalkire: Since the axe-windows change was not recent, I feel like it makes sense to group these changes as "fixes to broken tests". In the future, we'll ideally update this test at the same time that it would otherwise start to fail.

PR #917, opened 2020-09-02
Generate thirdpartynotices.html during signed build

jalkire:
#### Describe the change
Currently, we have a manually maintained thirdpartynotices.html file checked into our repo. Updating this file is annoying and easy to forget. This PR updates our signed build loop to generate that file automatically. The file in our repo is updated to be a placeholder for local builds. You can find an example of a generated notice file in the [artifacts of this build](https://dev.azure.com/mseng/1ES/_build/results?buildId=12997829&view=artifacts&type=publishedArtifacts).

[An equivalent PR exists for axe-windows](https://github.com/microsoft/axe-windows/pull/434).

#### PR checklist

- [n/a] Includes UI changes?


DaveTryon (approved):
_comment deleted_

DaveTryon (changes requested):
Something isn't right here... this looks like the notices file for Axe.Windows, not for AIWin. AIWin should include Axe.Windows, the ADO assemblies, etc.

DaveTryon (approved):
This is great--should devDependencies (MSTest*, codeCov, coverlet, Moq, Castle.Core, *Analyzers) be filtered out?

Also, it's interesting that there are 2 thirdpartynotices.html in the artifacts. They appear to be identical, but should there be more than one? It looks like axe.windows is only publishing one

jalkire:
> This is great--should devDependencies (MSTest*, codeCov, coverlet, Moq, Castle.Core, *Analyzers) be filtered out?
> 
> Also, it's interesting that there are 2 thirdpartynotices.html in the artifacts. They appear to be identical, but should there be more than one? It looks like axe.windows is only publishing one

The configuration for what gets generated is managed in the [Component Governance portal](https://dev.azure.com/mseng/1ES/_componentGovernance)--I think I have it updated correctly now, but it's a little tricky because the configuration is branch-specific, so I can't know for sure until this is in master. 

For the 2 thirdpartynotices, I believe that happened because I reran failed build steps due to a timeout on signing, so the file got generated/attached twice.

PR #925, opened 2020-09-04
feat(tt-aug-2020): Remove WCAG links from Automated Checks view

DaveTryon:
#### Describe the change
_Note: This is a proposal, written as though it's a done deal. This comment will be removed if/when we decide to follow this approach._

The WCAG link that is currently included in the automated checks header has no UIA representation, making it invisible to assistive technologies. After iterating with various experts and finding no viable way to fully support the WCAG link, we've opted to remove the WCAG link from the automated checks view. The link  information is still available if you view the result in the UIA tree, and since we already have many other fields that are available only in that way, this shouldn't be too much of an issue for existing users.

Another piece of feedback was that it's generally preferable to have Enter and Space perform the same action on a control. We used to use Enter to trigger the link, but now that the link has been removed, Enter now does the same thing as Space (i.e., select or deselect the items in the group). The help text for the group header has been updated accordingly.

Validated with NVDA to confirm that the link is no longer spoken, and that the help text matches the new behavior.

Screenshot in automated checks view:
image

Here's an A11yTest file showing the full UIA tree (GitHub made me zip it up since it's an unrecognized file format):
[AutomatedChecksWithLinksRemoved.a11ytest.zip](https://github.com/microsoft/accessibility-insights-windows/files/5177331/AutomatedChecksWithLinksRemoved.a11ytest.zip)

#### PR checklist

- [x] Does this address an existing issue? If yes, Issue# [1750586](https://mseng.visualstudio.com/1ES/_workitems/edit/1750586)
- [x] Includes UI changes?

### DIFF
--- a/src/AccessibilityInsights.SharedUx/Controls/TestTabs/AutomatedChecksControl.xaml.cs
+++ b/src/AccessibilityInsights.SharedUx/Controls/TestTabs/AutomatedChecksControl.xaml.cs
@@ -600,11 +600,6 @@ namespace AccessibilityInsights.SharedUx.Controls.TestTabs
                 }
                 e.Handled = true;
             }
-            else if (e.Key == Key.Return && Keyboard.FocusedElement is ListViewItem)
-            {
-                var btn = GetFirstChildElement<Button>(sender as DependencyObject) as Button;
-                ButtonElem_Click(btn, e);
-            }
         }
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


src/AccessibilityInsights.SharedUx/Controls/TestTabs/AutomatedChecksControl.xaml.cs
```diff
@@ -873,10 +867,6 @@ private void GroupItem_PreviewKeyDown(object sender, KeyEventArgs e)
                 {
                     exp.IsExpanded = true;
                 }
-                else
-                {
-                    urlLink?.Focus();
-                }
                 e.Handled = true;
             }
             else if (e.Key == Key.Left)
```
karanbirsingh, line 872: we might be able to remove the code for going left from the hyperlink too (line 884)

karanbirsingh (approved):

src/AccessibilityInsights.SharedUx/Controls/TestTabs/AutomatedChecksControl.xaml.cs
DaveTryon: Removed this and the hyperlink launching code

PR #933, opened 2020-09-14
feat(tt-aug_2020): Relocate the source link from the group header to each list item

DaveTryon:
#### Describe the change
The current UI puts the rule source in the group header in automated results view. This looks great but has accessibility issues because the link never makes it into the UIA tree. Feedback from users is that having the link at the top level is useful, so we're going with the approach of putting the link in each row. It's busy visually, but we'll see what sort of feedback we get and adjust accordingly.

The rule source link is announced a button instead of a hyperlink to better fit into the custom handling within the control.

#### PR checklist

- [x] Does this address an existing issue? If yes, Issue# - [1750586](https://mseng.visualstudio.com/1ES/_workitems/edit/1750586)
- [x] Includes UI changes?

image

### DIFF
--- a/src/AccessibilityInsights.SharedUx/Controls/TestTabs/AutomatedChecksControl.xaml
+++ b/src/AccessibilityInsights.SharedUx/Controls/TestTabs/AutomatedChecksControl.xaml
@@ -285,10 +285,6 @@
                                                     <local:CustomExpander.Header>
                                                         <StackPanel Focusable="False" Orientation="Horizontal" Height="20">
                                                             <Label Content="{Binding Path=Items[0].Description}" Style="{StaticResource lblRowStyle}" VerticalAlignment="Center" />
-                                                            <Label Margin="4,0,0,0" Padding="0" Style="{StaticResource lblLink}" 
-                                                                   VerticalAlignment="Center" FontWeight="SemiBold"
-                                                                   AutomationProperties.Name="{Binding Path=Items[0].Source, StringFormat='{x:Static Properties:Resources.lvResultsLabelAutomationPropertiesName}'}">
-                                                            </Label>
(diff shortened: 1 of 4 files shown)


src/AccessibilityInsights.SharedUx/Controls/TestTabs/AutomatedChecksControl.xaml.cs
```diff
@@ -873,10 +874,6 @@ private void GroupItem_PreviewKeyDown(object sender, KeyEventArgs e)
                 {
                     exp.IsExpanded = true;
                 }
-                else
-                {
-                    urlLink?.Focus();
-                }
                 e.Handled = true;
             }
             else if (e.Key == Key.Left)
```
karanbirsingh, line 879: we might be able to remove HyperLink block on line 884

src/AccessibilityInsights.SharedUx/Controls/TestTabs/AutomatedChecksControl.xaml
```diff
@@ -273,11 +288,6 @@
                                                             <Label Margin="4,0,0,0" Padding="0" Style="{StaticResource lblLink}"
```
karanbirsingh, line 288: I believe we can remove this enclosing Label

src/AccessibilityInsights.SharedUx/Controls/TestTabs/AutomatedChecksControl.xaml.cs
DaveTryon: Yup. Removing.

src/AccessibilityInsights.SharedUx/Controls/TestTabs/AutomatedChecksControl.xaml
DaveTryon: Removed, as well as the string resource that was only used here

karanbirsingh (approved):

Issue #854, opened 2020-07-23
Checkbox border has insufficient contrast against its adjacent color in "Enable telemetry dialogue"


iamrafan:
**Describe the bug**
The border of the checkbox in "enable telemetry dialogue" fails Non-text Contrast requirement as it has insufficient contrast against its adjacent background

The following AI-windows mode(s) have this issue:
- Regular mode
- High Contrast White mode

**Expected behavior**
The border of the checkbox should have a contrast ratio of at least 3:1 against its adjacent color

**Actual behavior**
The non-text contrast ratio of the checkbox's border against its background fall short of 3:1

**Screenshots or .GIF**
#### Screenshot of the issue in High Contrast White mode
![1 1](https://user-images.githubusercontent.com/22068434/88341240-184be380-ccf2-11ea-8d04-12c9c3105ff1.png)

**Desktop**
 - OS: Windows 10 Enterprise, Version 1903
 - Accessibility Insights for Windows Version: Production Release: v 1.1.1248.1

DaveTryon:
Bug was originally reported against light mode, but I'm getting a contrast ratio of 4.88 in light mode, which satisfies the requirement. The contrast in HC White is 1.8, which fails the requirement. All modes are satisfied after the change.

DaveTryon:
@iamrafan, this fix will be in release 1.1.1370.01 (will release October 2nd). Please regress and close once the build is out.

iamrafan:
Validated the fix in Canary Release: v1.1.1380.01, the checkbox border is now compliant in all 6 modes.

PR #943, opened 2020-09-30
fix: Checkboxes fail color contrast in HC White

DaveTryon:
#### Describe the change
Default checkboxes are failing WCAG 2.1 color contrast requirements in High Contrast White mode. They're OK in other modes, but this change applies in all modes. These are the default colors from .NET, so I'll probably file an issue there, as well (they probably haven't yet picked up the latest WCAG requirements).

The workaround is to create a new style called `CheckBoxContrastingBorder` and use this as the base style for all checkboxes in the product. This includes:
- The first-boot telemetry permissions dialog
- The Welcome screen
- The configuration page for event listening
- The Automated checks results page
- The Text Range find dialog

I've sanity checked all of these and found no issues other than the checkbox is a little darker than it use to be in the light theme. Here's a screenshot of the first-boot telemetry dialog in all 6 modes--I've confirmed that all contrasts are greater than 3.0:

image

#### PR checklist

- [x] Does this address an existing issue? If yes, Issue# - #854 
- [styles only] Includes UI changes?


jalkire (approved):
Approve with suggestion

Issue #862, opened 2020-07-23
Icons on the top menu have insufficient contrast against their background on mouse hover


iamrafan:
The following AI-windows mode(s) have this issue:
- Dark mode
- High Contrast White mode
- High Contrast Black mode

**Screenshots or .GIF**
#### Screenshot of the issue in High Contrast White mode
![9 1](https://user-images.githubusercontent.com/22068434/88346786-03754d00-ccfe-11ea-95fb-1d4c5f4b5d72.png)

**Desktop**
 - OS: Windows 10 Enterprise, Version 1903
 - Accessibility Insights for Windows Version: Production Release: v 1.1.1248.1

DaveTryon:
@iamrafan, this is fixed in https://github.com/microsoft/accessibility-insights-windows/releases/tag/v1.1.1374.01. Please regress/close at your convenience. Thanks!

iamrafan:
Validated the fixes in Canary Release: v1.1.1380.01, all 6 modes are compliant

PR #948, opened 2020-10-05
fix: Change glyph color on button hover

DaveTryon:
#### Describe the change
We have several cases where we have button that wraps a FabricIconControl. Even if the FabricIconControl is hover-aware, there's a range of mouse positions where IsMouseOver for the button is true, but IsMouseOver for the FabricIconControl is false, which causes us to fail color contrast. This change allows a FabricIconControl to be hover-aware for its button parent, so that we always have sufficient contrast.

Note that this fixes only the contrast issue of glyph color to button color--there may also be an issue of highlighted button color to background color that will be tracked separately.

Before change:
862 (Old)

After change:
862 (New)

#### PR checklist

- [x] Does this address an existing issue? If yes, Issue# - #862 
- [x] Includes UI changes?

### DIFF
--- a/src/AccessibilityInsights/MainWindow.xaml
+++ b/src/AccessibilityInsights/MainWindow.xaml
@@ -345,12 +345,12 @@
                                                     <DataTrigger Binding="{Binding Path=vmHilighter.State, Mode=OneWay, RelativeSource={RelativeSource Mode=FindAncestor, AncestorType=local:MainWindow}}" Value="Off">
                                                         <Setter Property="GlyphName" Value="UnSetColor"/>
                                                     </DataTrigger>
-                                                    <Trigger Property="IsMouseOver" Value="True">
+                                                    <DataTrigger Binding="{Binding RelativeSource={RelativeSource AncestorType=Button,AncestorLevel=1}, Path=IsMouseOver}" Value="True">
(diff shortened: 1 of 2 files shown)


src/AccessibilityInsights/MainWindow.xaml
```diff
@@ -345,12 +345,12 @@
                                                     <DataTrigger Binding="{Binding Path=vmHilighter.State, Mode=OneWay, RelativeSource={RelativeSource Mode=FindAncestor, AncestorType=local:MainWindow}}" Value="Off">
                                                         <Setter Property="GlyphName" Value="UnSetColor"/>
                                                     </DataTrigger>
-                                                    <Trigger Property="IsMouseOver" Value="True">
+                                                    <DataTrigger Binding="{Binding RelativeSource={RelativeSource AncestorType=Button,AncestorLevel=1}, Path=IsMouseOver}" Value="True">
```
DaveTryon, line 348: @jalkire, I'd love to leverage the `hoverAwareFabricIconOnButtonParent` style in the controls that change their glyphs, but I'm only able to set one style. The highlighter control is used all over the place, so I'm hesitant to try to refactor it. I'd welcome any suggestions for a clean way to approach this :)

jalkire: Spoke offline with Dave. Don't know of a low-cost way to reuse the style, unfortunately ☹

jalkire (approved):

Issue #843, opened 2020-07-20
Dark mode: Text content has insufficient contrast against its background


iamrafan:
**Describe the bug**
"Automated checks" text in default state (when no app is scanned) and the fix message when no failures are detected, have insufficient contrast against their backgrounds.

**To Reproduce**
1. Make sure color setting for default app mode in windows is set to 'Dark'
2. Launch AI-Windows
3. Navigate to Tests section in the left menu bar without scanning any app
4. Observe the text "Automated checks" under fast pass has insufficient contrast against the background
5. Navigate back to live inspect and run tests on a section of an app that might not have any issues
6. Go to "Results in UI Automation Tree"
7. Observe that the message under "Fix the following" in the right pane has insufficient contrast against the background

**Expected behavior**
Text content to have contrast ratio of at least 4.5:1 against its background

**Actual behavior**
The contrast ratio for text falls short of 4.5:1

**Screenshots or .GIF**
![4 1](https://user-images.githubusercontent.com/22068434/87987123-f30d6a00-ca92-11ea-822d-23f349fee580.png)
![4 2](https://user-images.githubusercontent.com/22068434/87987143-fc96d200-ca92-11ea-8638-d78e2590607b.png)

**Desktop**
 - OS: Windows 10 Enterprise, Version 1903
 - Accessibility Insights for Windows Version: Production Release: v 1.1.1248.1

DaveTryon:
@iamrafan, this is fixed in release [1.1.1380.01](https://github.com/microsoft/accessibility-insights-windows/releases/tag/v1.1.1380.01). Please regress and close at your convenience.

iamrafan:
Validated the fix in Canary Release: v1.1.1380.01

PR #953, opened 2020-10-07
fix: Some dark mode text has insufficient contrast

DaveTryon:
#### Describe the change
Issue #843 calls out 2 places where colors are incorrect in dark mode. This fixes both.

The first problem is due to `SelectedFGTabBrush` being set to the wrong value, so we drew the text for the "Automated Checks tab" in the wrong color. This is the only place this value is used, so we can just change the color value without creating another brush. I discovered that HC Black and HC White modes are also broken, so I had to change brushes for both dark and HC modes. I've included screenshots of all 6 modes, just for completeness. In all cases, the old version is on the left, and the new version is on the right:

Mode | Screenshot
--- | ---
Light (no change) | image
Dark (fixed) | image
HC1 (no change) | image
HC2 (no change) | image
HC Black (fixed) | image
HC White (fixed) | image

The second problem is because we didn't style the control that reports "There were no failures", so it always displays in the default text color. This got weird in some HC modes, since the default mapped colors were inconsistent--in HC1 it mapped to the Text color but in HC2 it mapped to the Disabled Text color. After this change, it's always displayed in the same color as the "Fix the following" header. Again, I've included screenshots of all 6 modes. In all cases, the old version is on the left, and the new version is on the right:

Mode | Screenshot
--- | ---
Light (no change) | image
Dark (fixed) | image
HC1 (different) | image
HC2 (different) | image
HC Black (no change) | image
HC White (no change) | image

#### PR checklist

- [x] Does this address an existing issue? If yes, Issue# - #843 
- [colors only] Includes UI changes?

### DIFF
--- a/src/AccessibilityInsights.SharedUx/Controls/ScannerResultControl.xaml
+++ b/src/AccessibilityInsights.SharedUx/Controls/ScannerResultControl.xaml
@@ -177,7 +177,7 @@
                                 </Style>
                             </TextBox.Style>
                         </TextBox>
-                        <TextBox Text="{x:Static Properties:Resources.ScannerResultControl_noFailures}" AutomationProperties.LabeledBy="{Binding ElementName=labelFixFollowing}">
+                        <TextBox Text="{x:Static Properties:Resources.ScannerResultControl_noFailures}" AutomationProperties.LabeledBy="{Binding ElementName=labelFixFollowing}" Foreground="{DynamicResource ResourceKey=PrimaryFGBrush}">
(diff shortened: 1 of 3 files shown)


DaveTryon:
closing and reopening to get unstuck

dbjorge (approved):

jalkire (approved):

DaveTryon:
Closing and reopening...again

DaveTryon:
Closing and reopening...again

Issue #842, opened 2020-07-20
Dark mode: Text color of the links on mouse hover has insufficient contrast against its background


iamrafan:
**Describe the bug**
The text of the links in getting started dialog and in settings -> About page have insufficient contrast against their background when they are hovered using a mouse in dark mode

**To Reproduce**
1. Make sure color setting for default app mode in windows is set to 'Dark'
2. Launch AI-Windows
3. User Color Contrast Analyser tool test the contrast ratio of the link text color on hover against its background in "Getting started" dialog and in 'settings ->About' section
5. Observe that the contrast ratio falls short of the 4.5 : 1 required for text content

**Expected behavior**
The link text should have a contrast ratio of at least 4.5:1 against its background on mouse hover state

**Actual behavior**
The contrast ratio of the link text against its background is less than 4.5:1

**Screenshots or .GIF**
![3 1](https://user-images.githubusercontent.com/22068434/87986281-c573f100-ca91-11ea-962f-45396e6e1855.png)

![3 2](https://user-images.githubusercontent.com/22068434/87986292-c9077800-ca91-11ea-8305-989d3fd59e86.png)

![3 3](https://user-images.githubusercontent.com/22068434/87986298-cad13b80-ca91-11ea-8440-6c862740a554.png)

**Desktop**
 - OS: Windows 10 Enterprise, Version 1903
 - Accessibility Insights for Windows Version: Production Release: v 1.1.1248.1

DaveTryon:
I've reached out to design to get a new color for the link in the hover state. Once we have the color, it's trivial to add. One note: Hyperlink colors don't change in HC modes, so I'm proceeding on the assumption that there is no color contrast requirement between hover and non-hover colors, as long as both states satisfy the 4.5 contrast ratio against the background.

DaveTryon:
@iamrafan, this is fixed in release [1.1.1380.01](https://github.com/microsoft/accessibility-insights-windows/releases/tag/v1.1.1380.01). Please regress and close at your convenience.

iamrafan:
> I've reached out to design to get a new color for the link in the hover state. Once we have the color, it's trivial to add. One note: Hyperlink colors don't change in HC modes, so I'm proceeding on the assumption that there is no color contrast requirement between hover and non-hover colors, as long as both states satisfy the 4.5 contrast ratio against the background.

Yes, color contrast is not required across states, color belonging to each state having sufficient contrast against the background is enough

[...]

PR #956, opened 2020-10-08
fix: Update Link color on hover (dark mode)

DaveTryon:
#### Describe the change
As called out in #842, the default link hover color fails color contrast requirements in dark mode. This PR adds the ability to override the link text and changes the color in dark mode. The link color in other modes is unchanged. Here's a GIF of the new link color in action:

links-2

#### PR checklist

- [x] Does this address an existing issue? If yes, Issue# - #842


jalkire (approved):
When I hover over links in HC modes, their color doesn't change. That's the current prod behavior too, so not a concern with this PR. Do you know if that's the desired behavior?

DaveTryon:
> When I hover over links in HC modes, their color doesn't change. That's the current prod behavior too, so not a concern with this PR. Do you know if that's the desired behavior?

That's the same behavior I observe in the OS as well, so I think it's intentional.

Issue #828, opened 2020-07-01
Dark mode: Issue filing options' text fails contrast when the app is opened in light mode and then color setting is changed to dark mode


iamrafan:
**Describe the bug**
The text of the radio buttons does not switch to  appropriate color when the color setting is changed to dark mode after the app is opened in light mode

**To Reproduce**
1. Make sure color setting for default app mode is set to 'Light'
![image](https://user-images.githubusercontent.com/22068434/86277575-c77e1a80-bb8b-11ea-8029-d539ea7a081e.png)

2. Launch AI-Windows
3. Go to settings -> Connection
4. Go back to windows color settings and now change the default app mode to 'dark'
5. Observe the color of text adjacent to the radio buttons remains dark even though the background is changed to dark

**Expected behavior**
The text color should change to an appropriate color with sufficient contrast against the changed background

**Actual behavior**
The text color remains same as in regular(light) mode

**Screenshots or .GIF**
![dark-1](https://user-images.githubusercontent.com/22068434/86278484-52134980-bb8d-11ea-9ee7-ee65ec9b2c23.png)

**Desktop**
 - OS: Windows 10 Enterprise, Version 1903
 - Accessibility Insights for Windows Version: Production Release: v 1.1.1248.1

adarshrema:
Needs investigation by the PM and design teams

DaveTryon:
@iamrafan, this is fixed in the release at https://github.com/microsoft/accessibility-insights-windows/releases/tag/v1.1.1382.01. Please regress/close at your convenience.

iamrafan:
Validated the fix in Canary Release: v1.1.1383.01

PR #961, opened 2020-10-13
fix: Use a style instead of a fixed color in issue filing radio buttons

DaveTryon:
#### Describe the change
Radio buttons for issue filing are dynamically created with a color from the theme that is in effect when the controls are created. As such, they're not responsive to theme changes that occur afterward. The solution is to assign a style instead of a color, and then use a theme-responsive color within the style. Since I need to specify the style from code, I chose a name that explicitly names it as a style, which is something that we don't do for styles loaded from XAML.

This GIF only shows changing between dark and light, but I tested changing between HC and non-HC, as well as between HC modes. They all transition as expected.

#### PR checklist

- [x] Does this address an existing issue? If yes, Issue# - #828 

### DIFF
--- a/src/AccessibilityInsights.SharedUx/Resources/Styles.xaml
+++ b/src/AccessibilityInsights.SharedUx/Resources/Styles.xaml
@@ -1563,4 +1563,8 @@
             </Trigger>
         </Style.Triggers>
     </Style>
+    <Style TargetType="{x:Type RadioButton}" x:Key="primaryFGRadioButtonStyle" BasedOn="{StaticResource {x:Type RadioButton}}">
+        <!-- This style is set programmatically, so renaming it will require a change in the code-behind -->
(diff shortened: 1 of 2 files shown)


src/AccessibilityInsights.SharedUx/Resources/Styles.xaml
```diff
@@ -1563,4 +1563,8 @@
             </Trigger>
         </Style.Triggers>
     </Style>
+    <Style TargetType="{x:Type RadioButton}" x:Key="primaryFGRadioButtonStyle" BasedOn="{StaticResource {x:Type RadioButton}}">
```
DaveTryon, line 1566: There may be opportunities to leverage this style to replace explicitly setting the FG color in XAML on a per-control basis, but I'll look at that as a separate PR once this is merged.

karanbirsingh (approved):

Issue #868, opened 2020-07-24
Focus indicator on 'selected' test in Fast Pass has insufficient contrast against its background


iamrafan:
The following AI-windows mode(s) have this issue:
- High Contrast White mode
- High Contrast Black mode

**Screenshots or .GIF**
#### Screenshot of the issue in High Contrast White mode
![15 1](https://user-images.githubusercontent.com/22068434/88356099-8f49a200-cd1b-11ea-9852-af0946f498ea.png)

**Desktop**
 - OS: Windows 10 Enterprise, Version 1903
 - Accessibility Insights for Windows Version: Production Release: v 1.1.1248.1

iamrafan:
Validated the fixes in Canary Release: v1.1.1388.01. All 6 modes are now compliant

PR #969, opened 2020-10-14
fix: Improve contrast of test tab header focus indicator with inner border

jalkire:
#### Describe the change
Addresses Focus indicator on 'selected' test in Fast Pass has insufficient contrast against its background #868. Currently, the focus indicator on a test tab header is the normal black or white outline, but the color of the selected tab can be close to black or white. We had a similar problem with our left nav bar buttons, so I used the same solution here. Now, when a test tab has focus, it gets an inner border with the opposite (black or white) color of the current focus indicator. This means that there will always be a black and white border to indicate focus, which will pass contrast requirements. Also, this active rectangle on the left side of the selected tab now has a better color for high contrast modes.

Screenshots in Light, Dark, HC Black, HC White, HC 1, and HC 2:
image
image
image
image
image
image

#### PR checklist

- [x] Does this address an existing issue? If yes, Issue# - #868
- [x] Includes UI changes?


DaveTryon (approved):
HC White is still kind of hard to see, but that may be the best we can do with this palette.

Issue #879, opened 2020-07-24
Settings button icon in Events has insufficient contrast against its background in 'selected' state


iamrafan:
The following AI-windows mode(s) have this issue:
- High Contrast White mode
- High Contrast Black mode

**Screenshots or .GIF**
#### Screenshot of the issue in High Contrast White mode
![26 1](https://user-images.githubusercontent.com/22068434/88357551-85766d80-cd20-11ea-983c-7bdaa21170ad.png)

**Desktop**
 - OS: Windows 10 Enterprise, Version 1903
 - Accessibility Insights for Windows Version: Production Release: v 1.1.1248.1

DaveTryon:
@iamrafan, this is fixed in the release at https://github.com/microsoft/accessibility-insights-windows/releases/tag/v1.1.1384.01. Please validate/close at your convenience.

iamrafan:
Validated the fix in Canary Release: v1.1.1388.01. All 6 modes are now compliant

PR #970, opened 2020-10-15
fix issues with FabricIconControls on buttons

DaveTryon:
#### Describe the change
This change fixes 3 types of issues (one reported, two not) associated with hover buttons with fabric icons, some of which also have pop-up context menus. The behavior of fixing them separately is bad (controls become invisible), so they're all in a single PR. I scanned the code for all instances and included them here.

Problem 1 (Reported): Buttons with pop-up context menus fail contrast requirements in HC White and HC Black modes once the user clicks on the button (the foreground color of the FabricIconControl doesn't get updated). This was reported in the events dialog, but it happens in several places. Root cause is that the foreground color of the FabricIconControl is never getting updated from its default color. Here's one example of the before and after using HC White:

Before | After
--- | ---
image | image

Problem 2 (Unreported): Once you've activated a context menu from a button containing a FabricIconControl, the control no longer responds to mouse hover. Root cause is that the button's dynamic styling is getting overwritten by a fixed color after the button is released. Here's the before and after using HC White--I included the welcome screen just to show the beginning of the loop

Before | After
--- | ---
icons-nohover | icons-hover

Problem 3 (Unreported): Several button icons fail contrast requirements on hover because they contain FabricIconControls that don't change color on mouse hover. The fix is simple--just use a style that knows about mouse hover of the parent button. Here's a sample from the color contrast analyzer, again in HC White:

Before | After
--- | ---
hover-fail | hover-succeed

#### PR checklist

- [x] Does this address an existing issue? If yes, Issue# - #879 
- [ ] Includes UI changes?

### DIFF
--- a/src/AccessibilityInsights.SharedUx/Behaviors/DropDownButtonBehavior.cs
+++ b/src/AccessibilityInsights.SharedUx/Behaviors/DropDownButtonBehavior.cs
@@ -1,6 +1,11 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using AccessibilityInsights.CommonUxComponents.Controls;
+using AccessibilityInsights.SharedUx.Utilities;
 using Microsoft.Xaml.Behaviors;
+using System;
+using System.Collections.Generic;
+using System.Linq;
 using System.Windows;
 using System.Windows.Controls;
 using System.Windows.Controls.Primitives;
@@ -15,11 +20,8 @@ namespace AccessibilityInsights.SharedUx.Behaviors
     public class DropDownButtonBehavior : Behavior<Button>
     {
         private bool isContextMenuOpen;
-
-        /// <summary>
-        /// Stores the button's background so that it can be restored later
-        /// </summary>
-        private Brush origBackground;
+        private IReadOnlyList<FabricIconControl> _containedFabricIconControls;
+        private Style _savedFabricIconControlStyle;
 
         /// <summary>
         /// Attach to necessary event handlers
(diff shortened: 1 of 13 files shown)


src/AccessibilityInsights.SharedUx/Behaviors/DropDownButtonBehavior.cs
```diff
@@ -90,7 +113,9 @@ protected override void OnDetaching()
         /// <param name="e"></param>
         void ContextMenu_Closed(object sender, RoutedEventArgs e)
         {
-            AssociatedObject.Background = origBackground;
```
DaveTryon, line 93: This was the cause of Problem 2 because it overrode the styling. Calling ClearValue restores the Background to its "default" state, which preserves the styling.

src/AccessibilityInsights.SharedUx/Behaviors/DropDownButtonBehavior.cs
```diff
@@ -52,17 +43,49 @@ private void AssocisatedObject_ContextMenuOpening(object sender, ContextMenuEven
...
+
+        private void StyleContainedFabricIconControls()
+        {
+            _containedFabricIconControls = ContainedControlFinder<FabricIconControl>.Find(AssociatedObject).ToList();
+
+            if (_containedFabricIconControls.Any())
+            {
+                // This assumes that all contained FabricIconControls use the same style
+                _savedFabricIconControlStyle = _containedFabricIconControls[0].Style;
+                Style pressedStyle = AssociatedObject.FindResource("fabricIconOnPressedButtonParentStyle") as Style;
```
DaveTryon, line 65: I tried various combinations of setting the Foreground on the FabricIconControl, but couldn't get it to work. I finally went with replacing the Style. This will have side effects if the FabricIconControl uses a style for anything besides colors

src/AccessibilityInsights.SharedUx/Utilities/ContainedControlFinder.cs
```diff
@@ -0,0 +1,31 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Collections;
+using System.Collections.Generic;
+using System.Windows;
+
+namespace AccessibilityInsights.SharedUx.Utilities
```
DaveTryon, line 8: I pulled this out and used a template, just in case we need the functionality elsewhere. It's internal but could easily be public if we wanted it to be

jalkire (approved):

Issue #872, opened 2020-07-24
Action buttons in patterns do not have any visual indication that they are actionable items


iamrafan:
The following AI-windows mode(s) have this issue:
- High Contrast White mode
- High Contrast Black mode
- High Contrast ```#1```
- High Contrast ```#2```

**Screenshots or .GIF**
#### Screenshot of the issue in High Contrast White mode
![19 1](https://user-images.githubusercontent.com/22068434/88356701-91acfb80-cd1d-11ea-8526-06bcd416b7c0.png)

**Desktop**
 - OS: Windows 10 Enterprise, Version 1903
 - Accessibility Insights for Windows Version: Production Release: v 1.1.1248.1

DaveTryon:
@iamrafan, this is fixed in the release found at https://github.com/Microsoft/accessibility-insights-windows/releases. Please validate and close at your convenience. Thanks!

iamrafan:
Validated the fixes in Canary Release: v1.1.1388.01. All 6 modes are now compliant

PR #972, opened 2020-10-16
fix: Make Action buttons more obvious in HC modes

DaveTryon:
#### Describe the change
Issue #872 points out that the patterns actions buttons have no visual indication in HC modes that they're anything more than generic text. This PR adds a border in HC modes. The eagle-eyed observer might note that it also makes the buttons in non-HC modes a tiny bit taller and rounds the corners, both of which look better to my (potentially biased) eye. Happy to discuss alternatives here.

One potential oddity is that I'm using the button foreground color as the button's border color. That's to ensure that the button outline is still visible on mouse hover when the pattern is selected in the UI. You can observe this in the following GIFs (the `InvokePattern` is selected, the `LegacyIAccessiblePattern` is not):

Mode | GIF
--- | ---
Light | 872-light
Dark | 872-dark
HC 1 | 872-hc1
HC 2 | 872-hc2
HC Black | 872-black
HC White | 872-white

#### PR checklist

- [x] Does this address an existing issue? If yes, Issue# - #872 
- [visual only] Includes UI changes?

### DIFF
--- a/src/AccessibilityInsights.SharedUx/Controls/PatternInfoControl.xaml
+++ b/src/AccessibilityInsights.SharedUx/Controls/PatternInfoControl.xaml
@@ -45,8 +45,8 @@
                                 VerticalContentAlignment="Center" VerticalAlignment="Center"
                                 Visibility="{Binding Path=ActionVisibility, BindsDirectlyToSource=True}"
                                 Content="{Binding Path=ActionName, BindsDirectlyToSource=True}"
-                                IsEnabled="True" Padding="4,0" FontWeight="Normal"
-                                Height="18" IsTabStop="False"
+                                IsEnabled="True" FontWeight="Normal"
+                                IsTabStop="False"
                                 Margin="5,0,0,0">
                         </Button>
                     </StackPanel>
--- a/src/AccessibilityInsights.SharedUx/Resources/Styles.xaml
+++ b/src/AccessibilityInsights.SharedUx/Resources/Styles.xaml
@@ -113,7 +113,7 @@
         <Setter Property="Template">
             <Setter.Value>
                 <ControlTemplate TargetType="{x:Type Button}">
-                    <Border Background="{TemplateBinding Background}">
+                    <Border Background="{TemplateBinding Background}" BorderBrush="{TemplateBinding Foreground}" BorderThickness="{DynamicResource ResourceKey=ActionsButtonBorderThickness}" CornerRadius="3" Padding="4,1">
                         <ContentPresenter HorizontalAlignment="Center" VerticalAlignment="Center"/>
                     </Border>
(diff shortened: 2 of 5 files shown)


src/AccessibilityInsights.SharedUx/Resources/Styles.xaml
```diff
@@ -113,7 +113,7 @@
         <Setter Property="Template">
             <Setter.Value>
                 <ControlTemplate TargetType="{x:Type Button}">
-                    <Border Background="{TemplateBinding Background}">
+                    <Border Background="{TemplateBinding Background}" BorderBrush="{TemplateBinding Foreground}" BorderThickness="{DynamicResource ResourceKey=ActionsButtonBorderThickness}" CornerRadius="3" Padding="4,1">
```
DaveTryon, line 116: The template will override any value set in the xaml file, so I moved the padding here to keep related things together.

src/AccessibilityInsights.SharedUx/Controls/PatternInfoControl.xaml
```diff
@@ -45,8 +45,8 @@
                                 VerticalContentAlignment="Center" VerticalAlignment="Center"
                                 Visibility="{Binding Path=ActionVisibility, BindsDirectlyToSource=True}"
                                 Content="{Binding Path=ActionName, BindsDirectlyToSource=True}"
-                                IsEnabled="True" Padding="4,0" FontWeight="Normal"
-                                Height="18" IsTabStop="False"
```
DaveTryon, line 49: The height needs to be dynamic to account for the variable border thickness

src/AccessibilityInsights.SharedUx/Resources/Styles.xaml
jalkire: Looking at this, the thought crosses my mind: is there something important the BtnActions style is doing with the ControlTemplate that BtnStandard isn't? BtnStandard​​​​​​​ already provides a border in HC modes. Are those new IsMouseOver colors importantly different vs BtnStandard's?

jalkire (approved):
Approved with a question

src/AccessibilityInsights.SharedUx/Resources/Styles.xaml
DaveTryon: The brush colors differ only in dark mode. There, BtnActions needs to have a lighter IsMouseOver color than BtnStandard because BtnActions is sometimes displayed against a highlighted background. I tried using the colors from BtnStandard and the change on mouse hover was almost imperceptible. I'll proceed with this change, but will investigate the possibility of using the lighter IsMouseOver color for BtnStandard. We'd still need the styling inside the ControlTemplate, but we might be able to consolidate the brushes.

Issue #866, opened 2020-07-24
Focus indicator on tabs in events has insufficient contrast against its background in 'selected' state 


iamrafan:
The following AI-windows mode(s) have this issue:
- High Contrast White mode

**Screenshots or .GIF**
#### Screenshot of the issue in High Contrast White mode
![13 1](https://user-images.githubusercontent.com/22068434/88355556-e3ec1d80-cd19-11ea-9c9a-7c471952125c.png)

**Desktop**
 - OS: Windows 10 Enterprise, Version 1903
 - Accessibility Insights for Windows Version: Production Release: v 1.1.1248.1

DaveTryon:
@iamrafan, this is fixed (by adding a contrasting border inside the focus rectangle) in the release found at https://github.com/microsoft/accessibility-insights-windows/releases/tag/v1.1.1390.01. Please validate/close at your convenience. Thanks!

iamrafan:
Validated the fix in Canary Release: v1.1.1398.01. All 6 modes are now compliant

PR #977, opened 2020-10-21
fix(HC): Add a contrasting border around tab icons when they're focused

DaveTryon:
#### Describe the change
As reported in #866, the focus rectangle on the tabs in the inspect view is really hard to see in HC modes, especially HC White. This PR changes the styling of the tabs so that they draw a border inside the focus rectangle. The color is chosen to match the tab background color in Light and Dark modes, but to contrast against it in HC modes. This makes the focus rectangle stand out a lot more clearly in HC modes.

There's on minor side effect--when the focus moves to the tab, the border causes a very small shift to the text in the tab, so there's a small visual "jump" that isn't there in the current code. I mention this only to call out that it's there, but I'm not proposing that we try to fix it--at least not unless we get complaints about it.

Here are the GIF's showing before and after in all 6 modes:
Mode | Before | After
--- | --- | ---
Light | Current-light | New-light
Dark | Current-dark | New-dark
HC 1 | Current-hc-1 | New-hc-1
HC 2 | Current-hc-2 | New-hc-2
HC Black | Current-hc-black | New-hc-black
HC White | Current-hc-white | New-hc-white

#### PR checklist

- [x] Does this address an existing issue? If yes, Issue# - #866 

### DIFF
--- a/src/AccessibilityInsights.SharedUx/Resources/Styles.xaml
+++ b/src/AccessibilityInsights.SharedUx/Resources/Styles.xaml
@@ -861,7 +861,7 @@
         <Setter Property="Template">
             <Setter.Value>
                 <ControlTemplate TargetType="{x:Type TabItem}">
-                    <Border BorderThickness="0,0,1,0" BorderBrush="White" Margin="0,-2,0,0">
+                    <Border x:Name="Overlay" BorderThickness="0,0,1,0" BorderBrush="White" Margin="0,-2,0,0">
                         <Grid x:Name="gdTab">
                             <ContentPresenter x:Name="ContentSite"
                                         VerticalAlignment="Center"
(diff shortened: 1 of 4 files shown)


src/AccessibilityInsights.SharedUx/Resources/Styles.xaml
```diff
@@ -886,6 +886,11 @@
                             <Setter TargetName="gdTab" Property="Background" Value="{DynamicResource ResourceKey=PrimaryBGBrush}"/>
                             <Setter TargetName="gdTab" Property="Background" Value="{DynamicResource ResourceKey=SelectedBrush}"/>
                         </MultiTrigger>
+                        <Trigger Property="IsFocused" Value="True">
+                            <Setter TargetName="Overlay" Property="BorderThickness" Value="3"/>
```
jalkire, line 890: Since the contrast issue only showed up in high contrast modes, you could use `BtnBrdrThickness` from Brushes.xaml (or follow a similar pattern) for the thickness here. This would avoid the "jumping" effect in non-HC modes. Might be more complicated than it's worth, though

jalkire (approved):
Approve with a suggestion

src/AccessibilityInsights.SharedUx/Resources/Styles.xaml
DaveTryon: I tried this, but there was still a jump, I think due to the margin still being set. I'm going to leave this as is for now, and we have the option to revisit at another time. I made one change, though, which was to use SystemColors.WindowColor as the border, so we also get a contrasting border (as opposed to just a thicker border) in HC 1 and HC 2. I've updated the GIFs accordingly

Issue #731, opened 2020-03-19
[Feature Request] Consider tighter MSI validation


DaveTryon:
VersionSwitcher currently checks only that the new MSI has a valid and non-revoked digital signature, and disregards _who_ signed the MSI. We might want to tighten this check to ensure that VersionSwitcher only installs _our_ MSI's. Here is a non-exhaustive list of options:

1. Validate the canonical name of the MSI's signing certificate against a collection of names built into the VersionSwitcher. Refuse to install if validation fails.
2. Validate the thumbprint of the MSI's signing certificate against a collection of thumbprints built into the VersionSwitcher. Refuse to install if validation fails.
3. Validate the canonical name or thumbprint of the MSI's signing certificate against data that gets added to the web-hosted metadata file. Pass the validation as a command line parameter into VersionSwitcher, Refuse to install if validation fails. 
4. Validate a hash of the MSI file against data that gets added to the web-hosted metadata file. Pass the validation as a command line parameter into VersionSwithcer. Refuse to install if validation fails.

sauravbhattacharya001:
There is some code in trust verifier: IsFileTrusted(string filePath). Can a similar be used in VersionSwitcher to help with this?

sauravbhattacharya001:
where can I find the canonical names and thumbprints?

DaveTryon:
@sauravbhattacharya001, the various options were out there to reflect the fact that there was some uncertainty about our long-term signing strategy. That uncertainty has since been worked out, and our plan for the foreseeable future is to use a certificate with the canonical name of **Microsoft Corporation** (that's the certificate used to sign the MSI files published at  https://github.com/Microsoft/accessibility-insights-windows/releases). With that decision in place, we'd probably opt for the first (and simplest) option listed above, structuring the code to use any match from a hardcoded list of canonical names, then only putting one item in the list--that makes it simple to support different canonical names should future decisions require it. We prefer the canonical name to a thumbprint because it represents less churn as certificates are renewed.

[...]

PR #982, opened 2020-10-28
fix: Validate MSI signer before installing it

DaveTryon:
#### Describe the change
Issue #731 came from a previous security review and was intended to prevent us from launching an MSI that we weren't sure that we trusted. We already verify that the MSI is authenticode signed and that the signing chain is valid. This change adds a check based on the organization that issued the certificate used to sign the MSI. I've confirmed that the same issuer is reported in current Production and Canary builds. 

This adds no unit tests to avoid adding binary files to the repo. We could definitely change that, if desired. 

This approach is future-proofed against _anticipated_ changes to the issuer of the signing certificate. Should that happen, we'd just add another entry in `TrustedCertIssuerEndings` some time before the change, and the transition from old to new will be seamless. We risk problems, however, if there are _unanticipated_ changes in the name of the organization of the issuing cert. That would only occur if there were major changes in the company signing policy, and we'd probably know about those in advance, so it seems like a reasonable risk--it also aligns with the guidance from the team than manages certs, and with how we handle this in the Unified client.

#### PR checklist

- [x] Does this address an existing issue? If yes, Issue# - #731 
- [n/a] Includes UI changes?

### DIFF
--- a/src/AccessibilityInsights.SetupLibrary/TrustVerifier.cs
+++ b/src/AccessibilityInsights.SetupLibrary/TrustVerifier.cs
@@ -57,9 +57,9 @@ namespace AccessibilityInsights.SetupLibrary
             using (X509Certificate cert = X509Certificate.CreateFromSignedFile(filePath))
             {
                 string issuer = cert.Issuer;
-                foreach (string trustecCertIssuerEnding in TrustedCertIssuerEndings)
+                foreach (string trustedCertIssuerEnding in TrustedCertIssuerEndings)
                 {
-                    if (issuer.EndsWith(trustecCertIssuerEnding, StringComparison.Ordinal))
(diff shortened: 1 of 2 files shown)


jalkire:
For reference, our electron application checks that the "Issued to" field matches "Microsoft Corporation" on windows. Specifically, it [looks like we're checking](https://github.com/develar/electron-builder/blob/aa3625d5f6b719b6a7634c4c7ba0aad1ffe7d3d6/packages/app-builder-lib/src/winPackager.ts#L103) the CN part of the Subject. If we're worried about that field changing here, we'll want to consider the same risk on the electron side as well

DaveTryon:
@jalkire, you commented as I was making the vary change that you described :)

karanbirsingh (approved):

PR #986, opened 2020-10-29
fix: Update visuals in Hot key grabber dialog

DaveTryon:
#### Describe the change
The HotkeyGrabDialog has a few issues:
- It always uses the default background color, which is a problem in dark mode, since it's putting white text on a white background
- It has no border, which is confusing on a busy screen
- The TextBlock that displays the shortcut gets truncated in HC modes

Changes:
- Specify a palette-keyed background color, border color, and a border thickness (even in non-HC modes)
- Make the TextBlock taller (even in non-HC modes)

Mode | Old behavior | New behavior
--- | --- | ---
Light | image | image
Dark | image | image
HC 1 | image | image
HC 2 | image | image
HC Black | image | image
HC White | image | image

#### PR checklist

- [styles only] Includes UI changes?


jalkire (approved):
It could be worth looking at the `ContainedDialog` approach for displaying this dialog--the thick border added here looks very different than the other similar dialogs we have (`TelemetryApproveContainedDialog`, `ChangeChannelContainedDialog`, `StartupModeControl`). This PR is still a huge improvement over the existing behavior so I'm approving, but it feels a little weird to introduce new, unique styling when we follow a different UI pattern for similar UI components

DaveTryon:
@adarshrema, I agree that this would be better using the ContainedDialog, but it's a larger change than I want to pick up to address this issue. Please keep this on your radar as a potential improvement (bug-level, not feature-level)

PR #1002, opened 2020-12-07
Chore: Switch from using Microsoft.CodeAnalysis.FxCopAnalyzers to Microsoft.CodeAnalysis.NetAnalyzers

RobGallo:
#### Describe the change

Based on recommendation from https://docs.microsoft.com/en-us/visualstudio/code-quality/migrate-from-fxcop-analyzers-to-net-analyzers?view=vs-2019

Also removed some disabled warnings which appear to no longer be necessary after the change.

### DIFF
--- a/build/NetFrameworkRelease.targets
+++ b/build/NetFrameworkRelease.targets
@@ -3,7 +3,7 @@
   <PropertyGroup>
     <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
 	<NoWarn>RS0016</NoWarn>
-    <!-- used by Microsoft.CodeAnalysis.NetAnalyzers -->
+      <!-- used by Microsoft.CodeAnalysis.NetAnalyzers -->
       <AnalysisMode>AllEnabledByDefault</AnalysisMode>
   </PropertyGroup>
 
--- a/build/NetFrameworkRelease.targets
+++ b/build/NetFrameworkRelease.targets
@@ -3,6 +3,8 @@
   <PropertyGroup>
     <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
 	<NoWarn>RS0016</NoWarn>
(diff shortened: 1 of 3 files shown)


dbjorge (changes requested):
@RobGallo The [migration guide you linked](https://docs.microsoft.com/en-us/visualstudio/code-quality/migrate-from-fxcop-analyzers-to-net-analyzers?view=vs-2019) notes this recommendation after replacing the package reference:

> 3. Enable additional rules: `Microsoft.CodeAnalysis.NetAnalyzers` is much more conservative compared to `Microsoft.CodeAnalysis.FxCopAnalyzers`. Unlike FxCopAnalyzers package, it only has a few correctness rules which are enabled by default as build warnings. You can enable additional rules by customizing the AnalysisMode MSBuild property. For example, setting the property to `AllEnabledByDefault` will enable all the applicable CA rules as build warnings by default.

Should we be enabling rules explicitly as part of this migration, like the guide suggests?

build/NetFrameworkRelease.targets
```diff
@@ -3,6 +3,8 @@
   <PropertyGroup>
     <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
 	<NoWarn>RS0016</NoWarn>
+    <!-- used by Microsoft.CodeAnalysis.NetAnalyzers -->
```
DaveTryon, line 6:
```suggestion
      <!-- used by Microsoft.CodeAnalysis.NetAnalyzers -->
```

DaveTryon (approved):

RobGallo:
@dbjorge 
Are you suggesting enabling individual rules rather than all by default? I've already enabled all by default.

dbjorge (approved):
@RobGallo My mistake! I missed that line the first time through the PR

PR #1023, opened 2021-01-11
fix: Remove spurious failure message in unit tests

DaveTryon:
#### Describe the change
Our PR tests currently produce the following exception on every execution:

```
Class Cleanup method ConfigurationModelTests.ClassCleanup failed. Error Message: System.IO.DirectoryNotFoundException: Could not find a part of the path 'D:\a\_temp\TestResults\Deploy_VssAdministrator 2021-01-11 13_22_45\user'.. Stack Trace:     at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
   at System.IO.Directory.Delete(String fullPath, String userPath, Boolean recursive, Boolean checkHost)
   at System.IO.Directory.Delete(String path)
   at AccessibilityInsights.SharedUxTests.Settings.ConfigurationModelTests.ClassCleanup() in D:\a\1\s\src\AccessibilityInsights.SharedUxTests\Settings\ConfigurationModelTests.cs:line 37
```
This isn't a fatal error, but it's a distraction when trying to find and fix real issues. The exception occurs because we try to delete the folder set as `ConfigurationFolderPath`, but we never created it. The minimum solution would be to just remove the line that deletes that folder, but it looks odd and could be prone to breakage if we later change the code or the test.

Stepping back, I wondered why we delete the directory at all, and the only reason I could find was to prevent potential interference with other unit tests. This PR solves that same problem in a different way: create a new folder on each iteration of the test. The `context.TestRunDirectory` is already unique per test run, and `context.TestName` provides the class name. The net result is that file artifacts created by this class will have no impact on other tests, whether in the same or in different test runs.

#### PR checklist

- [n/a] Includes UI changes?

### DIFF
--- a/src/AccessibilityInsights.SharedUxTests/Settings/ConfigurationModelTests.cs
+++ b/src/AccessibilityInsights.SharedUxTests/Settings/ConfigurationModelTests.cs
@@ -21,23 +21,15 @@ namespace AccessibilityInsights.SharedUxTests.Settings
     {
         private static FixedConfigSettingsProvider testProvider;
 
-        private static string TestBaseDirectory;
-
         [ClassInitialize()]
         public static void ClassInit(TestContext context)
         {
-            TestBaseDirectory = Path.Combine(context.TestRunDirectory, "ConfigurationModelTests");
+            string testBaseDirectory = Path.Combine(context.TestRunDirectory, "ConfigurationModelTests");
 
-            var configDir = Path.Combine(TestBaseDirectory, "config");
-            var userDir = Path.Combine(TestBaseDirectory, "user");
+            var configDir = Path.Combine(testBaseDirectory, "config");
+            var userDir = Path.Combine(testBaseDirectory, "user");
             testProvider = new FixedConfigSettingsProvider(configDir, userDir);
             Directory.CreateDirectory(testProvider.ConfigurationFolderPath);
-        }
-
-        [ClassCleanup()]
-        public static void ClassCleanup()
-        {
-            Directory.Delete(TestBaseDirectory);
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


karanbirsingh (approved):

src/AccessibilityInsights.SharedUxTests/Settings/ConfigurationModelTests.cs
```diff
@@ -24,19 +24,14 @@ public class ConfigurationModelTests
         [ClassInitialize()]
         public static void ClassInit(TestContext context)
         {
-            var configDir = Path.Combine(context.TestRunDirectory, "config");
-            var userDir = Path.Combine(context.TestRunDirectory, "user");
+            string testBaseDirectory = Path.Combine(context.TestRunDirectory, context.TestName);
```
peterdur, line 27: `context.TestName` will be the name of the first test method executed in the class, which in this case happens to be `ConfigurationModelTest`. Is that expected? If not, could use `context.FullyQualifiedTestClassName` instead, though it's a bit verbose (here it would be `"AccessibilityInsights.SharedUxTests.Settings.ConfigurationModelTests"`)

DaveTryon: Good point. The entire path needs to not exceed MAX_PATH (defined as 260) characters. Using FullyQualifedTestClassName in this case means that we have about 127 characters after the root path of the repo, plus about 15 characters for file names. That means that we'll be find as long as the root repo path doesn't exceed 120 characters. That should be plenty, and we can use a hardcoded value if/when it becomes a problem. I didn't want to use a hardcoded value at first, since I wanted a pattern that we could copy elsewhere and it would "just work".

I've made the change to use FullyQualifiedTestClassName

Issue #1007, opened 2020-12-17
[BUG] AI-Win does not clear user config on uninstall


jalkire:
**Describe the bug**
When AI-Win is uninstalled, its user configuration is not cleared. If someone opts in (or out) of telemetry, uninstalls, and then reinstalls the software, they are thus not asked to opt in to telemetry a second time.

**To Reproduce**
1. Install AI-Win
2. Opt in or out of telemetry
3. Uninstall AI-Win
4. Reinstall AI-Win
5. Launch AI-Win

**Expected behavior**
The user sees the telemetry opt-in dialog

**Actual behavior**
The user does not see the telemetry opt-in dialog (but their previous selection is respected)

**Additional context**
Reported by @LiLoDavis

DaveTryon:
@adarshrema, retaining the settings across uninstall is both intentional and required for upgrade and channel switching scenarios. If we opt to delete the config file during uninstall, we'll need to add some code to VersionSwitcher so that it stashes the config file before uninstalling the installed version and restores it (updating the release channel if changing channels) after installing the new version (but before launching the new version)

DaveTryon:
@adarshrema, should this be reclassified as a feature request?

adarshrema:
Keeping this open to investigate the scope of the issue.

ferBonnin:
@adarshrema @DaveTryon if we cannot do a clear user config on uninstall, we should document in our website how to completely delete the app and config (similar to how we do for [Accessibility Insights for Android - Uninstall](https://accessibilityinsights.io/docs/en/android/getstarted/setup/#uninstalling-accessibility-insights-for-android))

adarshrema:
Fix to delete user config on explicit uninstall scenario and save/ restore it for upgrade and version switch scenarios.

ferBonnin:
marking this as resolved and assigning to @LiLoDavis

DaveTryon:
@LiLoDavis, I'll add a link once this is in the Canary pipeline

DaveTryon:
The fix ran into issues in the signed build, so I had to back it out. Moving this back to active ☹️

DaveTryon:
[...]

PR #1038, opened 2021-01-29
fix: Use a custom action to delete user config files on uninstall

DaveTryon:
#### Describe the change

This PR is based on #1029, revised to successfully build in the SignedBuild environment, where the WiX toolset is not installed. We import the WiX components from NuGet, then use those components to build the package (AccessibilityInsights.CustomActions.CA.dll) that gets consumed by the MSI. Validated using the MSI published from https://dev.azure.com/mseng/1ES/_build/results?buildId=14224050&view=results.

##### Notes from #1029 (repeated for convenience)
For issue #1007, we've chosen to use a custom action per the following spec:
1. It is invoked only on uninstall
2. When invoked, it checks to see if a process named AccessibilityInsights.VersionSwitcher.exe is currently running
3. If AccessibilityInsights.VersionSwitcher.exe is not running, the custom action deletes all files in the configuration directory (ignoring subdirectories)

Ran full E2E tests using actual MSI files in the following scenarios (config files exist at the beginning of each scenario):
Scenario | Result
--- | ---
Install | Config files are preserved
Upgrade by manually running MSI (no VersionSwitcher) | Config files are preserved
Uninstall MSI while VersionSwitcher is running | Config files are preserved
Uninstall MSI while VersionSwitcher is not running | Config files are deleted

##### Signing implementation notes
1. The "typical" way to create a package is to compile both the assembly and the package in the same project. This didn't play nicely with signing, so it's split into 2 projects. `CustomActions.csproj` compiles and signs AccessibilityInsights.CustomActions.dll, and then `CustomActions.Package.csproj` builds the signed assemblies into a package. The latter project is what actually imports the WiX assemblies, so we build the package in all configurations, even though we only build the MSI in release. The updates to the dotnet restore command is to exclude `CustomActions.Package.csproj`, since WiX and dotnet restore don't play nicely together.
2. `CustomActions.Package.csproj` builds a placeholder assembly, then we inject the `OverwriteWithSignedAssemblies` target before doing the packaging. This target copies the signed versions of AccessibilityInsights.CustomActions.dll, Newtonsoft.Json.dll, and Microsoft.Deployment.WindowsInstaller.dll to the intermediate directory, so that the signed versions of the files end up in the package.
3. The package (AccessibilityInsights.CustomActions.CA.dll) is intentionally not signed. I had one build that signed the package with the same cert that we use for the assemblies that we build, and one build that signed the package with the cert that we use for 3rd party assemblies. ESRP ran into problems with both options, presumably due to the way that the package is generated (there's a WiX tool that synthesizes the DLL from other DLLs, so it's kind of a one-off). The packages gets immediately wrapped into the MSI, which is independently signed, so there's no way to tamper with the package without tampering with the MSI. We'll bring this up in our next security review, but for the moment there seems to be no way around this limitation.

Finally, this PR includes **a lot** of commits--that's an artifact of the fact that each iteration of the build pipeline required a commit and push. As such, there were several commits that didn't work and later got pulled out. I've created #1039 as a reference with just 2 commits--the first restores what we had with #1029, and the second combines all of the changes that went into getting signing working.

#### PR checklist

- [x] Does this address an existing issue? If yes, Issue# - #1007 
- [n/a] Includes UI changes?


src/AccessibilityInsights.CustomActions.Package/CustomAction.config
```diff
@@ -0,0 +1,32 @@
...
+
+          Note for .NET Framework v3.0 and v3.5, the runtime version is still v2.0.
+
+          In order to enable .NET Framework version 2.0 runtime activation policy, which is to load all assemblies
+          by using the latest supported runtime, @useLegacyV2RuntimeActivationPolicy="true".
+
+          For more information, see http://msdn.microsoft.com/en-us/library/bbx34a2h.aspx
+        -->
+
+        <supportedRuntime version="v4.0" />
```
DaveTryon, line 22: This file exists to set the runtime version, otherwise the CustomActions will fail to load

karanbirsingh (approved):

PR #1052, opened 2021-03-02
fix: Replace tab stops help icon with text

jalkire:
#### Details
Currently, tab stops has a help (i) icon that is a hyperlink. It does not show a tooltip on keyboard focus. This PR replaces that icon with text to avoid the need for a tooltip.

##### Motivation
Addresses an accessibility issue

##### Context
I tried to adapt `KeyboardToolTipButtonBehavior` to function on hyperlinks. Unfortunately, `Hyperlink` is not a `FrameworkElement` and thus cannot be set as a `ToolTip`'s `PlacementTarget` and does not prove screen location information. This makes it very difficult to position a keyboard focus triggered `ToolTip` on the screen for a `Hyperlink`. As such, this PR's approach was selected instead.

Screenshots, before on top, after below:
Dark Mode:
image
HC White:
image
HC Black:
image

#### Pull request checklist

- [x] Includes UI changes?


DaveTryon (approved):
Changes look good. It's a bit unfortunate that NVDA announces the tab stops as "Tab Stops, Learn more about tab stops, Learn more about tab stops link", which is more verbose than the old version of "Tab Stops, Guidance link", but it's consistent with how we handle other links and seems like the right thing to expose.

PR #1054, opened 2021-03-03
fix: Suppress warnings that VS 16.9.0 is suddenly flagging

DaveTryon:
#### Details

VisualStudio 16.9.0 recently released, and builds with the new version are suddenly broken due to 18 new warnings that were not occurring before. We've reached out to the Visual Studio team to understand why VS suddenly decided to raise these warnings, but the quickest mitigation to get everyone unstuck is to suppress these warnings globally. 

##### Motivation

Allow the team to develop with VS 16.9.0

##### Context

This is a temporary mitigation to get unstuck. It comes with the understanding that we'll actually evaluate and maybe fix these issues going forward.

#### Pull request checklist

- [n/a] Run through of all [test scenarios](https://github.com/Microsoft/accessibility-insights-windows/blob/master/docs/Scenarios.md) completed?
- [n/a] Does this address an existing issue? If yes, Issue# - 
- [n/a] Includes UI changes?
  - [n/a] Run the production version of Accessibility Insights for Windows against a version with changes.
  - [n/a] Attach any screenshots / GIF's that are applicable.

> Note: After the PR has been created, certain checks will be kicked off. All of these checks must pass before a merge.

PR #1061, opened 2021-03-04
fix: Address code that triggers CA1508 warnings

DaveTryon:
#### Details

In #1054, we suppressed CA1508 warnings, which deal with redundant code. Looking through these, some are legit but many of them are false positives. Details are called out in the comments.

#### Pull request checklist

- [] Run through of all [test scenarios](https://github.com/Microsoft/accessibility-insights-windows/blob/master/docs/Scenarios.md) completed?
- [n/a] Includes UI changes?

### DIFF
--- a/src/AccessibilityInsights.Extensions.AzureDevOps/ConnectionInfo.cs
+++ b/src/AccessibilityInsights.Extensions.AzureDevOps/ConnectionInfo.cs
@@ -45,16 +45,8 @@ namespace AccessibilityInsights.Extensions.AzureDevOps
         public ConnectionInfo(Uri serverUrl, TeamProject project, AdoTeam team)
         {
             ServerUri = serverUrl;
-            if (project != null)
-            {
-                TeamProject typedProject = project as TeamProject;
-                Project = typedProject ?? new TeamProject(project);
-            }
-            if (team != null)
-            {
-                AdoTeam typedTeam = team as AdoTeam;
-                Team = typedTeam ?? new AdoTeam(team);
-            }
+            Project = project;
+            Team = team;
         }
 
         /// <summary>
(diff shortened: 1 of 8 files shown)


src/AccessibilityInsights.Extensions.AzureDevOps/ConnectionInfo.cs
```diff
@@ -45,16 +45,8 @@ public class ConnectionInfo
...
         {
             ServerUri = serverUrl;
-            if (project != null)
-            {
-                TeamProject typedProject = project as TeamProject;
-                Project = typedProject ?? new TeamProject(project);
-            }
-            if (team != null)
-            {
-                AdoTeam typedTeam = team as AdoTeam;
```
DaveTryon, line 55: Since team is already an AdoTeam, typedTeam will be identical to team, which is non-null. As such, we can simply the assignment of Team.

src/AccessibilityInsights/MainWindowHelpers/StateMachine.cs
```diff
@@ -153,6 +153,7 @@ public void HandleBackToSelectingState()
             /// make sure that all Mode UIs are clean since new selection will be done. 
             CleanUpAllModeUIs();
 
+#pragma warning disable CA1508 // Dead code warning doesn't apply here
```
DaveTryon, line 156: This is claiming that these 2 lines can never be satisfied, but I don't believe it:
                || (this.CurrentPage == AppPage.Events)
                || (this.CurrentPage == AppPage.Inspect && SelectAction.GetDefaultInstance().IsPaused)

src/AccessibilityInsights.SharedUx/Controls/EventConfigurationControl.xaml.cs
```diff
@@ -235,6 +235,7 @@ public void SetLists(List<RecordEntitySetting> list, List<RecordEntitySetting> s
 
         private void UpdateListViews()
         {
+#pragma warning disable CA1508 // Dead code warning doesn't apply here
```
DaveTryon, line 238: Analyzer says that we're checking for null in the lines that new up the ListViewSelectionLock objects, but I don't believe it

jalkire (approved):

PR #1081, opened 2021-03-16
chore: Move property management into AIWin (from Axe.Windows)

DaveTryon:
#### Details

While addressing some analyzer issues in Axe.Windows, I realized that the following methods (and their backing code) in https://github.com/microsoft/axe-windows/blob/main/src/Desktop/UIAutomation/DesktopElementHelper.cs are not used anywhere in Axe.Windows, and should be migrated into AIWin:

`SetCorePropertiesList`
`GetCorePropertiesList`
`GetDefaultCoreProperties`

After talking with @jalkire, we've chosen the following plan of action:
1. Give AIWin its own implementation of this functionality (this PR)
2. Remove the code from Axe.Windows (https://github.com/microsoft/axe-windows/pull/550). 

I've tested the changes locally and we seem to have parity with the old code, including resetting to the default set of properties if the user happens to remove all properties from the config (it may be a separate conversation whether or not that behavior is correct)

##### Motivation

Improve code cohesion (and will help to address CA1024 warnings in Axe.Windows)

#### Pull request checklist

- [x] Does this address an existing issue? If yes, Issue# - 
- [x] Includes UI changes?

### DIFF
--- a/src/AccessibilityInsights.SharedUx/Controls/PatternInfoControl.xaml.cs
+++ b/src/AccessibilityInsights.SharedUx/Controls/PatternInfoControl.xaml.cs
@@ -1,11 +1,11 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 using AccessibilityInsights.CommonUxComponents.Controls;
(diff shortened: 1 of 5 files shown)


jalkire (approved):
Works locally!

PR #1089, opened 2021-03-22
chore: mention that ruleset bugs go to axe.windows

karanbirsingh:
#### Details

Often we need to transfer issue reports from AI-Windows to Axe.Windows because the questions are rule-related. This updates the issue template to encourage ruleset questions should go to Axe.Windows.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -6,9 +6,9 @@ labels: bug, needs triage
 assignees: ''
 
 ---
-If the bug is about the ruleset, consider filing this issue at [axe-windows](https://github.com/microsoft/axe-windows/issues/new/choose) instead.
-
 Please check whether the bug has [already been filed](https://github.com/Microsoft/accessibility-insights-windows/issues).
+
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -6,6 +6,8 @@ labels: bug, needs triage
 assignees: ''
 
 ---
+If the bug is about the ruleset, consider filing this issue at [axe-windows](https://github.com/microsoft/axe-windows/issues/new/choose) instead.
```
dbjorge, line 9: I'm not sure "the ruleset" is going to be clear enough for an issue-filer that isn't very familiar with the codebase; maybe expanding on this and/or including an example would be helpful?

dbjorge: I'd also rather have a bug in the wrong place than no bug, so maybe worth clarifying that if you aren't sure where to file it, it's okay to use this repo.

karanbirsingh: thanks, will amend!

dbjorge (approved):

Issue #1123, opened 2021-05-17
The tooltip displays previously selected item in a combo box


SergeySmirnov-Akvelon:
**Describe the bug**
We found oddities in Accessibility Insights behavior.  We have a ComboBox that contains "bbk" and "dd" items. For this case, the AI ​​works correctly and correctly highlights the selected items.
![issue-4364-expectedbehavior](https://user-images.githubusercontent.com/23376742/117785032-d96d2700-b24c-11eb-8ab7-5273abc534e9.gif)

But if I make the elements have the same number of characters ("bb" and "dd"), then the Accessibility Insights ​​​​starts to behave incorrectly.
![issue-4364-wrongbehavior](https://user-images.githubusercontent.com/23376742/117785341-28b35780-b24d-11eb-8611-284144e87368.gif)

Although in the Accessibility Insights ​​itself, it correctly indicates which element is selected now:
![image](https://user-images.githubusercontent.com/23376742/117785842-9d869180-b24d-11eb-9147-bde94cbb8ca0.png)

I have verified that this problem is reproducible for other applications as well:
**Microsoft Word**
![issue-4364-word](https://user-images.githubusercontent.com/23376742/117785884-a9725380-b24d-11eb-96ec-7a910145511a.png)

**WPF**
![issue-4364-WPF](https://user-images.githubusercontent.com/23376742/117785917-b2632500-b24d-11eb-8a36-13b92f504c9b.png)

**To Reproduce**
1. Create application with the Combobox.
2. Add the "aa" and "bb" items to ComboBox.
3. Run application and run Accessibility Insights
4.  Expand the Combobox and hover the mouse cursor over the "aa" and "bb" items

**Expected behavior**
Correct item should be displayed in ‘specific item focusable’ form

**Actual behavior**
Incorrect item is displayed in ‘specific item focusable’ form

**Desktop (please complete the following information):**
 - Microsoft Windows [Version 10.0.19041.388]
 - Accessibility Insights for Windows Version: v1.1.1562.01

jalkire:
I was able to repro on Word.

DaveTryon:
Here's a small sample app to demo this behavior. Note that it occurs even though all of the list items are the same length.
[ComboBoxTestSampleApp.zip](https://github.com/microsoft/accessibility-insights-windows/files/6497289/ComboBoxTestSampleApp.zip)

karanbirsingh:
[...]

PR #1125, opened 2021-05-18
fix: ensure Tooltip stays up to date with highlighted element

karanbirsingh:
#### Details

This PR fixes stale tooltip updates described in 1123.

It also suppresses an instance of CA1062 after the refactor - I think the warning is a false positive because `element != null && !element.BoundingRectangle.IsEmpty;` should short-circuit if `element` is null. Let me know if I'm mistaken or there's a better way to match with the rule.

##### Motivation

addresses #1123 

##### Context

This fix works because we move the `IsVisible` setter after `Highlighter.SetText`.

The issue in 1123 occurs because:
- The TextTip text is drawn when [`WndProc`](https://github.com/microsoft/accessibility-insights-windows/blob/main/src/AccessibilityInsights.SharedUx/Highlighting/TextTip.cs#L364) is triggered with `WM_ERASEBKGND`
- `WndProc` is triggered if we explicitly call [`InvalidateRect`](https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-invalidaterect) via [`Update`](https://github.com/microsoft/accessibility-insights-windows/blob/effa0a9d0958e5188877dc5e2c5f183043fc9ced/src/AccessibilityInsights.SharedUx/Highlighting/TextTip.cs#L159), or if it's [raised implicitly due to window resize](https://docs.microsoft.com/en-us/windows/win32/winmsg/wm-erasebkgnd)
- The `Highlighter.IsVisible` setter explicitly calls `Update` via [`LoadHighlighterMode`](https://github.com/microsoft/accessibility-insights-windows/blob/effa0a9d0958e5188877dc5e2c5f183043fc9ced/src/AccessibilityInsights.SharedUx/Highlighting/Highlighter.cs#L68).
- However, in `main`, `Highlighter.IsVisible` is set *before* `Highlighter.SetText` is called. This is problematic when the tooltip size stays constant, because [`SetWindowPos`](https://github.com/microsoft/accessibility-insights-windows/blob/effa0a9d0958e5188877dc5e2c5f183043fc9ced/src/AccessibilityInsights.SharedUx/Highlighting/TextTip.cs#L226) then won't trigger `WndProc` implicitly with `WM_ERASEBKGND` after the text change.
- The text 'lags one behind' because `Highlighter.SetText` is called at the end of [`SetElementInternal`](https://github.com/microsoft/accessibility-insights-windows/blob/effa0a9d0958e5188877dc5e2c5f183043fc9ced/src/AccessibilityInsights.SharedUx/Highlighting/HollowHighlightDriver.cs#L85). The explicit `Update` call occurs the next time around.

We fix this by moving `Highlighter.IsVisible = true` to the end, after all rectangle & text changes. Then these changes are reflected in `LoadHighlighterMode` > `Update`.

#### Pull request checklist

- [x] Does this address an existing issue? If yes, Issue# - 1123
- [ ] Includes UI changes?

### DIFF
--- a/src/AccessibilityInsights.SharedUx/Highlighting/HollowHighlightDriver.cs
+++ b/src/AccessibilityInsights.SharedUx/Highlighting/HollowHighlightDriver.cs
@@ -102,7 +102,7 @@ namespace AccessibilityInsights.SharedUx.Highlighting
                 }
             }
 
-            this.Highlighter.IsVisible = visible;
+            this.Highlighter.IsVisible = this.IsEnabled && visible;
         }
 
         /// <summary>
--- a/src/AccessibilityInsights.SharedUx/Highlighting/HollowHighlightDriver.cs
+++ b/src/AccessibilityInsights.SharedUx/Highlighting/HollowHighlightDriver.cs
@@ -75,7 +75,9 @@ namespace AccessibilityInsights.SharedUx.Highlighting
         /// <param name="el"></param>
         public void SetElement(A11yElement el)
         {
+#pragma warning disable CA1062 // Validate arguments of public methods
             SetElementInternal(el);
+#pragma warning restore CA1062 // Validate arguments of public methods
         }
 
         /// <summary>
@@ -84,7 +86,8 @@ namespace AccessibilityInsights.SharedUx.Highlighting
         /// <param name="element"></param>
         void SetElementInternal(A11yElement element)
         {
-            if (element != null && !element.BoundingRectangle.IsEmpty)
+            bool visible = element != null && !element.BoundingRectangle.IsEmpty;
+            if (visible)
             {
                 if (this.BoundingRectangle == null || element.BoundingRectangle.Equals(this.BoundingRectangle) == false)
                 {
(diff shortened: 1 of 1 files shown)


src/AccessibilityInsights.SharedUx/Highlighting/HollowHighlightDriver.cs
```diff
@@ -94,15 +97,12 @@ void SetElementInternal(A11yElement element)
...
                     var text = GetHighlightText(element);
                     this.Highlighter.SetText(text);
                 }
             }
-            else
-            {
-                this.Highlighter.IsVisible = false;
-            }
+
+            this.Highlighter.IsVisible = visible;
```
DaveTryon, line 105: In the original code, the guard on line 95 ensured that `IsVisible` was only set to true if `IsEnabled` was also true. We can retain that behavior with something like this:
```suggestion
            this.Highlighter.IsVisible = this.IsEnabled && visible;
```

DaveTryon:
Nice sleuthing to figure out the root cause here!

src/AccessibilityInsights.SharedUx/Highlighting/HollowHighlightDriver.cs
karanbirsingh: Thanks!

DaveTryon (approved):

Issue #542, opened 2019-09-16
Implement a channel switch progress indicator


Shobhit1:
**Is your feature request related to a problem? Please describe.**
So, if I am on a machine with slow internet connection or a VM (which generally has a slower internet connection then the base machine); changing channels from Production to Canary or from canary back to production take a long time. Also, if I open the accessibility-insights-windows application in between the process, the app seems to work in its last state.

**Describe the solution you'd like**
- a progress indicator on how the channel switching is happening.
- if the application in its previous channel is being uninstalled; then I shouldn't be able to use the application until I have switched channels.

Sirisha-Kommushetty:
Thanks for filing the issue. We're looking into how this would apply for the long term.

adarshrema:
Hello @shobhit1,
Thank you for your interest in Accessibility Insights and this feature request. This is great feedback and needs more investigation by the team.

We will keep this active and will update this here when we know more.

adarshrema:
Needs design input. Keeping this active to prioritize against other work items.

DaveTryon:
A very simple implementation of this would be cheap and a significant improvement over what we have (which is no feedback at all)

DaveTryon:
We want to pursue this, pending an approved design

PR #1164, opened 2021-07-26
fix: Add UI to version switcher

DaveTryon:
#### Details

Issue #542 asks for some sort of progress indicator when changing channels or applying upgrades. In the current code (see GIF below) we just exit the app, do some invisible work, then restart the app. The invisible work includes downloading a file from the web and might take a while if you're on a very slow web connection. This PR simply adds a very simple UI that tells the user that we're downloading, and displays a WPF progress bar to show that the app isn't hung. The progress bar assumes that 80% of the time is spent downloading (not always correct but a guess), then 5% validating, 10% doing the actual install, then 5% for final configuration.

The window is intentionally small to allow it to scale nicely to 400% zoom. It's auto-centered by WPF (it looks off-center in the GIF files because my capture window was a subset of the screen)

##### Motivation

Address #542. It may not be the final version, but it's better than nothing.

##### Screenshots

Note that there's a delay after pressing "Change Channel"--that's where the OS is displaying the LUA dialog, but it doesn't get picked up in the screen capture software. 

Description | GIF
--- | ---
Current (no UI at all) | channel-without-ui
Updated (simple UI) | channel-with-percent-ui

#### Pull request checklist

- [x] Does this address an existing issue? If yes, Issue# - #542 
- [x] Includes UI changes?


DaveTryon:
@ferBonnin, please give this a once-over for the experience. It's intentionally very simple to minimize impact on PM and design. Thanks!

ferBonnin:
@DaveTryon this looks good to me as its very similar to the other wizard UI and simple in the experience. Have we tested its accessibility and ensured it provides the appropriate announcements?

karanbirsingh (approved):

PR #1187, opened 2021-09-02
fix: Use underlines for hyperlinks

DaveTryon:
#### Details

In a recent trusted tester pass, a bug was filed about distinguishing hyperlinks only by color, since they had no underlines. This was a design decision made a long time ago, but since we've put underlines in some places in response to trusted tester bugs, I just went ahead and put underlines in all hyperlinks. More specifically, this involved 2 changes:

1. We were suppressing the underlines by overriding the **TextDecorations** property. The PR removes all of those overrides.
2. In the tbLink style, we were only setting the Underline style on mouse hover. This now sets the style all the time

##### Motivation

Address trusted tester bug

##### Screenshots
I'm just including the Color Contrast screen and the test results screen in the screenshots, even though the change impacts other places, as well:

Location | Before | After
--- | --- | ---
Color Contrast | image | CCA
Test Results | No_Underlines | Underlines

##### Context

The bug also makes a comment about color contrast ratio, but I checked with @ferBonnin who clarified that if we use underlines our current colors will be OK.

#### Pull request checklist

- [x] Does this address an existing issue? If yes, Issue# - [1868536](https://dev.azure.com/mseng/1ES/_workitems/edit/1868536)
- [x] Includes UI changes?

### DIFF
--- a/src/AccessibilityInsights.SharedUx/Controls/PrivacyLearnMore.xaml
+++ b/src/AccessibilityInsights.SharedUx/Controls/PrivacyLearnMore.xaml
@@ -14,8 +14,7 @@
                 AutomationProperties.Name="{x:Static Properties:Resources.PrivacyLearnMoreLink}" 
                 NavigateUri="https://privacy.microsoft.com/en-US/privacystatement" 
                 RequestNavigate="Hyperlink_RequestNavigate" 
-                FocusVisualStyle="{DynamicResource {x:Static SystemParameters.FocusVisualStyleKey}}" 
-                TextDecorations="None">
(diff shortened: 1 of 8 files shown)


jalkire:
Does this concern apply to our hyperlink-styled buttons as well, or only hyperlinks? For example, the breadcrumbs in the top left area are only distinguished from normal text by their color.

jalkire (approved):
Approved with a question

DaveTryon:
@jalkire, I've updated the PR to also include cases where we use buttons as links

PR #1192, opened 2021-09-10
fix: Allow tooltips to display until the user takes action

DaveTryon:
#### Details

The current code hides tooltips after 5 seconds, which violates current accessibility requirements. Change the code to display them until the user takes some action that would dismiss them.

We actually had code that was hardcoded to a 5 second dismissal of the tooltips, but 5 seconds also seems to be the default value from WPF. I found no way to tell WPF not to hide tooltips at all, but I was able to set the duration to int.MaxValue

##### Motivation

Address trusted tester bug

##### Screenshots

These screenshots only show the tooltip for about 15 seconds, but it will last much longer. In fact, int.MaxValue in milliseconds represents about 24 days of actual time...

Old | New
--- | ---
Tooltip-Timer | Tooltip-Indefinite

#### Pull request checklist

- [x] Does this address an existing issue? If yes, Issue - [1869230](https://mseng.visualstudio.com/DefaultCollection/1ES/_workitems/edit/1869230)
- [x] Includes UI changes?

### DIFF
--- a/src/AccessibilityInsights.SharedUx/Behaviors/KeyboardToolTipButtonBehavior.cs
+++ b/src/AccessibilityInsights.SharedUx/Behaviors/KeyboardToolTipButtonBehavior.cs
@@ -60,18 +60,6 @@ namespace AccessibilityInsights.SharedUx.Behaviors
                     tt.Placement = PlacementMode.Left;
                     tt.PlacementRectangle = new Rect(0, (sender as Control).Height + 5, 0, 0);
                     tt.IsOpen = true;
-
-                    // set up a timer to clear the tooltip after 5 secs if focus has not been changed
-                    DispatcherTimer timer = new DispatcherTimer() { Interval = TimeSpan.FromSeconds(5) };
-                    timer.Tick += delegate (object _sender, EventArgs _e)
-                    {
-                        ((DispatcherTimer)timer).Stop();
-                        if (tt.IsOpen)
-                        {
-                            tt.IsOpen = false;
-                        }
-                    };
-                    timer.Start();
                 }
             }
         }
--- a/src/AccessibilityInsights/App.xaml.cs
+++ b/src/AccessibilityInsights/App.xaml.cs
@@ -116,9 +116,9 @@ namespace AccessibilityInsights
             }
 
             // WPF hides tooltips after a few seconds, which is bad for accessibility.
-            // Override the default for the longest possible value.
+            // Override the default to 1 day
             ToolTipService.ShowDurationProperty.OverrideMetadata(typeof(UIElement),
-                new FrameworkPropertyMetadata(int.MaxValue));
+                new FrameworkPropertyMetadata((int)TimeSpan.FromDays(1).TotalMilliseconds));
         }
     }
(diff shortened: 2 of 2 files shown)


jalkire:
Is there a way for the user to dismiss these now that they won't go away on their own (or should there be?)? I tried esc with no luck, which dismisses tooltips in Edge and Explorer

jalkire (approved):
Approve with a question/suggestion

peterdur:
> Is there a way for the user to dismiss these now that they won't go away on their own (or should there be?)? I tried esc with no luck, which dismisses tooltips in Edge and Explorer

[This page in the WPF repo](https://github.com/dotnet/wpf/blob/920b0b06b3246c851dd61a37eeb6238b01368971/src/Microsoft.DotNet.Wpf/specs/tooltip.md#closing-a-tooltip) suggests that Ctrl should close the tooltip (may not be true in the version of WPF we're using, though?)

jalkire:
> > Is there a way for the user to dismiss these now that they won't go away on their own (or should there be?)? I tried esc with no luck, which dismisses tooltips in Edge and Explorer
> 
> [This page in the WPF repo](https://github.com/dotnet/wpf/blob/920b0b06b3246c851dd61a37eeb6238b01368971/src/Microsoft.DotNet.Wpf/specs/tooltip.md#closing-a-tooltip) suggests that Ctrl should close the tooltip (may not be true in the version of WPF we're using, though?)

Ctrl isn't working for me, but Ctrl+Shift+F10 is. So at least there is a way!

peterdur:
A comment on [this stack overflow page](https://stackoverflow.com/questions/896574/forcing-a-wpf-tooltip-to-stay-on-the-screen) suggests using `int32.MaxValue` here may be dangerous as it's [converted to a TimeSpan](https://github.com/dotnet/wpf/blob/9a89bc76b2d13d906c31db46728e2bdd79efa946/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/PopupControlService.cs#L416) using [TimeSpan.FromMilliseconds](https://github.com/dotnet/runtime/blob/b106e50cef63cf831d1d0ba6f2be503478ff0720/src/libraries/System.Private.CoreLib/src/System/TimeSpan.cs#L231) which converts to a `double` and multiplies by 10000, then is converted back to milliseconds [(still as a double)](https://github.com/dotnet/runtime/blob/b106e50cef63cf831d1d0ba6f2be503478ff0720/src/libraries/System.Private.CoreLib/src/System/TimeSpan.cs#L113) and then [compared against `int32.MaxValue`](https://github.com/dotnet/wpf/blob/3ca050a987c290309f01c235ceddf29cab798f14/src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Windows/Threading/DispatcherTimer.cs#L91), throwing an ArgumentOutOfRangeException if greater. This _should_ be working out OK, since we're not seeing this exception thrown, but we are scraping against the guard rail. Given that `int32.MaxValue` milliseconds is 24 days, do we want to choose a more conservative number (maybe 1 day?)

PR #1195, opened 2021-09-15
fix: Add a handler to clear button tooltip when Escape is pressed

DaveTryon:
#### Details

#1192 Made tooltips stay visible indefinitely (instead of disappearing after 5 seconds). What was lacking, though, was a way to dismiss the tooltip to get it out of the way, and the bug was reactivated as a result. This PR adds a keyboard handler to dismiss the tooltip when the user presses Escape, as described in https://www.w3.org/WAI/WCAG21/Understanding/content-on-hover-or-focus.html.

##### Motivation

Fix trusted tester bug

##### Gif of feature in action
esc-tooltips

##### Context

Dismiss on escape was raised in the comments for #1192, but since the framework dismissed on Shift+Ctrl+F10, we believed that would suffice. To meet WCAG guidelines, we need something easier.

#### Pull request checklist

- [x] Does this address an existing issue? If yes, Issue - [1869230](https://mseng.visualstudio.com/1ES/_workitems/edit/1869230)
- [x] Includes UI changes?

### DIFF
--- a/src/AccessibilityInsights.SharedUx/Behaviors/KeyboardToolTipButtonBehavior.cs
+++ b/src/AccessibilityInsights.SharedUx/Behaviors/KeyboardToolTipButtonBehavior.cs
@@ -6,7 +6,6 @@ using System.Windows;
 using System.Windows.Controls;
 using System.Windows.Controls.Primitives;
 using System.Windows.Input;
-using System.Windows.Threading;
 
 namespace AccessibilityInsights.SharedUx.Behaviors
 {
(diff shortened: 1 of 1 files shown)


src/AccessibilityInsights.SharedUx/Behaviors/KeyboardToolTipButtonBehavior.cs
```diff
@@ -6,7 +6,6 @@
 using System.Windows.Controls;
 using System.Windows.Controls.Primitives;
 using System.Windows.Input;
-using System.Windows.Threading;
```
DaveTryon, line 9: This should have been removed with #1192

jalkire (approved):

PR #1196, opened 2021-09-16
fix: Ensure windows bounds at all zoom levels

DaveTryon:
#### Details

Our window positioning code has some problems:
1. It never resizes the window, even if it's tool big to fit on screen
2. It fails to handle cases where the window position is `double.NaN` (which happens on first boot). As documented at https://docs.microsoft.com/en-us/dotnet/api/system.double.nan?view=net-5.0, comparing `double.NaN` to any other value (except `double.NaN`) will return false, and this is breaking the current code.

Because of these problems, it's possible to have parts of the app inaccessible on first boot at 200% zoom.

The fix is to rework the code that sets the initial window position so that the entire app is always in the virtual window, even if we need to resize the window (old code could only reposition, not resize). I also added some unit tests to make it easier to exercise the various pathways. This meant adding a new project to contain these unit test, but hopefully future changes to this project will also include unit tests. If we were to omit the unit tests, we'd only have modify 1 file.

##### Motivation

Address bug from trusted tester pass

##### Screenshots
These were all captured on first boot with 200% zoom on a 1280x1024 display (the highest zoom/smallest size that we support), run after deleting all config files (so it simulates the first boot experience):

View | Screenshot
--- | ---
Telemetry | image
Welcome | image
Live Inspect | image
About dialog | image

##### Context

There was some discussion of adding a ScrollViewer inside the About page. Looking at the other pages (Connection and Settings), though, neither of those has a ScrollViewer, so it's coming from external. To test this, I (locally) forced the About dialog to extend beyond the window size, and the ScrollBar appeared without any need to explicitly specify it. Since it's not needed, I left it out of the final PR and just focused on the window positioning.

#### Pull request checklist

- [x] Does this address an existing issue? If yes, Issue - [1868538](https://mseng.visualstudio.com/DefaultCollection/1ES/_workitems/edit/1868538)
- [x] Includes UI changes?

### DIFF
--- a/src/AccessibilityInsights/MainWindowHelpers/Initialize.cs
+++ b/src/AccessibilityInsights/MainWindowHelpers/Initialize.cs
@@ -422,6 +422,23 @@ namespace AccessibilityInsights
                 SystemParameters.VirtualScreenWidth, SystemParameters.VirtualScreenHeight);
         }
 
+        /// <summary>
+        /// Return the first non-NaN number from an array of potentialNumbers.
+        /// Will return 0.0 if they're all NaN.
+        /// </summary>
+        private static double GetFirstNumber(double[] potentialNumbers)
+        {
+            foreach(double potentialNumber in potentialNumbers)
+            {
+                if (!double.IsNaN(potentialNumber))
+                {
+                    return potentialNumber;
+                }
+            }
+
+            return 0.0;
+        }
+
         internal static void EnsureWindowIsInVirtualScreenWithInjection(AppLayout layout,
             double windowTop, double windowLeft, double virtualLeft,
             double virtualTop, double virtualWidth, double virtualHeight)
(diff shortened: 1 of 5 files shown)


src/AccessibilityInsights/MainWindowHelpers/Initialize.cs
```diff
@@ -411,45 +411,83 @@ private Action InvokeAction(Action action)
...
+            double virtualTop, double virtualWidth, double virtualHeight)
         {
-            var layout = ConfigurationManager.GetDefaultInstance().AppLayout;
+            double virtualRight = virtualLeft + virtualWidth;
+            double virtualBottom = virtualTop + virtualHeight;
 
             if (layout != null)
             {
+                layout.Top = GetFirstNumber(new double[] { layout.Top, windowTop, 200.0 });
+                layout.Left = GetFirstNumber(new double[] { layout.Left, windowLeft, 200.0 });
```
jalkire, line 452: Am I correct in understanding that these 200.0s are fallback values in case of NaNs? If so, this might benefit from a comment or pulling out as a const like `fallbackWindowPosition =  200.0`

jalkire (approved):
Approve with suggestion

Issue #1226, opened 2021-10-25
[BUG] Canary only updates to the latest prod release


jalkire:
**Describe the bug**
Our canary update mechanism attempts to update to the latest prod (not canary) release. This only impacts the autoupdater on launch--when you switch channels, ai-win updates to the correct latest version for your selected channel. 

**To Reproduce**
1. Install the latest prod version of ai-win (currently 1741.1)
2. Modify its config file (`<user dir>\AppData\Local\AccessibilityInsights\V1\Configurations\Configuration.json`) to contain `"ReleaseChannel": "Canary"`
3. Launch ai-win
4. Observe that it does not attempt to update despite the several canary releases since prod

**Expected behavior**
You should receive an update prompt on launch to update to the [latest canary release](https://github.com/microsoft/accessibility-insights-windows/releases)

**Actual behavior**
No update is prompted or occurs

**Additional context**
- From the [signed build pipeline](https://dev.azure.com/mseng/1ES/_build?definitionId=8122&_a=summary), I downloaded old releases and followed the above repro steps with them. I found that starting with 1733.1 ([build](https://dev.azure.com/mseng/1ES/_build/results?buildId=16093849&view=results)), ai-win updates to the latest prod release only, not to the latest canary
- This does not impact prod channel updates
- @DaveTryon noticed this issue due to some weirdness in our telemetry data

DaveTryon:
Fiddler shows that the signed Canary build is always requesting the Production data. Curiously, it requests Canary data if run under the debugger. Bottom line, though, is that it shouldn't impact any production customers.

DaveTryon:
@jalkire, this is fixed in the release at https://github.com/microsoft/accessibility-insights-windows/releases/tag/v1.1.1767.01. If you'll install it and set your ReleaseChannel to Canary, it will auto-upgrade to the next Canary build when it comes out. I just queued a build that will become 1.1.1767.02 so we'll have a target for upgrading.

jalkire:
1.1.1767.1 successfully updated to 1.1.1767.2 👍

PR #1204, opened 2021-09-29
fix: Allow system administrators to suppress telemetry by policy

DaveTryon:
#### Details

Create a registry-based policy to disable telemetry from a computer. The suppression is system-wide and persists across uninstall/reinstall. The suppression is added by creating the following registry value:

Key: HKEY_LOCAL_MACHINE\SOFTWARE\Policies\Accessibility Insights for Windows
Value (DWORD): DisableTelemetry = 1

In terms of a reg file, it looks like this:
```
Windows Registry Editor Version 5.00

[HKEY_LOCAL_MACHINE\SOFTWARE\Policies\Accessibility Insights for Windows]
"DisableTelemetry"=dword:00000001
```

If this value exists and is set to 1, then the initial telemetry prompt is skipped. Inside the settings dialog, the standard telemetry block is also changed, as shown here:

Key not set (default) | Key set (with this change)
--- | ---
image | image

The key is read when the app starts up, and changes to the key will be ignored until after the app is restarted.

##### Motivation

Addresses #1182 

##### Context

I also walked through this extended scenario:
1. Disable telemetry via registry
2. Start without configuration files
3. Telemetry dialog is skipped (user can't enable it)
4. Close app
5. Enable telemetry via registry
6. Start app (with or without configuration files)
7. Telemetry dialog displays

This works because the ShowTelemetryDialog flag defaults to true, then never gets set to false. When the registry override is removed, we display the telemetry dialog, as expected.

#### Pull request checklist

- [ ] Run through of all [test scenarios](https://github.com/Microsoft/accessibility-insights-windows/blob/main/docs/Scenarios.md) completed?
- [x] Does this address an existing issue? If yes, Issue - #1182 
- [x] Includes UI changes?
  - [x] Run the production version of Accessibility Insights for Windows against a version with changes.
  - [x] Attach any screenshots / GIF's that are applicable.

> Note: After the PR has been created, certain checks will be kicked off. All of these checks must pass before a merge.

PR #1232, opened 2021-11-02
fix: Ensure proper sequencing during initialization

DaveTryon:
#### Details

Fix canary upgrade problem by ensuring that ordering assumptions are respected during initialization. For the autoupdate to work as expected, it's important that `Container.SetAutoUpdateReleaseChannel` gets called before `Container.GetDefaultInstance`. When #1204 was added, the code is in the proper sequence, and the upgrade works as expected when built for the debug config. When built for the release config, though, I was able to prove that the calls were occurring in the wrong order. It seems that the optimizer decided to change the sequence of calls between the API's, meaning that the autoUpdate code always loaded the production data.

The change here is to move the telemetry initialization code into a separate method so the optimizer can no longer change the call order. As an aside, enabling telemetry in a method called `PopulateConfigurations` was a bit off, so having it in a separate method is also better coding style. I included a comment in `PopulateConfigurations` as a reminder about the calling order dependency.

##### Motivation

Address #1226 

#### Pull request checklist

- [x] Does this address an existing issue? If yes, Issue - #1226 
- [n/a] Includes UI changes?

### DIFF
--- a/src/AccessibilityInsights/MainWindow.xaml.cs
+++ b/src/AccessibilityInsights/MainWindow.xaml.cs
@@ -162,6 +162,8 @@ namespace AccessibilityInsights
             // create necessary config folders & their internal config files
             PopulateConfigurations(telemetryBuffer);
 
+            InitializeTelemetry();
+
             SetFontSize();
 
             InitCommandBindings();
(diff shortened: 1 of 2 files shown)


jalkire (approved):
It would be nice if we could have a more robust way to avoid this risk in the future, but I can't think of any great options

Issue #845, opened 2020-07-21
Dark mode: Header text in "how to fix" has insufficient contrast against its background on mouse hover


iamrafan:
**Describe the bug**
The header text in "how to fix" section in the right pane has insufficient contrast against its
background on mouse hover

**To Reproduce**
1. Make sure color setting for default app mode in windows is set to 'Dark'
2. Launch AI-Windows
3. Scan any app and navigate to "Results in UI Automation Tree"
4. Hover the mouse on the headers "Rule" and "Issues" under the test details section in "how to fix" tab
5. Observe that the header text has insufficient contrast against its hovered background

**Expected behavior**
The header text should have contrast ratio of at least 4.5:1 against its background

**Actual behavior**
The contrast ratio for the header text is less than 4.5:1

**Screenshots or .GIF**
![5](https://user-images.githubusercontent.com/22068434/88111245-6d58ef80-cb62-11ea-8cf7-87085a05abdc.png)

**Desktop**
 - OS: Windows 10 Enterprise, Version 1903
 - Accessibility Insights for Windows Version: Production Release: v 1.1.1248.1

ferBonnin:
reactivating to allow original issue creator to close. 
@iamrafan could you please regress and close?

DaveTryon:
@iamrafan, this is in any build after https://github.com/microsoft/accessibility-insights-windows/releases/tag/v1.1.1780.01

PR #1242, opened 2021-11-09
AI-Win color contrast bug

ThanyaLeif:
#### Details

This is a quick fix to the color contrast problem reported in bug #845 

Due to problems with the current implementation of the GridViewColumnHeader, it is not possible to customize the background properties (the color changes are not reflected in the element). This is the reason why I decided to change the foreground color to increase the ratio between the header and the background as shown in the following picture.

image
(The image above shows the color contrast analyzer test of the background color against the foreground color, which gives a ratio of 8.917:1)

| Mode  | Image |
| ------------- | ------------- |
| Light  | image |
| Dark  | image |
| HC1  | image |
| HC2  | image|
| HC Light  | image |
| HC Dark  | image  |

##### Motivation

This PR intends to address issue #845 through a quick fix that allows the header to be displayed during the onMouseOver event, since during this bug's investigation multiple errors related to the GridViewColumnHeader implementation were discovered, such as:

- The inability to change the background color of the element
- Inconsistency between multiple implementations of the GridViewColumnHeader element throughout the project.
- The unintentional ability to move columns around.

A suitable solution to all these problems will be addressed in a future bug.

#### Pull request checklist

- [x] Run through of all [test scenarios](https://github.com/Microsoft/accessibility-insights-windows/blob/main/docs/Scenarios.md) completed?
- [x] Does this address an existing issue? If yes, Issue #845  
- [x] Includes UI changes?

### DIFF
--- a/src/AccessibilityInsights.SharedUx/Controls/ScannerResultControl.xaml
+++ b/src/AccessibilityInsights.SharedUx/Controls/ScannerResultControl.xaml
@@ -83,7 +83,11 @@
                             <ListView.View>
                                 <GridView x:Name="gvRules">
                                     <GridViewColumn>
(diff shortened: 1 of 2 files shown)


DaveTryon:
@ThanyaLeif, can you please also include the information for light mode and the 4 Windows HC modes? Thanks!

DaveTryon (approved):

PR #1254, opened 2021-11-22
feat(compliance-fall-21): Improve localizability in SharedUx project

DaveTryon:
#### Details

This PR improves localization in the `AccessibilityInsights.SharedUx` project. Some of the changes replace a hardcoded string with a new resource string, while others replace hardcoded concatenation of resource strings with a parameterized resource string (which would allow the parameters to be used in any order).

##### Motivation

Global Readiness component of compliance feature.

##### Context

Each commit includes strings from a single .cs file. It might be easier to review each commit separately. There is at least 1 overlapping commit, because the first commit to `HierarchyViewNodeModel` created a bug that got fixed in the second commit to `HierarchyViewNodeModel`. The second commit also renamed some of the string resources to be more consistent with the new ones (and to simplify the names).

This PR does not fully address the various places where we hardcode the formatting strings like for how data is presented (strings like `"{0}: {1}` or `"{0} {1}"`, but that feels beyond the scope of this effort. Perhaps we can improve that in the next compliance pass or future slack time, perhaps not.

The other thing that I haven't changed is that I'm still using `CultureInfo.InvariantCulture` for user-visible strings. I may make a separate pass to do that if I have time later during this feature.

#### Pull request checklist

- [n/a] Does this address an existing issue? If yes, Issue: 
- [n/a] Includes UI changes?

### DIFF
--- a/src/AccessibilityInsights.SharedUx/Controls/TestTabs/AutomatedChecksControl.xaml.cs
+++ b/src/AccessibilityInsights.SharedUx/Controls/TestTabs/AutomatedChecksControl.xaml.cs
@@ -23,10 +23,8 @@ using System.Windows.Automation.Peers;
 using System.Windows.Controls;
 using System.Windows.Controls.Primitives;
 using System.Windows.Data;
-using System.Windows.Documents;
 using System.Windows.Input;
 using System.Windows.Media;
-using System.Windows.Navigation;
 using System.Windows.Threading;
 
 namespace AccessibilityInsights.SharedUx.Controls.TestTabs
--- a/src/AccessibilityInsights.SharedUx/Controls/TestTabs/AutomatedChecksControl.xaml.cs
+++ b/src/AccessibilityInsights.SharedUx/Controls/TestTabs/AutomatedChecksControl.xaml.cs
@@ -317,7 +317,8 @@ namespace AccessibilityInsights.SharedUx.Controls.TestTabs
                     this.chbxSelectAll.IsEnabled = ScreenshotAvailable;
                     this.lvResults.ItemsSource = null;
                     this.ElementContext = ec;
-                    this.tbGlimpse.Text = "Target: " + ec.Element.Glimpse;
+                    this.tbGlimpse.Text = string.Format(CultureInfo.InvariantCulture,
+                        Properties.Resources.AutomatedChecksControl_TargetFormat, ec.Element.Glimpse);
                     UpdateUI();
                 }
             }
--- a/src/AccessibilityInsights.SharedUx/Enums/FileFilters.cs
+++ b/src/AccessibilityInsights.SharedUx/Enums/FileFilters.cs
@@ -7,11 +7,11 @@ namespace AccessibilityInsights.SharedUx.Enums
     /// </summary>
     public static class FileFilters
     {
-        public const string A11yFileFilter = "AccessibilityInsights files (*.a11ytest,*.a11yevent)|*.a11ytest;*.a11yevent|A11y Test files (*.a11ytest)|*.a11ytest|A11y Events files (*.a11yevent)|*.a11yevent|All files (*.*)|*.*";
-        public const string TestFileFilter = "AccessibilityInsights Test files (*.a11ytest)|*.a11ytest|All files (*.*)|*.*";
-        public const string EventsFileFilter = "AccessibilityInsights Events files (*.a11yevent)|*.a11yevent|All files (*.*)|*.*";
-        public const string TestExtension = ".a11ytest";
-        public const string EventsExtension = ".a11yevent";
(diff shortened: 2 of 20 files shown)


src/AccessibilityInsights.SharedUx/ViewModels/HierarchyNodeViewModel.cs
```diff
@@ -301,10 +302,8 @@ public HierarchyNodeViewModel(A11yElement ce, bool showUncertain, bool isLiveMod
         /// <param name="isLiveMode">Is node in live mode</param>
         private void UpdateIconInfoAndAutomationName(bool isLiveMode)
         {
-            StringBuilder sb = new StringBuilder();
-            sb.Append(this.Element.Glimpse);
-
             this.IconVisibility = Visibility.Collapsed;
+            string automationNameFormat = Resources.HierarchyNodeViewModel_AutomationNameDefaultFormat;
```
DaveTryon, line 306: The first attempt here set this to null, which then caused an error in live mode. The revised default is just {0}

src/AccessibilityInsights.SharedUx/Controls/TestTabs/AutomatedChecksControl.xaml.cs
```diff
@@ -23,10 +23,8 @@
 using System.Windows.Controls;
 using System.Windows.Controls.Primitives;
 using System.Windows.Data;
-using System.Windows.Documents;
```
DaveTryon, line 26: The 2 using blocks were were unused before this change--this just cleans them up

src/AccessibilityInsights.SharedUx/ViewModels/BaseActionViewModel.cs
```diff
@@ -1,11 +1,9 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 using AccessibilityInsights.CommonUxComponents.Dialogs;
-using AccessibilityInsights.SharedUx.Dialogs;
```
DaveTryon, line 4: These using blocks were already unneeded--I just cleaned them up while I was in the file

src/AccessibilityInsights.SharedUx/ViewModels/EventConfigNodeViewModel.cs
```diff
@@ -10,9 +10,9 @@
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.ComponentModel;
+using System.Globalization;
 using System.Linq;
 using System.Windows;
-using System.Windows.Input.StylusPlugIns;
```
DaveTryon, line 15: Opportunistic cleanup

src/AccessibilityInsights.SharedUx/Properties/Resources.resx
```diff
@@ -1551,4 +1551,7 @@ Do you want to change the channel? </value>
   <data name="ApplicationSettingsControl_TelemetryDisabledByAdministrator" xml:space="preserve">
     <value>Telemetry has been disabled by an administrator of this computer.</value>
   </data>
+  <data name="A11yElementView_ExecutionError" xml:space="preserve">
+    <value>Execution result failed!</value>
```
jalkire, line 1558: I don't see the ! in the original string--was that an intentional addition?

src/AccessibilityInsights.SharedUx/Enums/FileFilters.cs
```diff
@@ -7,11 +7,11 @@ namespace AccessibilityInsights.SharedUx.Enums
     /// </summary>
     public static class FileFilters
     {
-        public const string A11yFileFilter = "AccessibilityInsights files (*.a11ytest,*.a11yevent)|*.a11ytest;*.a11yevent|A11y Test files (*.a11ytest)|*.a11ytest|A11y Events files (*.a11yevent)|*.a11yevent|All files (*.*)|*.*";
-        public const string TestFileFilter = "AccessibilityInsights Test files (*.a11ytest)|*.a11ytest|All files (*.*)|*.*";
-        public const string EventsFileFilter = "AccessibilityInsights Events files (*.a11yevent)|*.a11yevent|All files (*.*)|*.*";
-        public const string TestExtension = ".a11ytest";
-        public const string EventsExtension = ".a11yevent";
```
jalkire, line 14: Is it the right move to pull these out into resource files? It seems like even if we localized, the file extensions themselves would remain unchanged.

jalkire (approved):
Approved with a couple minor suggestions

src/AccessibilityInsights.SharedUx/Properties/Resources.resx
DaveTryon: This was a typo--thanks for catching. Fixed

Issue #1290, opened 2022-02-15
[BUG] Crash in beaker overlay rendering when trying to use Shift+F9 to focus any app


dbjorge:
**Describe the bug**
A user attempting to use Accessibility Insights for Windows against a prerelease Windows Insider build is encountering an issue where Accessibility Insights crashes reliably when attempting to use a keyboard shortcut (shift+F9) to invoke a scan against a focused app.

The crash occurs as gdiplus attempts to load one of our font files to render the "beaker" icon we overlay over the focused app. The crash can be worked around by going to the Accessibility Insights settings and switching the "Highlight Mode" setting to one which doesn't include the beaker icon.

The earliest reported build that hits the issue was `rs_prerelease 22533.1001.220107-2122`, which became the public Insider dev channel build on 1/12/22 (the issue was reported internally earlier than this writing, but we deferred filing a public issue until the build became public). However, I was *not* able to reproduce the crash in a VM running the same build. The reporting user is a JAWS user, but we have no specific reason to believe that this is related to the crash. The user reported that uninstalling, re-downloading, and reinstalling Accessibility Insights did not prevent the crash.

There is a dump of the crash available internally; employees can refer to our internal DRI notes for details.

**To Reproduce**
1. Use the above Windows Insider build
2. ??? (unknown setup, no local repro known yet)
3. Open Accessibility Insights
4. Under Settings, ensure that Highlight Mode is set to something that includes "Beaker"
5. Use the keyboard to focus on any other app (eg, Notepad)
6. Use Shift+F9 to activate Accessibility Insights
7. Observe crash

**Expected behavior**
Accessibility Insights doesn't crash.

**Actual behavior**
Accessibility Insights does crash.

**Screenshots or .GIF**
n/a

**Desktop (please complete the following information):**
 - OS: `rs_prerelease 22533.1001.220107-2122`
 - Accessibility Insights for Windows Version: 1.1.1741.1 (production)
 - Target Application: Any (eg, Notepad)
 - Target Application Version: n/a

**Additional context**
Priority requested - medium (reporting user has a workaround, but app crash is a very serious issue)

dbjorge:
We believe this is resolved by #1291, but since we still don't have a local repro, I'm going to reach out to the reporting user to verify that the fix as-merged resolves the issue on their end before closing the issue.

dbjorge:
The reporting user was able to verify that the fix as-implemented in Canary still works to prevent the crash.

PR #1291, opened 2022-02-17
fix: prevent crash in Highlighter beaker rendering due to FontFamily use-after-dispose

dbjorge:
#### Details

This PR fixes issue #1290, where one user reported an issue where Application Insights would crash while trying to render the "beaker" icon as part of its highlighter overlay.

The root cause of the crash is that the code path in question was rendering the icon using a `FontFamily` pulled out of a `PrivateFontCollection` which we had already `Dispose()`ed. The `Families` getter in the `PrivateFontCollection` doesn't document the lifetime requirements of the `FontFamily` objects it returns and its implementation is misleading in that it *looks* like it's cloning a copy of the underlying native font objects with a separate lifetime, but the native clone still maintains some links to data backed by the font collection, so it isn't safe to use it this way in practice.

##### Motivation

Addresses issue #1290

##### Context

I considered a more contained version of this fix which just expanded the scope of the `using` block that we were previously using to control the lifetime of the `PrivateFontCollection` to the end of the function it's used in. I went with the approach here instead because it solves the same problem but also has the benefit of moving the slightly expensive "load and parse the font" operation out of the middle of a rendering/window message handling path and into a "only once during startup" path instead, which results in a marginally more fluid user experience.

I also included a change here that changes an existing exception handling path to report any exceptions that occur while loading the font, per discussion last month with @DaveTryon. This didn't turn out to be relevant to this bug, but I kept it since it seemed like a good change to make.

We don't have a local repro of the issue to test against, but I did send an earlier version of this PR to the originally reporting user to verify, and they confirmed that it fixed the issue. Once I've addressed any initial review comments, I'll send that user a copy of the final candidate fix to double check that the fix still works as expected.

#### Pull request checklist

- [x] Does this address an existing issue? yes, #1290
- [no] Includes UI changes?

### DIFF
--- a/src/AccessibilityInsights.SharedUx/Highlighting/Win32SnapshotButton.cs
+++ b/src/AccessibilityInsights.SharedUx/Highlighting/Win32SnapshotButton.cs
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 using AccessibilityInsights.CommonUxComponents.Utilities;
+using AccessibilityInsights.Extensions.Helpers;
 using AccessibilityInsights.SharedUx.Utilities;
 using AccessibilityInsights.Win32;
 using System;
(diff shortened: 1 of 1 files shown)


peterdur (approved):
Read, built, and ran in debugger. Verified family/fonts not disposed until app closed or Windows theme changed. LGTM!

PR #1302, opened 2022-03-18
feat: Add local (disk-based) telemetry

DaveTryon:
#### Details

We've discussed the possibility of adding local telemetry to AIWin. This PR proposes an approach that we can use for discussion, if nothing else. Basically, it adds a new implementation of the `ITelemetry` interface and allows the MEF container to handle multiple targets. It updated the `TelemetrySink` class to understand how to send data to all implementations of `ITelemetry`, then it adds a disk-based logger and unit tests. I've attached a ZIP file [AccessibilityInsightsLocalTelemetry.zip](https://github.com/microsoft/accessibility-insights-windows/files/8308112/AccessibilityInsightsLocalTelemetry.zip) with a redacted version of the file that gets written. Since this file can get fairly large, I opted to truncate the file at the beginning on each app session. This will have issues if multiple instances of the app run concurrently, but it's possible to fix this without too much difficulty if we choose to do so.

Logging goes to a file at `%temp%\AccessibilityInsightsLocalTelemetry.txt`, but only if all these conditions are true:
1. This is a non-MSI build. This includes dev builds, PR builds, CI builds, etc. It's a trivial change to [product.wxs](https://github.com/microsoft/accessibility-insights-windows/blob/main/src/MSI/Product.wxs) if we want to include this in MSI builds
2. Group policy has not disabled telemetry
3. The user has opted into telemetry

The file format is fairly arbitrary and driven by what seemed reasonable. It's easy to change. It's intended to be human-readable, not machine-readable.

##### Motivation

We've discussed this ability but never had the time to really pursue it. I had some free cycles, so I decided to make a first stab. Good or bad, at least it's a data point about some of the benefits or issues. It also makes it easy to catch potential issues. (As an aside, I've confirmed that Application Insights scrubs paths from file names in stack traces--this code doesn't yet do that, although it would be fairly easy to add)

##### Context

As mentioned above, if we decide that we like this, we should probably update [product.wxs](https://github.com/microsoft/accessibility-insights-windows/blob/main/src/MSI/Product.wxs) to include it in all builds so that our various environments are consistent.

My initial thought was to have this independently controlled from the regular telemetry channels. That would be difficult without changing how we handle the group policy and opt-in, and then it occurred to me that this also has the advantage of allowing for data to be easily searched in the data sent to Application Insights. That's also why I decided to keep the time in UTC format instead of local time zone.

#### Pull request checklist

- [n/a] Includes UI changes?


src/AccessibilityInsights.Extensions.DiskLoggingTelemetry/ReportedExceptionData.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+
+namespace AccessibilityInsights.Extensions.DiskLoggingTelemetry
+{
+    public class ReportedExceptionData: PublishBaseData
```
karanbirsingh, line 8: (nit) a space might be nice here for consistency

karanbirsingh:
Verified that this is robust to external exclusive locks on the log file

PR #1303, opened 2022-03-24
chore: Run automatic cleanup tool

DaveTryon:
#### Details

In #1302, @karanbirsingh pointed out that the spacing around an interface implementation was slightly off. After fixing it, I wondered how many other places had similar issues. The answer is "a lot". I also noticed that we have many places where we include unnecessary `using` statements. Rather than try to fix them by hand, I found that Microsoft publishes a Visual Studio extension called [Power Commands for Visual Studio](https://marketplace.visualstudio.com/items?itemName=VisualStudioPlatformTeam.PowerCommandsforVisualStudio) that can apply normalizations across an entire project from a single command. I ran the extension using the default set of fixers, which are:

- Format document (includes things like consistent spacing, braces, etc.)
- Remove unnecessary usings
- Sort usings
- Apply file header preferences

The last 2 fixers did nothing to our code, since we've previously sorted using blocks and have no configured file header preferences. The first 2 fixers, though, found and fixed a lot more issues than I'd found manually.

This PR is fairly large (nearly 20% of our files), but the individual changes are trivial--the vast majority of them are tiny whitespace tweaks that just got missed in previous changes. I've reviewed them, but it would be good to get a second reviewer

##### Motivation

Code consistency

#### Pull request checklist

- [n/a] Includes UI changes?

### DIFF
--- a/src/AccessibilityInsights.Extensions.AzureDevOps/Enums/AzureDevOpsField.cs
+++ b/src/AccessibilityInsights.Extensions.AzureDevOps/Enums/AzureDevOpsField.cs
@@ -1,6 +1,5 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
-using System;
 
 namespace AccessibilityInsights.Extensions.AzureDevOps.Enums
 {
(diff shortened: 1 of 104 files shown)


src/AccessibilityInsights.Extensions.AzureDevOps/Enums/AzureDevOpsField.cs
```diff
@@ -31,11 +30,11 @@ public static string ToApiString(this AzureDevOpsField value)
             switch (value)
             {
                 // These fields are all documented at https://docs.microsoft.com/en-us/rest/api/azure/devops/wit/Revisions/List?view=azure-devops-rest-5.0
-                case AzureDevOpsField.Title:           return "System.Title";
```
DaveTryon, line 34: The original version might be slightly more readable due to the column alignment, but this seems OK

src/AccessibilityInsights.SharedUx/Controls/ScannerResultControl.xaml.cs
```diff
@@ -123,9 +123,9 @@ private void SetScannerResultTreeView(A11yElement e)
         private void UpdateTree()
         {
             IEnumerable<ScanListViewItemViewModel> itemViewModel = from l in this.List
```
DaveTryon, line 125: This goes really far to the right--I might introduce a local variable just to move everything to the left

DaveTryon: This was already a local variable. I changed the type to `var` to reduce the indent a little bit

jalkire:
Should we add a note about recommending Power Commands for Visual Studio to our readme, contributing guide, or some other project doc? Also, is there a way we can enforce its checks during build? (doing so wouldn't need to be a part of this PR of course)

jalkire (approved):
Approved with one suggestion

DaveTryon:
From @jalkire:
> Should we add a note about recommending Power Commands for Visual Studio to our readme, contributing guide, or some other project doc? Also, is there a way we can enforce its checks during build? (doing so wouldn't need to be a part of this PR of course)

Newer versions of VS (which we get in the PR builds) support [EditorConfig files](https://docs.microsoft.com/en-us/visualstudio/ide/create-portable-custom-editor-options?view=vs-2019), which can provide enforcement and even (I think) format on save support. Let's try to add EditorConfig in a future PR

PR #1353, opened 2022-06-07
feat(framework-issues-ui): Refactor AutomatedChecksControl

DaveTryon:
#### Details

This PR splits `AutomatedChecksControl` into 2 classes--`AutomatedChecksControl` and `AutomatedChecksCustomListControl`. An additional class, `AutomatedChecksCustomListControlContext`, was added to encapsulate data that needs to be shared between the classes. The intent is to leave existing functionality precisely as it is, but to make it possible for `AutomatedChecksControl` to end up with 2 instances of `AutomatedChecksCustomListControl`. The first instance will hold rule failures that have no associated framework issue link. The second instance will hold rule failures that have associated framework links.

The previous code sets the AccessibleName property of the DataGrid by specifying a label. I've updated this to set the property via a parameter that gets specified in code instead of XAML.

##### Motivation

Feature work

##### Context

The bug filing code and the link handling code in this class are almost identical to what's in `ScannerResultControl`, which will also be refactored. Time permitting, a later PR may consolidate the common code from the 2 classes.

We considered adding a ViewModel here but decided that we'd settle for just splitting the code.

#### Pull request checklist

- [x] Does this address an existing issue? If yes, Issue [1956683](https://mseng.visualstudio.com/1ES/_workitems/edit/1956683)
- [n/a] Includes UI changes?


src/AccessibilityInsights.SharedUx/Controls/CustomControls/AutomatedChecksCustomListControl.xaml.cs
```diff
@@ -0,0 +1,689 @@
...
+            InitializeComponent();
+            AddHandler(Thumb.DragDeltaEvent, new DragDeltaEventHandler(Thumb_DragDelta), true);
+        }
+
+        internal void SetControlContext(AutomatedChecksCustomListControlContext controlContext)
+        {
+            _controlContext = controlContext ?? throw new ArgumentNullException(nameof(controlContext));
+
+            CheckBoxSelectAll.IsEnabled = ScreenshotAvailable;
+            AutomationProperties.SetName(lvResults, controlContext.DataGridAccessibleName);
```
jalkire, line 67: One way to do this in XAML would be to create a `DataGridAccessibleName` `DependencyProperty` in this class. You could then set `DataGridAccessibleName` in AutomatedChecksControl.xaml when you declare this control. A simple example of this pattern would be [`PlaceHolderTextBox`](https://github.com/microsoft/accessibility-insights-windows/blob/main/src/AccessibilityInsights.CommonUxComponents/Controls/PlaceholderTextBox.xaml.cs)--the property `Placeholder` is declared in the codebehind, [bound to in XAML (in the control's style in this case)](https://github.com/microsoft/accessibility-insights-windows/blob/f7b4c82e34341137333397731ffa9f80a4d29e74/src/AccessibilityInsights.SharedUx/Resources/Styles.xaml#L1557), and then set when [the control is used](https://github.com/microsoft/accessibility-insights-windows/blob/f7b4c82e34341137333397731ffa9f80a4d29e74/src/AccessibilityInsights.SharedUx/Controls/PropertyInfoControl.xaml#L37)

jalkire (approved):
Looks like mostly the same code as before + some improvements 👍 
One suggestion on using a DependencyProperty, but I'd call that optional

PR #1354, opened 2022-06-09
feat(framework-issues-ui): Add framework issues in AutomatedChecks view

DaveTryon:
#### Details

This exposes the framework-related failures in AutomatedTests view. This consists of:
1. Add `AutomatedChecksCustomListControl.SectionHeader` to allow the XAML to control the presence or absence of a header above the `ListView`. The header is visible only when a SectionHeader is specified, otherwise it is collapsed.
2. Since the `AutomatedChecksCustomListControl` now has more than 1 element, I adjusted the code to hide the entire control instead of just the `ListView`
3. Add a second `AutomatedChecksCustomListControl` to `AutomatedChecksControl`. Update the names to make it clear that one is for framework-related issues and one is for non-framework-related issues
4. Update `AutomatedChecksControl` to appropriately filter results based on the `RuleResult.FrameworkIssueLink` property
5. Ensure that `AutomatedChecksControl` always updates both controls correctly. These just duplicate existing code--I thought about some sort of delegate that would loop through a set of controls but decided that it wasn't worth it when we have no plans to have more than 2 instances of the control
6. Wrap the `AutomatedChecksCustomListControl` controls in a StackPanel inside a ScrollViewer, so that the results can be accessed even when they extend beyond the size of the pane. This introduces a subtle visual difference--see screenshots below.
7. In #1353, we failed to properly set the `AutomationId` property on the `ListView` (it was incorrectly being set on the parent). This is now fixed and the value is set for both `ListView` controls
8. Updating the UITests to account for the new behavior

##### Motivation

Feature work

##### Screenshots
Note that the cases marked with an asterisk were simulated by forcing the code to assign all results to a single control. The rules themselves are not representative of actual framework issues

Description | Screenshot
--- | ---
No errors found | image
Framework errors only* | image
Non-framework errors only* | image
Framework and non-framework errors | image
Framework and non-framework errors with scrollbar | image

__Subtle visual differences__
The existing code always displays the header row (including the check/uncheck all control), even when scrolling way down the list. The new code allows the header row to move within the scroll window--this was a compromise to make the overall experience smooth and to avoid nested scrollbars when there are 2 ListView controls. We can probably improve this design at a future date, but I took the approach of making the MVP simple for now. These screenshots show the display after you scroll to the end of a list that exceeds the app size

Description | Screenshot
--- | ---
Current UX | image
New UX | image

#### Pull request checklist

- [x] Does this address an existing issue? If yes, Issue" [1956688](https://mseng.visualstudio.com/1ES/_workitems/edit/1956688)
- [x] Includes UI changes?


src/AccessibilityInsights.SharedUx/Controls/TestTabs/AutomatedChecksControl.xaml
```diff
@@ -59,10 +59,18 @@
                 </Grid>
             </Grid>
             <Border Grid.Row="1" BorderBrush="{DynamicResource ResourceKey=TabBorderBrush}" BorderThickness="0,1,0,0" VerticalAlignment="Stretch" HorizontalAlignment="Stretch">
-                <local:AutomatedChecksCustomListControl KeyboardNavigation.TabNavigation="Once" DataGridAccessibleName="{x:Static Properties:Resources.lblTitleContent}"
-                          x:Name="listControl" HorizontalAlignment="Stretch" VerticalAlignment="Stretch" BorderThickness="0"
-                          AutomationProperties.AutomationId="{x:Static Properties:AutomationIDs.AutomatedChecksResultsListView}" Background="Transparent">
-                </local:AutomatedChecksCustomListControl>
+                <ScrollViewer HorizontalScrollBarVisibility ="Disabled" VerticalScrollBarVisibility="Auto">
```
jalkire, line 62: Maybe this is just part of the design, but it looks off to me that the two lists independently scroll horizontally but scroll vertically together--why not enable horizontal scroll here as well?
screenshot showing the framework and normal failures each scrolling horizontally separately while both scroll vertically together

jalkire:
This could come in a separate PR, but the LiveMode and LoadTestFile UI tests should likely be updated to reflect these changes (the former is broken but doesn't run in our pipeline due to flakiness, the latter works because the loaded test file contains no framework issues)

src/AccessibilityInsights.SharedUx/Controls/TestTabs/AutomatedChecksControl.xaml
jalkire: Similarly, now the horizontal scroll bar no longer sticks to the bottom of the app and instead moves around with the listview (before on left, after on right):
screenshot of ai-win automated checks view showing horizontal scroll at the bottom of the page before and in the middle of the page under the listview on the right
Maybe we're ok with this change, just want to make sure.

jalkire (approved):

DaveTryon:
@jalkire, since @nang4ally is out today, I'm going to merge this to get the refactored UITests into place so that we have it ready for the upcoming changes in the ScannerResultsControl. It should be a relatively small change if we later decide to alter the scrollbar behavior

Issue #1377, opened 2022-06-30
[BUG] Reference to Unofficial.Microsoft.mshtml


drasticactions:
Hey folks,

I was going through the third-party notices for this app and saw that "Unofficial.Microsoft.mshtml" is being referenced.

![MicrosoftTeams-image](https://user-images.githubusercontent.com/898335/176783939-f129a00c-e4ad-4d30-8ee6-c6ade150de09.png)

https://github.com/microsoft/accessibility-insights-windows/blob/main/src/AccessibilityInsights.Extensions.AzureDevOps/Extensions.AzureDevOps.csproj#L39

This library (https://github.com/ogaudefroy/Unofficial.Microsoft.mshtml) is an "Unofficial" repack of "Microsoft.mshtml.dll", a Microsoft-built library. I believe it's the original IE rendering engine? I'm confused why a Microsoft branded product would need to use an "unofficial" nuget? It's also marked as MIT licensed, which (IMO) is kinda silly, as that person can make no such claims to it. They don't own the original library.

Is this actively being used? Could this be a vestige of something earlier? It was inserted two years ago.

DaveTryon:
@drasticactions, we've merged a fix for this. It will appear in Canary build 1.1.2014.1, due to be published at https://github.com/microsoft/accessibility-insights-windows/releases/tag/v1.1.2014.01 on July 8th. Please regress the issue once the release is available and close out at your convenience. Thanks!

PR #1384, opened 2022-07-05
fix: Remove dependency on legacy assembly

DaveTryon:
#### Details

As called out in #1377, the ADO issue filing code has a dependency on an assembly that Microsoft no longer supports but that someone has published to NuGet with a license that is probably incorrect. This code gets used to modify the HTML that is used to attach a screenshot to the issue in ADO.

This PR replaces the code that uses the old assembly with code that uses the HtmlAgilityPack [package](https://www.nuget.org/packages/HtmlAgilityPack), which is _genuinely_ MIT licensed, is actively maintained, and is fully open source at [this GitHub repo](https://github.com/zzzprojects/html-agility-pack/).

In addition to minor API differences, the HTML produced is different in 3 ways:
1. The old library forced all HTML tags to upper case; the new library doesn't do this. HTML is case-insensitive so we don't try to reproduce this
2. The old library inserted `\r\n` in various places in the output stream; the new library doesn't do this. These characters might make the output a little easier for humans to parse, but isn't needed for our usage. We won't try to reproduce this.
3. The old library wrapped its output in a `<BODY>` tag; the new library doesn't do this. For API compatibility, we add this explicitly

Tested by creating new ADO issues with the modified code and confirming that the end result matches expectations.

The PR also updates the file set in the WXS file and makes sure that the 3rd party file gets properly signed

##### Context

In checking the WXS file, I noticed some files that should probably be in the WXS file but are not. Looking at the [csproj file](https://github.com/zzzprojects/html-agility-pack/blob/master/src/HtmlAgilityPack.Net45/HtmlAgilityPack.Net45.csproj), none of these assemblies are used by `HtmlAgilityPack.dll`, so I'll add those file references in a separate PR.

#### Pull request checklist

- [x] Does this address an existing issue? If yes, Issue: #1377
- [n/a] Includes UI changes?

### DIFF
--- a/src/AccessibilityInsights/AccessibilityInsights.csproj
+++ b/src/AccessibilityInsights/AccessibilityInsights.csproj
@@ -13,6 +13,7 @@
   <ItemGroup>
     <Drop3PartySignedFile Include="$(TargetDir)\Ben.Demystifier.dll" />
     <Drop3PartySignedFile Include="$(TargetDir)\CommandLine.dll" />
+    <Drop3PartySignedFile Include="$(TargetDir)\HtmlAgilityPack.dll" />
     <Drop3PartySignedFile Include="$(TargetDir)\Microsoft.Deployment.WindowsInstaller.dll" />
     <Drop3PartySignedFile Include="$(TargetDir)\Newtonsoft.Json.dll" />
   </ItemGroup>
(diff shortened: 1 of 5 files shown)


src/AccessibilityInsights/AccessibilityInsights.csproj
```diff
@@ -13,6 +13,7 @@
   <ItemGroup>
     <Drop3PartySignedFile Include="$(TargetDir)\Ben.Demystifier.dll" />
     <Drop3PartySignedFile Include="$(TargetDir)\CommandLine.dll" />
+    <Drop3PartySignedFile Include="$(TargetDir)\HtmlAgilityPack.dll" />
```
DaveTryon, line 16: This is what triggers signing of HtmlAgilityPack.dll

PR #1386, opened 2022-07-06
chore: Update wxs file to match package contents

DaveTryon:
#### Details

Our WXS file has not been kept current when package contents change. This PR just updates the WXS file to match the list of assemblies in `./src/AccessibilityInsights/bin/release/net472`, with the intentional omission of the local disk logging assembly.

##### Motivation

Keep the code up to date

##### Context

Most of these changes probably came from dependabot updates. I wonder if we should consider having a pipeline check that intelligently compares the WXS file with the contents of the folder, then fails if they get out of sync. I'll add this to my list of things to do in slack time...

#### Pull request checklist

- [n/a] Includes UI changes?


DaveTryon:
> Is it possible we don't need to ship all of the files we currently package?
> 
> In the meantime, I support keeping the package in line with the bin folder so I'm approving this

I'm certain that some assemblies (like the API entry points) aren't fully needed. I imagine that the System.* files are probably getting resolved to something else. I'm taking the approach of defaulting to keeping them in sync unless we have an explicit reason to make them diverge.

PR #1388, opened 2022-07-07
chore: Add unit test to keep WXS file in sync with package updates

DaveTryon:
#### Details

Add a unit test that will fail if a new file appears in the drop folder without a corresponding entry in either the WXS file or as an entry in a list of files that we _intentionally_ exclude from the WXS file. Since the MSI file builds only in release builds, the unit tests also run only in release builds.

##### Motivation

We recently noticed (see #1386) that the set of files in our WXS file differed unexpectedly from the set of files that existed when running locally. This apparently happened as new assemblies were added in some package updates. We want to be able to detect this case during the PR that upgrades a package version, so that our MSI and non-MSI installs behave the same way.

#### Pull request checklist

- [n/a] Does this address an existing issue? If yes, Issue
- [n/a] Includes UI changes?


src/MsiFileTests/WxsValidationTests.cs
```diff
@@ -0,0 +1,104 @@
...
+        {
+            HashSet<string> filesInSection = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+
+            using (XmlReader reader = XmlReader.Create(wxsFile))
+            {
+                bool inCorrectSection = false;
+
+                while (reader.Read())
+                {
+                    switch (reader.NodeType)
```
DaveTryon, line 77: This might be better without the switch--I'll update it and evaluate

src/MsiFileTests/WxsValidationTests.cs
```diff
@@ -0,0 +1,96 @@
...
+            string wxsFile = Path.GetFullPath(Path.Combine(repoRoot, @"MSI\Product.wxs"));
+            HashSet<string> productComponentExclusions = new HashSet<string>(StringComparer.OrdinalIgnoreCase)
+            {
+                "AccessibilityInsights.Extensions.DiskLoggingTelemetry.dll",
+                "Axe.Windows.Automation.dll",
+            };
+
+            CompareWxsSectionToDropPath(repoRoot, @"AccessibilityInsights\bin\release\net472", wxsFile, "ProductComponent", productComponentExclusions);
+            CompareWxsSectionToDropPath(repoRoot, @"AccessibilityInsights\bin\release\net472\IssueTemplates", wxsFile, "IssueTemplates");
+            CompareWxsSectionToDropPath(repoRoot, @"AccessibilityInsights.VersionSwitcher\bin\release\net472", wxsFile, "VersionSwitcher");
```
jalkire, line 33: One potential drawback to this approach is that if we hypothetically add a new directory to the drop, we'll have to remember to add it here or else its files will go unvalidated. I could imagine some kind of logic where we ensure that every `Directory` node in the wxs file has been compared with a directory (such as a comparison between the total count of `Directory` nodes under `INSTALLFOLDER` vs the number of `GetFilesIncludedInWxsComponent` calls), but that might be excessive given that we almost never add new directories

jalkire (changes requested):
This is great! One comment with an idea for further validation, but I might be overthinking things 🙂 The other comment relates to the failing PR build

src/MsiFileTests/WxsValidationTests.cs
```diff
@@ -0,0 +1,96 @@
...
+            string currentFolder = Path.GetFullPath(Assembly.GetExecutingAssembly().Location);
+            string repoRoot = Path.GetFullPath(Path.Combine(currentFolder, @"..\..\..\..\.."));
+            string wxsFile = Path.GetFullPath(Path.Combine(repoRoot, @"MSI\Product.wxs"));
+            HashSet<string> productComponentExclusions = new HashSet<string>(StringComparer.OrdinalIgnoreCase)
+            {
+                "AccessibilityInsights.Extensions.DiskLoggingTelemetry.dll",
+                "Axe.Windows.Automation.dll",
+            };
+
+            CompareWxsSectionToDropPath(repoRoot, @"AccessibilityInsights\bin\release\net472", wxsFile, "ProductComponent", productComponentExclusions);
```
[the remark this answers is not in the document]
DaveTryon: These tests are skipped in debug

jalkire (approved):

Issue #1398, opened 2022-08-01
[BUG] AADSTS90015: Requested query string is too long.


NCarlsonMSFT:
Please check whether the bug has [already been filed](https://github.com/Microsoft/accessibility-insights-windows/issues).

This repo contains code for Accessibility Insights for Windows. We maintain code for its automated checks rule engine at [axe-windows](https://github.com/microsoft/axe-windows/issues/). Most false positive questions or rule clarifications belong in [axe-windows](https://github.com/microsoft/axe-windows/issues/new/choose), whereas questions about the Accessibility Insights tool UI should go here. If you aren't sure, it's definitely OK to use this repo. Thanks for contributing!

**Describe the bug**
When trying to use the File issue feature against ADO creating the issue fails with "AADSTS90015: Requested query string is too long."

**To Reproduce**
1. Connect to an Azure Board for filing issues
2. Open an a11ytest file with an issue
3. Click "File Issue:

**Expected behavior**
A new bug is created in the ADO board

**Actual behavior**
Authentication fails

**Screenshots or .GIF**
![image](https://user-images.githubusercontent.com/14239288/182260326-28796412-43e5-4841-afa2-04994f62c967.png)

**Desktop (please complete the following information):**
 - OS: Windows 10 21H2 19044.1826
 - Accessibility Insights for Windows Version: 1.1.1997.01
 - Target Application: Visual Studio
 - Target Application Version: 17.4

**Additional context**
I've tried the mitigations recommended in https://github.com/microsoft/accessibility-insights-windows/issues/1336 to no avail.
Others on my team are being blocked by the same issue.

madalynrose:
Thank you for using Accessibility Insights for Windows, @NCarlsonMSFT! We have reproduced this locally and confirmed the bug you identified. We are working on investigating further and will respond here once we determine next steps.

madalynrose:
We are working on a more robust fix, but should someone find this issue because they are blocked, I can offer a very hacky temporary workaround:
1. right-click the dialog with the error message and select "Inspect" from the menu
2. select the console inside of DevTools and type `window.location.href = "https://dev.azure.com/YOUR-TEAM-HERE"`
3. close the dialog and retry filing the issue

The dialog seems to work okay once it is able to confirm authentication succeeded but the query parameters used to prepopulate the issue template make the url too long for it to do so.

DaveTryon:
[...]

PR #1400, opened 2022-08-04
fix: Ensure ADO file issue dialog can reach authenticated page

madalynrose:
#### Details

This PR updates the File Issue dialog's state machine to detect when an authentication redirect is taking place and confirm the dialog can reach the page without the query parameters before visiting the full pre-populated issue template.

##### Motivation

Addresses issue #1398 

##### Context

The error state in #1398 appears to happen because the `WebView2` dialog that opens during the File Issue flow does not load with the cached credentials from the ADO connection, causing the dialog to redirect to AAD endpoints as it logs in. These redirects extend the url and include the original query parameters in addition to new query parameters that allow for post-authentication redirection to the original page. These new urls are both past the 2000-character limit and prevent the dialog from ever reaching the original url. We found that once the dialog reaches an authenticated state it successfully opens the original url on subsequent attempts and that a shorter initial url will allow the dialog to successfully navigate the authentication redirects.

Because of this, we implemented a flow where we send dialog to a blank issue template if we detect auth redirects and then send it to the original url after we detect it has successfully reached the template.

#### Pull request checklist

- [ ] Includes UI changes?

### DIFF
--- a/src/AccessibilityInsights.Extensions.AzureDevOps/FileIssue/IssueFileForm.cs
+++ b/src/AccessibilityInsights.Extensions.AzureDevOps/FileIssue/IssueFileForm.cs
@@ -5,6 +5,7 @@ using Microsoft.Web.WebView2.Core;
 using Microsoft.Web.WebView2.WinForms;
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Windows.Forms;
@@ -22,6 +23,9 @@ namespace AccessibilityInsights.Extensions.AzureDevOps.FileIssue
         enum State
         {
             Initializing,
+            NeedsAuthentication,
+            Authenticating,
+            Authenticated,
             TemplateIsOpen,
             Saving,
             Saved,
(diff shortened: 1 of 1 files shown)


jalkire (approved):

src/AccessibilityInsights.Extensions.AzureDevOps/FileIssue/IssueFileForm.cs
```diff
@@ -110,8 +113,21 @@ private State updateState(bool revert = false)
                     case State.Initializing:
                         if (currentUri.Host == Url.Host && currentUri.AbsolutePath == Url.AbsolutePath)
                             _currentState = State.TemplateIsOpen;
+                        else
+                        {
```
DaveTryon, line 117: having braces around line 118 but not 115 seems unbalanced, somehow. Given the pattern used in this file, maybe remove the braces around line 118?

src/AccessibilityInsights.Extensions.AzureDevOps/FileIssue/IssueFileForm.cs
```diff
@@ -128,37 +144,46 @@ private State updateState(bool revert = false)
...
-
-                    // URL looks like "_queries/edit/2222222/..." where 2222222 is issue id
-                    // or is "_workitems/edit/2222222"
-                    // or is "_workitems?id=2222222"
-                    url = url.Substring(endIndex);
-                    int result;
-                    bool worked = int.TryParse(new String(url.TakeWhile(Char.IsDigit).ToArray()), out result);
-                    if (worked)
+                case State.Saving:
+
```
DaveTryon, line 150: Remove extra newline here?

DaveTryon:
@madalynrose, I love this change! I made a couple of tiny stylistic comments that you're free to pick up or ignore, as you see fit

DaveTryon (approved):

PR #1419, opened 2022-08-23
chore: Remove DisableDarkMode config setting

DaveTryon:
#### Details

Remove support for the `DisableDarkMode` config setting. This setting has no UI and required manual editing of the config file.

##### Motivation

When dark mode was introduced in #719, it was gated behind a non-UI config setting named `EnableDarkMode` so that people had to explicitly _opt in_ to use dark mode. When dark mode was enabled by default in #764, we provided a non-UI config setting named `DisableDarkMode` so that people had a way to _opt out_ from dark mode in AIWin. Both of these settings were included because we were concerned about dark mode bugs that could potentially block usability. That was 2 years ago and dark mode has matured a lot since then, so this config setting can now be removed.

#### Pull request checklist

- [ ] Includes UI changes?


dbjorge (approved):
LGTM!

PR #1422, opened 2022-08-24
chore: Embed symbols and SourceLink in released binaries

DaveTryon:
#### Details

Embed symbol (PDB) files into the shipping assemblies and enable SourceLink. This is similar to what we did for Axe.Windows in [#719](https://github.com/microsoft/axe-windows/pull/719) and [#727](https://github.com/microsoft/axe-windows/pull/727). In this case, though, we're moving directly to centrally managed settings, instead of first adding them in the projects, then centralizing them.

##### Motivation

A recent issue included a dump file in which the symbols did not resolve (because they aren't indexed). Looking at crash reports, the embedded symbols in Axe.Windows are getting recognized, so embedding symbols in AIWin should be helpful. It increases the file size by about 20%, but that's a small price to pay if we have better debugging support.

##### Context

I tested this by building a signed build from this branch, then installing it locally. I was able to attach VS to the running process and set function breakpoints, When the breakpoint triggered, VS used the SourceLink data to download the sources to a local cache and provide source-level debugging. While this shows that the information is available, we won't have proof that it works in the broader debugging ecosystem until we deploy it and watch for results in the crash reporting system.

#### Pull request checklist

- [n/a] Includes UI changes?

### DIFF
--- a/build/NetFrameworkRelease.targets
+++ b/build/NetFrameworkRelease.targets
@@ -39,7 +39,21 @@
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
+    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.1.1">
+      <PrivateAssets>all</PrivateAssets>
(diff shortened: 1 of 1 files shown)


dbjorge (approved):
This sounds reasonable to me! I agree with your reasoning from https://github.com/microsoft/axe-windows/pull/727 that the downside of slightly larger nuget package size is outweighed by the upside of removing debugging configuration steps.

Issue #1420, opened 2022-08-23
[BUG] Suggestion for requesting support for unsupported actions


DaveTryon:
**Is your feature request related to a problem? Please describe.**
When AIWin encounters a UIA Action that it doesn't know how to support, it displays a message that tells the user "if you want this action to be provided in this tool, please click below button", followed by a button that says "I want this Action!!!"

Screenshot showing this for the GetIAccessible action on the LegacyIAccessiblePattern:
![image](https://user-images.githubusercontent.com/45672944/186268662-3337e698-36ee-4d46-9ffa-680131de0481.png)

Clicking on the button does _absolutely nothing_. There is no user feedback, and the button handler does nothing. It once sent a telemetry data point, but that was removed as we prepared for our 2019 GA release.

**Describe the solution you'd like**
Instead of a do-nothing button, replace the text with a suggestion to open an issue requesting support for the missing action. Maybe something like: "Accessibility Insights for Windows does not yet support the _LegacyIAccessiblePattern.GetIAccessible_ action. To request support for this action, please upvote an existing issue at [https://github.com/microsoft/accessibility-insights-windows/issues](https://github.com/microsoft/accessibility-insights-windows/issues). If no issue already exists, you can open a new issue (feature request) to request support for this action."

This is better in several ways:
1. The user action is actually recorded
1. The user knows what is happening
1. The user has visibility into the status of the feature request
1. The feature request fits into our existing workflow
1. There is an opportunity to have a conversation with the user to understand the context of the feature request
1. There is a history of requests, allowing for upvoting or referring to previous requests

**Describe alternatives you've considered**
It would be simple to create a telemetry data point to send here, but that does very little. If the user has opted _out_ of telemetry, it does exactly nothing. Even if the user has opted _in_ to telemetry, it does nothing to cover points 2-6 listed in the proposed solution.

**Additional context**
It's less than a day of work to implement this change, but I've marked this as a feature request purely for workflow reasons.

DaveTryon:
This was released in Canary release https://github.com/microsoft/accessibility-insights-windows/releases/tag/v1.1.2115.01

PR #1448, opened 2022-10-12
fix: Provide effective way for user to request Pattern Action

DaveTryon:
#### Details

Revise the unsupported action dialog as outlined in #1420. This replaces a button that does nothing with directions of how to use AIWin's issues to request support for an unsupported pattern action. These dialogs are not themed in dark mode.

##### Screenshots
Mode | Before | After
--- | --- | ---
Light/Dark | image | image
Aquatic | image | image
Desert | image |  image
Dusk | image | image
Night Sky | image | image

##### Motivation

Address #1420, although style the action in bold instead of italic to make it stand out more

#### Pull request checklist

- [x] Does this address an existing issue? If yes, Issue #1420 
- [x] Includes UI changes?

### DIFF
--- a/src/AccessibilityInsights.SharedUx/Properties/Resources.resx
+++ b/src/AccessibilityInsights.SharedUx/Properties/Resources.resx
@@ -156,9 +156,6 @@
   <data name="GeneralActionView_ExecuteAction_Failed" xml:space="preserve">
     <value>Failed!!</value>
   </data>
-  <data name="UnsupportedActionPart1BeforeActionName" xml:space="preserve">
-    <value>Accessibility Insights for Windows does not yet support the</value>
-  </data>
   <data name="ReturnA11yElementsView_btnRun_Click_Input_should_be_int_value_greater_than_equal_to_0" xml:space="preserve">
     <value>Input should be int value greater than or equal to 0.</value>
   </data>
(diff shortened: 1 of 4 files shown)


src/AccessibilityInsights.SharedUx/Properties/Resources.resx
```diff
@@ -156,11 +156,8 @@
   <data name="GeneralActionView_ExecuteAction_Failed" xml:space="preserve">
     <value>Failed!!</value>
   </data>
-  <data name="ButtonContent" xml:space="preserve">
-    <value>I want this Action!!!</value>
-  </data>
-  <data name="TextBlockText" xml:space="preserve">
-    <value>The below Action is not supported yet. If you want this action to be provided in this tool. please click below button.</value>
+  <data name="UnsupportedActionPart1BeforeActionName" xml:space="preserve">
```
dbjorge, line 159: Can this use one format string rather than 3 separate data values, for better localizability in case the hyperlink and action name need to be in a different order? eg,

```xml
<data name="UnsupportedActionMessage" xml:space="preserve">
  <value>Accessibility Insights for Windows does not yet support the {0} action. To request support for this action, please upvote an existing issue at {1}. If no issue already exists, you can open a new issue (feature request) to request support for this action.</value>
  <comment>{0} is the programmatic ID of the action in question. {1} is a clickable hyperlink whose text is a URL where the user can file issues.</comment>
</data>
```

This would require the code-behind to do something a bit more complicated than the usual `string.Format` to fill in the format specifier, maybe something along the lines of this (rough, untested) code:

```cs
Regex formatStringPlaceholderRegex = new Regex(@"(\{\d})");
string[] formatStringParts = formatStringPlaceholderRegex.Split(formatStringFromResources, 2);
var messageRuns = formatStringParts.Select(part => {
  switch(part) {
    case "{0}": return new Run { Text = actionName; FontWeight = FontWeight.Bold; };
    case "{1}": return new Hyperlink { /* ... etc ... */ };
    default: return new Run { Text = part };
  }
});
this.unsupportedActionMessageTextBlock.Inlines.Clear();
this.unsupportedActionMessageTextBlock.Inlines.AddRange(messageRuns);
```

dbjorge (approved):
The UX change looks good to me! Left one localizability suggestion, but approving-with-comment - you have more context than I do about whether the suggestion is worth the extra complexity in AI4Win.

src/AccessibilityInsights.SharedUx/Properties/Resources.resx
DaveTryon: Thanks, @dbjorge! I was following a pattern that we use in a lot of places, but I prefer your pattern. Your "rough, untested" code was remarkably close to the final iteration. Hopefully we'll be able to use similar patterns in other places where we build up text runs via concatenation.

Issue #1438, opened 2022-09-12
[BUG] items in events data grid have UIA name "Axe.Windows.Desktop.UIAutomation.EventHandlers.EventMessage"


codeofdusk:
**Describe the bug**
Data items representing individual events captured by Accessibility Insights have the name `Axe.Windows.Desktop.UIAutomation.EventHandlers.EventMessage` in the UIA tree. This is read by screen readers as you navigate, making the event grid very hard to use.

**To Reproduce**
1. Start NVDA and Accessibility Insights.
2. Record events with the tool.
3. Navigate to the grid of recorded events.
4. Press down arrow.

**Expected behaviour**
NVDA reads a sane event description.

**Actual behaviour**
NVDA says "Axe.Windows.Desktop.UIAutomation.EventHandlers.EventMessage data item", followed by information about the event as children gain focus.

**Desktop (please complete the following information):**
 - OS: [e.g. Windows 10 1809] (Get the version by running `winver` from the command line) Windows 11 SV2 (22622.598)
 - Accessibility Insights for Windows Version: 06abd42
 - Target Application: [e.g. Chrome, Wildlife Manager] Windows Console (`conhost.exe`)
 - Target Application Version: [e.g. 22] inbox

DaveTryon:
Set name on element to "{Event Type} from {Event Source} at {Event Time}"

codeofdusk:
Yes, this seems like a reasonable reading order.

PR #1461, opened 2022-10-28
fix: Give items in events data grid readable UIA name

sfoslund:
#### Details

Address https://github.com/microsoft/accessibility-insights-windows/issues/1438 by implementing the suggestion to set name on element to "{Event Type} from {Event Source} at {Event Time}".

##### Motivation

Addresses issue https://github.com/microsoft/accessibility-insights-windows/issues/1438

#### Pull request checklist

- [x] Run through of all [test scenarios](https://github.com/Microsoft/accessibility-insights-windows/blob/main/docs/Scenarios.md) completed?
- [x] Does this address an existing issue? https://github.com/microsoft/accessibility-insights-windows/issues/1438
- [x] Includes UI changes?

### DIFF
--- a/src/AccessibilityInsights.SharedUx/Controls/EventRecordControl.xaml
+++ b/src/AccessibilityInsights.SharedUx/Controls/EventRecordControl.xaml
@@ -158,10 +158,16 @@
                       SelectionChanged="listEvents_SelectionChanged" SelectionMode="Single"
                       HorizontalContentAlignment="Stretch"
                       FontSize="11" Margin="2,0"
-                      Style="{StaticResource dgStyle}" CellStyle="{StaticResource dgcStyle}" RowStyle="{StaticResource dgrStyle}" ColumnHeaderStyle="{StaticResource dgchStyle}">
+                      Style="{StaticResource dgStyle}" CellStyle="{StaticResource dgcStyle}" ColumnHeaderStyle="{StaticResource dgchStyle}">
                 <i:Interaction.Behaviors>
                     <behaviors:ColumnResizeHotkeyBehavior/>
                 </i:Interaction.Behaviors>
+                <DataGrid.RowStyle>
+                    <Style BasedOn="{StaticResource dgrStyle}" TargetType="DataGridRow">
+                        <!-- Empty row name to override WPF default of using long data type -->
+                        <Setter Property="AutomationProperties.Name" Value=" "/>
+                    </Style>
+                </DataGrid.RowStyle>
                 <DataGrid.Columns>
                     <DataGridTextColumn Binding="{Binding Path=TimeStamp}" ElementStyle="{StaticResource DataGridTextCellStyle}" Header="{x:Static Properties:Resources.DataGridTextColumnHeaderTimeStamp}" Width="Auto" IsReadOnly="True" />
                     <DataGridTextColumn Binding="{Binding Path=EventId, Converter={StaticResource cvtIdToName}}" ElementStyle="{StaticResource DataGridTextCellStyle}" Header="{x:Static Properties:Resources.DataGridTextColumnHeaderEventName}" Width="Auto" IsReadOnly="True" />
--- a/src/AccessibilityInsights.SharedUx/Controls/EventRecordControl.xaml
+++ b/src/AccessibilityInsights.SharedUx/Controls/EventRecordControl.xaml
@@ -164,8 +164,7 @@
                 </i:Interaction.Behaviors>
                 <DataGrid.RowStyle>
                     <Style BasedOn="{StaticResource dgrStyle}" TargetType="DataGridRow">
-                        <!-- Empty row name to override WPF default of using long data type -->
-                        <Setter Property="AutomationProperties.Name" Value=" "/>
+                        <Setter Property="AutomationProperties.Name" Value="{Binding Path=EventId, Converter={StaticResource cvtIdToName}}"/>
                     </Style>
                 </DataGrid.RowStyle>
                 <DataGrid.Columns>


codeofdusk:
This definitely improves things!

However, the timestamp is still repeated after the full row is read out. NVDA log:

```
IO - speech.speech.speak (16:37:07.145) - MainThread (24588):
Speaking ['Events', 'data grid', CancellableSpeech (still valid)]
...
IO - speech.speech.speak (16:37:07.162) - MainThread (24588):
Speaking ["Text_TextChanged from terminal 'Command Prompt' at 16:35:17.014", 'data item', 'selected', '107 of 118', CancellableSpeech (still valid)]
...
IO - speech.speech.speak (16:37:07.184) - MainThread (24588):
Speaking ['16:35:17.014', '16:35:17.014', 'row 107', 'Time Stamp', 'column 1', CancellableSpeech (still valid)]
```

It looks like the row of the grid takes focus, then focus springs to its first cell (the timestamp), resulting in this repetition.

Also CCing @rperez030.

sfoslund:
> I believe that is the case when users are expected to move focus between individual cells, otherwise the row should receive focus.

I don't know the background on this, but given we're explicitly setting the focus to the cell instead of the row in our code, I assume that means that we expect users to move focus between individual cells (correct me if I'm wrong @DaveTryon). If that's the case, then it sounds like reading the content of the first cell in the row is expected.

rperez030:
If that is the case maybe the rows should have a more succinct name (E.G., the event name or the time stamp) so that is announced when focus moves from row to row or have no name at all so that only the element with focus is announced. I would be curious to understand why navigation between individual cells is allowed. Is it possible to interact with the content of individual cells somehow. Is it perhaps for easier navigation so that screen reader users can focus on one column and go down that column hearing just the value of that cell? That would be a good reason for giving the rows an empty name.

src/AccessibilityInsights.SharedUx/Controls/EventRecordControl.xaml
```diff
@@ -158,10 +158,23 @@
...
-                      Style="{StaticResource dgStyle}" CellStyle="{StaticResource dgcStyle}" RowStyle="{StaticResource dgrStyle}" ColumnHeaderStyle="{StaticResource dgchStyle}">
+                      Style="{StaticResource dgStyle}" CellStyle="{StaticResource dgcStyle}" ColumnHeaderStyle="{StaticResource dgchStyle}">
                 <i:Interaction.Behaviors>
                     <behaviors:ColumnResizeHotkeyBehavior/>
                 </i:Interaction.Behaviors>
+                <DataGrid.RowStyle>
+                    <Style BasedOn="{StaticResource dgrStyle}" TargetType="DataGridRow">
+                        <Setter Property="AutomationProperties.Name">
+                            <Setter.Value>
+                                <MultiBinding StringFormat="{}{0} from {1} at {2}">
```
DaveTryon, line 169: Can we move this string into a resource file somewhere? One option might be to bind to a property of a class (ViewModel if we have one), then build the string there

DaveTryon:
@sfoslund, this design has been unchanged for at least 5 years, and the "why" is probably lost to time. @codeofdusk and @rperez030 are both far more expert in screen reader use than I am, and if they have easy-to-implement suggestions that improve the screen reader experience, then I'm all for it.

I'm not sure if that conversation is in addition to this change or as a replacement of this change. It might make sense to merge this PR to get an improved experience, then revisit it if an even better experience is defined

sfoslund:
If we don't have a specific reason to allow focus on a specific cell, then I wonder if the right fix here is to disallow that such that the user can only focus on the full row, which will have the detailed name. @codeofdusk, @rperez030 does that make sense to you?

sfoslund:
@codeofdusk I've changed our focus logic to go to the full row instead of the first cell of that row and confirmed that this removed the redundancy in the name NVDA reads out. Let me know what you think.

codeofdusk:
@sfoslund After this change, while there is no more repetition, I'm no longer able to navigate through the grid.

rperez030:
I haven't checked out the PR to test, but what I would expect to happen is that the grid is treated like a list where the keyboard just moves from row to row and screen readers read the content of the row as we move with up and down arrow keys. If we want to explore an individual row, we will need to use object navigation, touch cursor or equivalent to review the content of individual cells. @codeofdusk  is that your experience? If we want Excel like navigation inside the grid, which may have some advantage depending on the use case, we have to give rows a name that makes sense because the name may be announce when focus moves from row to row, or give them an empty name which may raise an accessibility issue in a fast pass.

sfoslund:
Discussed this offline with @codeofdusk and we tossed around 3 main possibilities for the behavior on this:

1. Give rows full names ("event from event source at time") and only allow users to focus on rows (as opposed to individual cells). This removes the duplication issue but @codeofdusk didn't love the idea of disallowing focus on individual cells
2. Give rows full names and by default (using up and down keys) focus on full rows. If the user uses right and left keys, allow focus on individual cells. @codeofdusk didn't like that this continues to produce duplication when focus is on cells. This is the option that this PR is currently implementing.
3. Give rows empty names and always navigate between cells. This could potentially be confusing, since this would mean that when focus goes to the grid it will only read the timestamps, and you have to navigate further to discover the other data items in the grid. 

All 3 options should be reasonable to implement but we concluded that there are pros and cons to all them. Since he is the domain expert, @codeofdusk volunteered to think on this more/ ask around to come to a conclusion of which of these options (or any alternatives) would be ideal. Once a conclusion is reached, I'll update this PR with an implementation.

rperez030:
Option 3 may be improved by giving a meaningful name to the grid itself so that it will be read upon tabbing to it. The screen reader may announce the name (E.G. "recorded events"), the screen reader will indicate that it is a grid, and the first cell that receives focus will be read. The column headers are currently well defined, so the values with their corresponding labels will be read as the user navigates in an Excel like fashion through the grid with the arrow keys as long as the setting to read column headers for tables is enabled in the screen reader. I wouldn't expect it to be particularly confusing for the typical audience of an app like this one. there could also be something in the documentation that explains how to use the recording events function with keyboard / screen reader.

sfoslund:
After discussing offline with @codeofdusk, we decided the right move here is to give the rows an empty name and ensure that the table's name is meaningful. Since the table was already named `Events Data Grid`, I left that as is and this PR now just changes the row name to be empty.

rperez030:
I kind of expected that, and it is unfortunate in this case. It is because we are forcing a table to have row headers when the reality is that many tables don't have, and don't require, row headers.
I think it should be a very short string that adds value. the name of the event recorded in that row perhaps? @codeofdusk  what do you think?

codeofdusk (approved):
A touch more verbose than would be ideal, but it's far, far better than before, and we shouldn't let the perfect be the enemy of the good. LGTM!

PR #1486, opened 2022-11-18
chore: Be smarter about recompiling

DaveTryon:
#### Details

The `Win32` project generates a couple of files on the fly, and these files are then included in all of the non-test projects. This  works nicely for building but has an unfortunate side effect. These files get generated on _every compile_, which causes the _entire solution_ to recompile on _every compile_. Since a compile is on the order of 20 seconds, the inner debugging loop is slower than it ought to be, especially when you're working on unit tests. This PR does the following:
- Since the attributes are always the same, move them into a static source file that gets automatically included into every project that we ship.
- Only generate `A11yInsightsVersionInfo.cs` if it either does not exist or has changed, and only when building (not cleaning)

Validation scenarios:
- Build with a clean environment (ran `git clean -fdx` before starting VS). Build succeeds
- Manually edit the generated files and build again. Generated files get rebuilt
- Rebuild. Generated files do not get rebuilt
- Modify the .csproj file to simulate a different version, then Rebuild. Generated files get rebuilt

##### Motivation

Speed up inner dev loop

#### Pull request checklist

- [n/a] Includes UI changes?


src/AccessibilityInsights.Win32/Win32.csproj
```diff
@@ -14,12 +14,22 @@
...
+                      Lines="[assembly: System.Reflection.AssemblyVersion(&quot;$(BuildNumber)&quot;)] // Generated by build"
+                      Overwrite="True" />
   </Target>
 
-  <Target Name="CreateAttributesInfoFile">
-    <WriteLinesToFile File="$(MSBuildProjectDirectory)\..\bld\A11yInsightsAttributesInfo.cs" Overwrite="True" Lines="[assembly: System.CLSCompliant(false)] // Generated by build" />
+  <Target Name="CreateAttributesFile" BeforeTargets="BeforeCompile">
+    <WriteLinesToFile File="$(MSBuildProjectDirectory)\..\bld\A11yInsightsAttributesInfo.cs"
+                      Condition="!Exists('$(MSBuildProjectDirectory)\..\bld\A11yInsightsAttributesInfo.cs')"
+                      Lines="[assembly: System.CLSCompliant(false)] // Generated by build"
```
dbjorge, line 27: The target for `A11yInsightsAttributesInfo.cs` looks like it's just writing static content; can this one be replaced with a static file instead of dynamic build-time generation?

(or maybe just removed entirely? [Assemblies are non-CLS-compliant by default](https://learn.microsoft.com/en-us/dotnet/api/system.clscompliantattribute?view=net-6.0))

src/AccessibilityInsights.Win32/Win32.csproj
```diff
@@ -14,12 +14,22 @@
     <BuildNumber>1.1.$([System.Math]::Round($([System.DateTime]::ParseExact($(CurrentDateString),`yyyy-MM-dd`,null).Subtract($(CountDate)).TotalDays)).ToString(`0000`))$(Revision)</BuildNumber>
   </PropertyGroup>
 
-  <Target Name="CreateVersionFile">
-    <WriteLinesToFile File="$(MSBuildProjectDirectory)\..\bld\A11yInsightsVersionInfo.cs" Overwrite="True" Lines="[assembly: System.Reflection.AssemblyVersion(&quot;$(BuildNumber)&quot;)] // Generated by build" />
+  <Target Name="CreateVersionFile" BeforeTargets="BeforeCompile">
+    <WriteLinesToFile File="$(MSBuildProjectDirectory)\..\bld\A11yInsightsVersionInfo.cs"
```
dbjorge, line 18: Instead of this strategy, did you consider just adding `WriteOnlyWhenDifferent="true"` to the `WriteLinesToFile` directive? I think that would:

* still avoid unnecessary rebuilds (the version file timestamp won't be updated unless necessary, so it won't trigger a rebuild unless actually changed)
* be simpler (wouldn't need the separate clean step)
* be more correct if someone leaves Visual Studio running on a dev box overnight (rebuilds with the new date's version number the next morning even without an explicit clean)

src/AccessibilityInsights.Win32/Win32.csproj
DaveTryon: Analyzer CA1014 wants this attribute to be explicitly set, so using the default value doesn't do the trick. I've moved this into a static file that gets automatically pulled into all shipping projects

src/AccessibilityInsights.Win32/Win32.csproj
DaveTryon: I love this idea... but it didn't work. I modified the .csproj file to generate a different version, and it never overwrote the file with the new content. The only option that I've been able to reliable make work is to have the clean target adn set `Overwrite` to true.

DaveTryon: OK, finally got it to work. I had forgotten to remove the condition to check for the file existence 🤦

codeofdusk (approved):
Not too familiar with the build system here but changes look reasonable to me.

PR #1492, opened 2022-11-29
feat(upgrader-security) : Generate signed manifest

DaveTryon:
#### Details

Generate the signed manifest file that will eventually replace the human-curated and unsigned manifest files that we currently use. The approach is simple:

1. Add a new project that builds `AccessibilityInsights.Manifest.dll`. This project builds only in release, and only _after_ the MSI has been fully created
2. Use a script to create a JSON file that contains information about the generated MSI file
3. Embed this script into the signed manifest assembly. In a future PR, the client will learn how to extract the signed manifest information and use it to ensure that we're installing the exact MSI that was built by the pipeline.

A test iteration of the signed build pipeline can be found [here](https://dev.azure.com/mseng/1ES/_build/results?buildId=18854875&view=results). The section dealing with the signed build of the manifest appears [here](https://dev.azure.com/mseng/1ES/_build/results?buildId=18854875&view=logs&j=7c3ebc89-e7dd-5791-8c02-2ea1f3a27b50&t=9b81c1c0-976a-538f-a1e1-f4d170544876&l=2597). Full disclosure, I made a couple of tiny changes to the PowerShell script to make the logging a little clearer. These changes have been validated locally but will not appear in a signed build until after this merges.

##### Motivation

Part of feature work to improve our updater security

##### Context

- A future PR will publish the manifest files as a build artifact
- A future PR will include the client-side changes for consuming this file. The code is written and validated, but I'm trying to keep the PR sizes as manageable as possible.
- A future PR (after the client code changes are merged) will leverage the client code in the pipeline to ensure that the client code and the manifest remain in sync

#### Pull request checklist

- [n/a] Includes UI changes?


src/props/prod-version.props
```diff
@@ -0,0 +1,7 @@
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+    <!-- If this is a required update, set this to the same version as what we're building.
+	     If this is an optional update, set this to the most recently released prod version. -->
+    <MinimumProdVersion Condition="$(MinimumProdVersion) == ''">1.1.2089.01</MinimumProdVersion>
```
sfoslund, line 5: It's kind of unfortunate that we'll have to remember to update another version number before releasing... Would it be possible to put this alongside the props we have to update before releasing? Or is there another mechanism we can use to make sure we don't forget this?

sfoslund (approved):

src/props/prod-version.props
DaveTryon: Agreed that having 2 things in 2 different places is less than ideal. Fortunately, this is a temporary state. After we complete the whole set of features, manual updates of the `release-info.json` files should go away. The files will still need to exist for downlevel clients, but we should be able to completely automate these file updates, just like we already do for the Canary release channel. We currently update 2 things. During the transition (probably just 1 prod release) we'll update 3 things. After the transition, we'll only update 1 thing.

PR #1445, opened 2022-10-04
Play sound when scanning independent of the screen reader flag

codeofdusk:
##### Summary of the issue
Accessibility Insights for Windows has an option to play a sound when scanning, but it is only honoured when the system screen reader flag is set.

##### Description of how this pull request fixes the issue
Remove the SR flag check. We might bring it back in another location if we need it down the track.

##### Testing strategy
* Run the latest canary build with the sr flag set (run NVDA with no command line options).
    *  With the option to play a sound while scanning active, observe that the Windows NFC connection sound is played while scanning a complex app (such as `devenv`).
* Run the latest canary build without the sr flag set (run nvda with `--no-sr-flag` on the command line).
    *  With the option to play a sound while scanning active, observe that no sound is played while scanning a complex app (such as `devenv`).
* Run a build of this PR with the SR flag set and unset (as above).
    * Observe that the Windows NFC connection sound is played while scanning.

##### Motivation
Inspired by discussions surrounding #1440.

#### Pull request checklist

- [n/a] Run through of all [test scenarios](https://github.com/Microsoft/accessibility-insights-windows/blob/main/docs/Scenarios.md) completed?
- [n/a] Does this address an existing issue? If yes, Issue# - 
- [n/a] Includes UI changes?
  - [n/a] Run the production version of Accessibility Insights for Windows against a version with changes.
  - [n/a] Attach any screenshots / GIF's that are applicable.

Note: After the PR has been created, certain checks will be kicked off. All of these checks must pass before a merge.

Issue #1444, opened 2022-10-04
[Feature Request] Make scan sound unique


codeofdusk:
**Is your feature request related to a problem? Please describe.**
When the corresponding option is enabled, Accessibility Insights plays the sound at `%systemroot%\media\Windows Proximity Connection.wav` (Windows NFC connection sound) when scanning.

**Describe the solution you'd like**
Use a unique sound not included with Windows.

**Describe alternatives you've considered**
Keep sound as-is.

DaveTryon:
@asksep, we think this is worth doing, but we need an appropriate sound. Assigning to you for follow-up

Issue #1440, opened 2022-09-13
[Feature Request] Report to screen reader users when event recording starts and stops


codeofdusk:
**Is your feature request related to a problem? Please describe.**
Although apparently displayed visually, it is immediately unclear to screen reader users when event recording starts or stops.

**Describe the solution you'd like**
A few possible solutions:
* Raise UIA notification events when event recording starts and stops:
    * Works for both speech and Braille users.
    * May not be viable, as NVDA [ignores notification events from background apps](https://github.com/nvaccess/nvda/blob/master/source/NVDAObjects/UIA/__init__.py#L2037) and event recording is toggled using a global keyboard shortcut pressed while the app under inspection is in focus. I'm not sure if this is common to all screen readers – maybe @josephsl knows why NVDA does this?
* Play a short, distinct sound:
    * Sound should be different for starting and stopping recording.
    * Braille-only users still may not know whether event recording starts and stops (such as deafblind users who can't hear the sound).

**Describe alternatives you've considered**
Keep the behaviour as-is.

josephsl:
Hi, when I worked on IUIAutomation5’s notification event handler, I intended notifications from all apps to be announced by NVDA, but @leonardDeR pointed out that people would be distracted if they hear volume level change notifications while using apps other than File Explorer, hence the compromise was made to announce notifications from focused app. There is (or rather was) a pull request that introduced a setting to let NVDA announce notifications from everywhere, only from focused app, or no announcements. Introducing this setting to NVDA will allow folks to use raise notification function from UIA for various tasks, although keep in mind that there are restrictions placed in places such as Windows Calculator where handling notification event can become a distraction (more so when an event in addition to UIA notification (such as name property change) is also fired by the same object that notified users fraction of a second prior). Thanks.

DaveTryon:
@asksep will take this to design office hours to get a sense of what makes here

DaveTryon:
Consider "provide feedback with sound" that is independent of the screen reader presence (as Office does)

PR #1495, opened 2022-12-01
feat(audio-feedback): Introduce the ternary setting

codeofdusk:
#### Details
In #1445, I mentioned the possibility of switching to a ternary (auto/always/never) setting for sound feedback:

* "automatic" plays sound according to the system screen reader flag (default).
* "always" always plays a scanning sound.
* "never" does not play a sound in any circumstances, equivalent to the previous default behaviour.

This PR replaces the current binary `PlayScanningSound` option with a new ternary setting, mapping `true` to `Always`.

##### Motivation
Part of refining the sound feedback strategy (#1440, #1444).

##### Context
Better selection of "which sound to play" in `ProgressRingControl` to follow in subsequent PRs (I wanted to make the changes here as minimal as possible).

There are some odd behaviours that repro intermittently (scanning sound played for a few seconds after I <kbd>Alt</kbd>+<kbd>Tab</kbd>bed), and in one case it played infinitely during an update, but I can't consistently and reliably determine that they're caused by this PR (i.e. they might be current issues).

#### Pull request checklist

- [x] Includes UI changes?

Note: After the PR has been created, certain checks will be kicked off. All of these checks must pass before a merge.


src/AccessibilityInsights.CommonUxComponents/Controls/ProgressRingControl.xaml.cs
```diff
@@ -71,28 +71,6 @@ private static void WithSoundPropertyChanged(DependencyObject d, DependencyPrope
         private bool isPlayingSound;
         private bool withSound;
 
-        /// <summary>
```
DaveTryon, line 74: Removing this allows a couple of using blocks to be removed from the top of this class

src/AccessibilityInsights.SharedUx/Controls/SettingsTabs/ApplicationSettingsControl.xaml
```diff
@@ -210,8 +210,16 @@
...
-                <ToggleButton Grid.Column="2" Style="{StaticResource tgbSlider}" Name="tgbtnPlayScanningSound" AutomationProperties.LabeledBy="{Binding ElementName=lblPlayScanningSoundLabel}" IsChecked="{Binding PlayScanningSound}"/>
+                <Label Name="lblSoundFeedbackLabel" Grid.Column="0" Content="{x:Static Properties:Resources.lblSoundFeedbackLabelContent}" Padding="0" Style="{StaticResource LblText}" VerticalAlignment="Center"/>
+                <ComboBox Grid.Column="2" x:Name ="soundFeedbackComboBox"
+                          SelectedIndex="{Binding SoundFeedbackMode}"
+                          AutomationProperties.Name="{x:Static Properties:Resources.ApplicationSettingsControl_SoundFeedbackModeComboboxAutomationName}"
+                          PreviewKeyDown ="cbComboBox_PreviewKeyDown"
+                          FocusVisualStyle="{DynamicResource {x:Static SystemParameters.FocusVisualStyleKey}}">
+                    <ComboBoxItem x:Name="cbiSFAUTO" Content="{x:Static Properties:Resources.cbiSFAUTOAutomationPropertiesName}" AutomationProperties.Name="{x:Static Properties:Resources.cbiSFAUTOAutomationPropertiesName}"/>
```
DaveTryon, line 219: naming nit: Having AUTO, ALWAYS, and NEVER in all caps is counterintuitive. Can we use Auto, Always, and Never, instead? This would apply to both the item name and the string resource name. 

The all caps in the highlight combo comes from the fact that they're abbreviations (HLTT = HighLighter + ToolTop), but that doesn't apply here.

src/AccessibilityInsights.SharedUx/Controls/SettingsTabs/ApplicationSettingsControl.xaml.cs
```diff
@@ -204,7 +204,7 @@ public bool IsConfigurationChanged(ConfigurationModel config)
             {
                 return true;
             }
-            if (config.PlayScanningSound != (this.DataContextVM.PlayScanningSound))
+            if ((int) config.SoundFeedback != (this.DataContextVM.SoundFeedbackMode))
```
DaveTryon, line 207: For consistency with the HighligherMode code, maybe the cast should be on the right hand side?
```suggestion
            if ((config.SoundFeedback != (SoundFeedbackMode)(this.DataContextVM.SoundFeedbackMode))
```

src/AccessibilityInsights.SharedUx/ViewModels/ApplicationSettingsViewModel.cs
```diff
@@ -119,17 +119,17 @@ public int HighlighterMode
             }
         }
 
-        bool _playingScanningSound;
-        public bool PlayScanningSound
+        int _soundMode;
```
DaveTryon, line 122: naming nit: I'd suggest `_soundFeedbackMode` to emphasize that this is a backing field

src/AccessibilityInsights.SharedUxTests/Settings/ConfigurationModelTests.cs
```diff
@@ -181,7 +181,7 @@ private static ConfigurationModel GetDefaultConfig()
             Assert.AreEqual(issueReporterSerializedConfigs, config.IssueReporterSerializedConfigs);
             Assert.IsTrue(config.IsUnderElementScope);
             Assert.AreEqual(200, config.MouseSelectionDelayMilliSeconds);
-            Assert.IsFalse(config.PlayScanningSound);
+            Assert.AreEqual(SoundFeedbackMode.Auto, config.SoundFeedback);
```
DaveTryon, line 184: Would it make sense to have unit tests that read in non-default values?

src/AccessibilityInsights.SharedUx/Controls/SettingsTabs/ApplicationSettingsControl.xaml
```diff
@@ -210,8 +210,16 @@
                     <ColumnDefinition Width="151"/>
                     <ColumnDefinition Width="50"/>
                 </Grid.ColumnDefinitions>
-                <Label Name="lblPlayScanningSoundLabel" Grid.Column="0" Content="{x:Static Properties:Resources.lblPlayScanningSoundLabelContent}" Padding="0" Style="{StaticResource LblText}" VerticalAlignment="Center"/>
-                <ToggleButton Grid.Column="2" Style="{StaticResource tgbSlider}" Name="tgbtnPlayScanningSound" AutomationProperties.LabeledBy="{Binding ElementName=lblPlayScanningSoundLabel}" IsChecked="{Binding PlayScanningSound}"/>
+                <Label Name="lblSoundFeedbackLabel" Grid.Column="0" Content="{x:Static Properties:Resources.lblSoundFeedbackLabelContent}" Padding="0" Style="{StaticResource LblText}" VerticalAlignment="Center"/>
```
DaveTryon, line 213: Line 211 of this file sets the control column width to 50, which was correct for the toggle but is too narrow for the combobox. We should set it to 180, which is the same width that is set for the highlight combobox.

src/AccessibilityInsights.SharedUx/Settings/ConfigurationModel.cs
```diff
@@ -614,7 +614,7 @@ public static ConfigurationModel GetDefaultConfigurationModel(FixedConfigSetting
                 SelectionByMouse = true,
                 ShowWelcomeScreenOnLaunch = true,
                 AlwaysOnTop = true,
-                PlayScanningSound = false,
+                SoundFeedback = SoundFeedbackMode.Auto,
```
DaveTryon, line 617: Someone who is upgrading could have `PlayScanningSound` set to either true or false. Do both of these map to `SoundFeedbackMode.Auto`? If not, we should provide appropriate mapping inside the `LoadDataFromJSON` method. 

At some future time, we'll drop support for versions that use `PlayScanningSound` and will want to remove that key from any dictionaries that we read. It might be worth adding a TODO inside `LoadDataFromJSON` so we have a reminder

DaveTryon (commented):
I added some comments in the code.

In the PR description, you chose n/a for UI changes, which is inaccurate since the settings UI is changing. Also, it might be good to have a less verbose feature name, since git tools visually truncate the commit title. Maybe feat(audio-feedback) ?

src/AccessibilityInsights.SharedUx/Settings/ConfigurationModel.cs
codeofdusk: That depends – there are (at least) two ways to look at that:

* The production release doesn't include #1445. Therefore, a `true` value is technically equivalent to our new `Auto` setting (play a sound only when the SR flag is set). Map `true` to `Auto` as not to change behaviour.
* I made #1445 because the old behaviour (play a sound only when the setting is enabled and the SR flag is set) seems counterintuitive and, as far as I can tell, isn't documented. Therefore, if a user explicitly set the "play scanning sound" option to `true`, they probably mean `Always` (i.e. they don't care whether a screen reader is running if they've explicitly enabled the option). Map `true` to `Always`.

DaveTryon (approved):

PR #1496, opened 2022-12-01
chore: Flush telemetry at app shutdown

DaveTryon:
#### Details

We use [ApplicationInsights](https://www.nuget.org/packages/Microsoft.ApplicationInsights) for our telemetry, which buffers telemetry events for about 30 seconds, then sends them in a burst. This works fine most of the time, but we lose telemetry events that occur in the last few seconds of the app's lifetime. This is especially noticeable with updates, since the app exits within milliseconds of creating the event.

This PR connects the MainWindow's existing onClosed handler to the `Flush` method from ApplicationInsights. Due to how the code is structured, it goes through the following layers:
- `MainWindow.onClosed` calls `Logger.FlushAndShutDown`
- `Logger.FlushAndShutDown` calls (through an interface) into `AITelemetrySink.FlushAndShutDown`
- `AITelemetrySink.FlushAndShutDown` calls into each available implementation of `ITelemetry.FlushAndShutDown` inside a try/catch loop
- `AITelemetry.FlushAndShutDown` calls the ApplicationInsights `Flush` method, which immediately uploads any buffered telemetry events

Unit test are also updated. This was validated by creating a scenario where a locally built AIWin would update, then checking the uploaded telemetry to confirm that the `Upgrade_DoInstallation` event (which occurs immediately before shutdown) is correctly uploaded via ApplicationInsights

##### Motivation

Telemetry that we can't capture isn't very useful ☹️ 

##### Context

One potential problem to note: Uploading telemetry requires a network call, so uploading any buffered data will introduce a nonzero delay when exiting the app. This won't be user-perceivable because it occurs _after_ the window has closed. It does, however, create the following potential race condition:
- When updating, AIWin triggers the VersionSwitcher then exits after the VersionSwitcher is successfully running
- AIWin tries to flush buffered telemetry on shutdown. This is normally very fast, but suppose it takes an exceptionally long time
- If VersionSwitcher gets to the point where it needs to run the MSI install, the MSI install will fail if AIWin is still running
- Update will fail

Given that VersionSwitcher downloads the MSI (currently 7MB) from the web, then does work to ensure that the MSI signature chain is still trustworthy, the telemetry upload should normally take _much_ less time than the time to download and validate the MSI. It could still happen, however, in the unlikely scenario that the ApplicationInsights  geo-redundant collection points were _all_ offline. Should that occur, any of the following mitigations would unblock the user:
- Restart the app and don't accept the upgrade (we allow Prod to stay 1 version back unless there's a critical fix that we need to get out)
- Install the most recent Canary version from https://github.com/microsoft/accessibility-insights-windows/releases
- Uninstall AIWin and reinstall it from www.AccessibilityInsights.io (this ensure the most current production version)
- Try again later in a few minutes on the assumption that the collection points will be back online
- Run the app without telemetry

If we see this race condition happening frequently in the wild, the `onClosed` handler could esaily wrap the call to `Logger.FlushAndShutDown` in a `Task`, then specify a maximum time (maybe 1 second) that it's willing wait for that `Task` to complete. My inclination is to take a YAGNI approach on this until we have a demonstrated need, but I could possibly be convinced otherwise.

#### Pull request checklist

- [n/a] Includes UI changes?

### DIFF
--- a/src/AccessibilityInsights.Extensions.DiskLoggingTelemetry/LocalTelemetry.cs
+++ b/src/AccessibilityInsights.Extensions.DiskLoggingTelemetry/LocalTelemetry.cs
@@ -89,5 +89,13 @@ namespace AccessibilityInsights.Extensions.DiskLoggingTelemetry
             _logWriter.LogThisData("Exception was reported",
                 JsonConvert.SerializeObject(exceptionData, Formatting.Indented));
         }
+
+        /// <summary>
+        /// Application is shutting down, so flush any pending telemetry
+        /// </summary>
+        public void FlushAndShutDown()
+        {
+            // We never cache telemetry, so this does nothing
+        }
     }
 }
--- a/src/AccessibilityInsights.Extensions.DiskLoggingTelemetryTests/LocalTelemetryUnitTests.cs
+++ b/src/AccessibilityInsights.Extensions.DiskLoggingTelemetryTests/LocalTelemetryUnitTests.cs
@@ -269,5 +269,12 @@ namespace AccessibilityInsights.Extensions.DiskLoggingTelemetryTests
             AssertDictionariesAreEquivalent(ContextProperties, exceptionData.ContextProperties);
             _logWriterMock.VerifyAll();
         }
+
+        [TestMethod]
+        public void FlushAndShutDown_DoesNothing()
+        {
+            _testSubject.FlushAndShutDown();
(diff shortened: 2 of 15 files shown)


sfoslund (approved):
LGTM!

Re: the potential race condition you discuss in the PR description context section, it makes sense to me to assume that this is extremely unlikely unless proven wrong and therefore not set up any specific logic to handle it at this time.

Issue #1477, opened 2022-11-10
Lots of debug warnings when running in IDE


DaveTryon:
**Describe the bug**
When starting a debug build of AIWin in the IDE, it generates a series of error messages. Customers won't see these warnings, but each warning has the potential to represent a customer-impacting issue, so they deserve to be investigated.

**To Reproduce**
1. Run a debug build from within Visual Studio
2. Look at the Output window
3. Notice the data errors

**Expected behavior**
Data errors generally shouldn't happen

**Actual behavior**
The following errors are displayed (this is a filtered list to get just the errors that appear to be directly attributable to our code, with duplicates removed):
```
System.Windows.Data Error: 40 : BindingExpression path error: 'SnippetLink' property not found on 'object' ''MainWindow' (Name='')'. BindingExpression:Path=DataContext.SnippetLink; DataItem='StackPanel' (Name='spHowToFix'); target element is 'TextBlock' (Name='tbSnippetLink'); target property is 'NoTarget' (type 'Object')
System.Windows.Data Error: 40 : BindingExpression path error: 'projects' property not found on 'object' ''ConfigurationViewModel' (HashCode=45834852)'. BindingExpression:Path=projects; DataItem='ConfigurationViewModel' (HashCode=45834852); target element is 'TreeView' (Name='serverTreeview'); target property is 'ItemsSource' (type 'IEnumerable')
System.Windows.Data Error: 5 : Value produced by BindingExpression is not valid for target property.; Value='-4' BindingExpression:Path=ActualWidth; DataItem='LiveModeControl' (Name='ctrlLiveMode'); target element is 'ColumnDefinition' (HashCode=25265097); target property is 'MaxWidth' (type 'Double')
System.Windows.Data Error: 1 : Cannot create default converter to perform 'two-way' conversions between types 'System.Double' and 'System.Windows.GridLength'. Consider using Converter property of Binding. BindingExpression:Path=Width; DataItem='Grid' (Name=''); target element is 'ColumnDefinition' (HashCode=5180964); target property is 'Width' (type 'GridLength')
System.Windows.Data Error: 8 : Cannot save value from target back to source. BindingExpression:Path=Width; DataItem='Grid' (Name=''); target element is 'ColumnDefinition' (HashCode=5180964); target property is 'Width' (type 'GridLength') ArgumentException:'System.ArgumentException: '*' is not a valid value for property 'Width'.
[...]

PR #1497, opened 2022-12-01
fix: Fix debug build errors

sfoslund:
#### Details

Fix the first batch of debug errors from https://github.com/microsoft/accessibility-insights-windows/issues/1477:

| Error | Change | Notes |
|---|---|---|
| `BindingExpression path error: 'SnippetLink' property not found on 'object'...` | `ScannerResultControl.xaml.cs` | Explicitly set data context to null on start up so it doesn't try to populate properties that do not exist yet | 
| `BindingExpression path error: 'projects' property not found on 'object'...` | `ConfigurationControl.xaml.cs` | Explicitly set items source to null on start up so it doesn't try to populate properties that do not exist yet |
| `Cannot create default converter to perform 'two-way' conversions...` | `EventDetailControl.xaml` and `PropertyInfoControl.xaml` | I can't discern a reason that the `OneWayToSource` mode is necessary, so removing it. |
| `Cannot save value from target back to source. BindingExpression:Path=Width;` | Same as above | Same as above |
| `Cannot find source for binding with reference 'ElementName=labelResults'...` | `ScannerResultCustomListControl.xaml` | This line seems to have been missed in a prior refactoring: https://github.com/microsoft/accessibility-insights-windows/commit/15fd62773c826fd33a5d302731f8f4398cf3f416|

There are several more errors that this PR does not cover, so leaving the original issue open for that reason. 

##### Motivation

Address https://github.com/microsoft/accessibility-insights-windows/issues/1477

#### Pull request checklist

- [x] Does this address an existing issue? If yes, https://github.com/microsoft/accessibility-insights-windows/issues/1477
- [n/a] Includes UI changes?

### DIFF
--- a/src/AccessibilityInsights.Extensions.AzureDevOps/ConfigurationControl.xaml.cs
+++ b/src/AccessibilityInsights.Extensions.AzureDevOps/ConfigurationControl.xaml.cs
@@ -31,6 +31,7 @@ namespace AccessibilityInsights.Extensions.AzureDevOps
         {
             InitializeComponent();
             DataContext = VMConfig;
+            serverTreeview.ItemsSource = null;
         }
 
         private bool canSave;
--- a/src/AccessibilityInsights.Extensions.AzureDevOps/ConfigurationControl.xaml.cs
+++ b/src/AccessibilityInsights.Extensions.AzureDevOps/ConfigurationControl.xaml.cs
@@ -31,7 +31,6 @@ namespace AccessibilityInsights.Extensions.AzureDevOps
         {
             InitializeComponent();
             DataContext = VMConfig;
-            serverTreeview.ItemsSource = null;
         }
 
         private bool canSave;
@@ -67,10 +66,6 @@ namespace AccessibilityInsights.Extensions.AzureDevOps
         /// </summary>
         public ConfigurationViewModel VMConfig { get; private set; } = new ConfigurationViewModel();
 
-        /// <summary>
-        /// List of team projects
-        /// </summary>
-        public BindingList<TeamProjectViewModel> projects { get; private set; } = new BindingList<TeamProjectViewModel>();
         public override Action UpdateSaveButton { get; set; }
 
         #region configuration updating code
@@ -250,11 +245,11 @@ namespace AccessibilityInsights.Extensions.AzureDevOps
                 try
                 {
                     ToggleLoading(true);
-                    Dispatcher.Invoke(projects.Clear);
+                    Dispatcher.Invoke(VMConfig.Projects.Clear);
                     var newProjectList = await UpdateTeamProjects().ConfigureAwait(true); // need to come back to original UI thread.
-                    Dispatcher.Invoke(() => newProjectList.ForEach(p => projects.Add(p)));
+                    Dispatcher.Invoke(() => newProjectList.ForEach(p => VMConfig.Projects.Add(p)));
                     ToggleLoading(false);
-                    Dispatcher.Invoke(() => serverTreeview.ItemsSource = projects);
+                    Dispatcher.Invoke(() => serverTreeview.ItemsSource = VMConfig.Projects);
                     Dispatcher.Invoke(() => serverTreeview.Focus());
                 }
 #pragma warning disable CA1031 // Do not catch general exception types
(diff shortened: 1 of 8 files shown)


src/AccessibilityInsights.SharedUx/Controls/ScannerResultControl.xaml.cs
```diff
@@ -37,6 +37,7 @@ public ScannerResultControl()
             InitializeComponent();
             _list = new List<ScanListViewItemViewModel>();
             Resources.Source = new Uri(@"pack://application:,,,/AccessibilityInsights.SharedUx;component/Resources/Styles.xaml", UriKind.Absolute);
+            this.spHowToFix.DataContext = null;
```
codeofdusk, line 40: No need for explicit `this`:

```suggestion
            spHowToFix.DataContext = null;
```

codeofdusk (approved):
Can't speak to the XAML but the C# changes look reasonable to me!

src/AccessibilityInsights.Extensions.AzureDevOps/ConfigurationControl.xaml.cs
```diff
@@ -31,6 +31,7 @@ public ConfigurationControl()
         {
             InitializeComponent();
             DataContext = VMConfig;
+            serverTreeview.ItemsSource = null;
```
DaveTryon, line 34: This clears the error, but I'm not sure it's the correct fix. The  warning refers to a non-existent `ConfigurationViewModel.projects` property, then the `ChangeStates` property sets `serverTreeview.ItemsSource = projects`, where `projects` is on `ConfigurationControl` instead of the ViewModel. Do you think it would make sense to move `ConfigurationControl.projects` to `ConfigurationViewModel.projects` ?

(As a side note, the property should really be PascalCased, so `Projects`)

sfoslund: > Do you think it would make sense to move ConfigurationControl.projects to ConfigurationViewModel.projects

Yup, that makes sense to me, I'll implement that

src/AccessibilityInsights.Extensions.AzureDevOps/ConfigurationControl.xaml.cs
```diff
@@ -249,11 +245,11 @@ private async void ChangeStates(ControlState state)
...
                 {
                     ToggleLoading(true);
-                    Dispatcher.Invoke(projects.Clear);
+                    Dispatcher.Invoke(VMConfig.Projects.Clear);
                     var newProjectList = await UpdateTeamProjects().ConfigureAwait(true); // need to come back to original UI thread.
-                    Dispatcher.Invoke(() => newProjectList.ForEach(p => projects.Add(p)));
+                    Dispatcher.Invoke(() => newProjectList.ForEach(p => VMConfig.Projects.Add(p)));
                     ToggleLoading(false);
-                    Dispatcher.Invoke(() => serverTreeview.ItemsSource = projects);
+                    Dispatcher.Invoke(() => serverTreeview.ItemsSource = VMConfig.Projects);
```
DaveTryon, line 252: With the binding working correctly, we may be able to remove the code to update the ItemsSource

src/AccessibilityInsights.SharedUx/Controls/ScannerResultControl.xaml.cs
```diff
@@ -37,6 +37,7 @@ public ScannerResultControl()
             InitializeComponent();
             _list = new List<ScanListViewItemViewModel>();
             Resources.Source = new Uri(@"pack://application:,,,/AccessibilityInsights.SharedUx;component/Resources/Styles.xaml", UriKind.Absolute);
+            spHowToFix.DataContext = null;
```
DaveTryon, line 40: This one also seems odd. The debugger says that `spHowToFix.DataContext` is already null so this shouldn't change anything. The warning refers to not finding `MainWindow.HowToFixText`, which doesn't exist because the `HowToFixText` property is on `ScanListViewItemViewModel`. I'm not sure how to fix the context mismatch on this--maybe we're missing a `ScannerResultControlViewModel` ?

sfoslund: Yeah, I wasn't sure how to address this either. I ended up going with this because of the logic in the [UpdateTree method](https://github.com/microsoft/accessibility-insights-windows/blob/main/src/AccessibilityInsights.SharedUx/Controls/ScannerResultControl.xaml.cs#L111). My assumption is that we reassign the datacontext there when necessary, and the default in that method if for the datacontext to be null. As a result, I figured it would make sense to start by explicitly initializing it as null too, which will only last until UpdateTree is first called anyways. I'm definitely open to other suggestions though!

DaveTryon: Yeah, looking [here](https://github.com/microsoft/accessibility-insights-windows/blob/main/src/AccessibilityInsights.SharedUx/Controls/ScannerResultControl.xaml.cs#LL217C14-L217C14) we use null as a valid value. Let's go with this, even though it's a bit odd

DaveTryon (approved):

PR #1498, opened 2022-12-02
chore: Update to newer NetAnalyzers

DaveTryon:
#### Details

Current builds are cluttered with the following warning:
```
The .NET SDK has newer analyzers with version '7.0.0' than what version '6.0.0' of 'Microsoft.CodeAnalysis.NetAnalyzers' package provides. Update or remove this package reference.
```

https://www.nuget.org/packages/Microsoft.CodeAnalysis.NetAnalyzers has been in v7 preview for about 9 months but hasn't yet shipped. This PR bumps to the most recent preview version to reduce the clutter, with the expectation that dependabot will notify us when v7 is officially released. These changes were made by directly editing the .targets file, but are based on IDE-generated changes from https://github.com/microsoft/axe-windows/pull/853.

##### Motivation

Reduce noise in builds (local and pipeline), use latest tools

##### Context

We want to pin our versions (as opposed to just using `latest`) to avoid random errors that suddenly appear for no obvious reason, and because dev machines and pipeline machines aren't necessarily on the same version of Visual Studio.

#### Pull request checklist

- [n/a] Includes UI changes?

### DIFF
--- a/build/NetFrameworkRelease.targets
+++ b/build/NetFrameworkRelease.targets
@@ -27,7 +27,7 @@
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
-    <PackageReference Include="Microsoft.CodeAnalysis.NetAnalyzers" Version="6.0.0">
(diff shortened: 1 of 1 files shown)


sfoslund (approved):
I'd been meaning to look at this warning for a while, glad you were able to fix it!

PR #1493, opened 2022-11-30
feat(upgrader-security): Consume signed manifests, add telemetry and docs

DaveTryon:
#### Details

This adds the client-side code to consume signed manifests (it temporarily also consumes unsigned manifests), as well as the code to use the data in the signed manifests to ensure that we only install non-tampered manifests. I tried to break this PR down into smaller pieces, but I couldn't get it any smaller than this. A reasonable overview of the process is added in `ChangingVersions.md`, but here's a (lengthy) breakdown of the code changes, in rough order of execution:

##### Fetching the manifest
- We add 3 new fields to the `ChannelInfo` class. These fields contain the metadata about the signed MSI file.
- We define a `StreamMetadata` class to provide more information about HTTP calls.
- We define an `InterceptingWebClient` class that allows us to capture the fully resolved Uri that was the actual source of whatever we loaded.
- In `GitHubClient.LoadUriContentsIntoStream`, we use the `InterceptingWebClient` to build a `StreamMetadata` that describes the web resource that we fetched.
- In `ChannelIntoUtilities.GetChannelFromStream`, we first try to parse the stream as a signed manifest (including checking the signature chain). If that fails, we try to parse the stream as an unsigned manifest. The code to use the unsigned manifest will be removed after all release channels have migrated to signed manifests.
- We define an `EnrichedContentInfo` class that derives from `ContentInfo`, adding the `StreamMetadata` as a member. This gives us a single object that contains the update information.
- We synthesize `ChannelInfo.MinimumVersion` if we're reading an unsigned manifest (this data depends on the release channel and is not included in signed manifest files).
- We add 3 manifest metadata fields to the `IAutoUpdate` interface. These exist for telemetry, since extensions can't publish telemetry events directly.
- The `GithubAutoUpdate.AutoUpdate` class populates the metadata properties based on the data from the `EnrichedChannelnfo` object, and the manifest metadata fields get bundled with the `Upgrade_GetUpgradeOption` telemetry action.

##### Running the upgrade
- We define the `ExecutionResult` enum to capture possible outcomes of running VersionSwitcher.
- We define the `ResultBearingException` class, which is just a regular `Exception` with an `ExecutionResult` attached.
- We define the `ResultExecutionWrapper` class, which allows us to have layers of failure capture, and the most granular failure will be preserved.
- We define an `ExecutionHistory` class that will be used for telemetry from VersionSwitcher. It also replaces the existing EventLogger and ExceptionReporter classes that VersionSwitcher used for local debugging.
[...]

PR #1505, opened 2022-12-03
feat(upgrader-security): Add integration tests for manifests

DaveTryon:
#### Details

Add tests to catch potential manifest mismatches at build time. This is the "future PR" mentioned in the "Context" paragraph of #1493. These tests essentially execute the code that will be run on the client to make sure that the manifest is available and that it contains the expected data. Since these tests rely on the `MSI` and `Manifest` projects, and since these projects are excluded from debug builds, the `ManifestTests` project is likewise excluded from debug builds.

##### Motivation

Keep manifest mechanism robust

##### Context

I wanted to add a test that would check for signing of the manifest assembly, but that wouldn't play very nicely with unsigned builds. The signed pipeline already verifies the manifest for correct signing, so it seemed safe to omit that test. I also had to create a way to bypass the `TrustVerifier` check for unsigned builds. It would be cool if this bypass were skipped in the `SignedRelease` build, but I wasn't sure how to make that work.

#### Pull request checklist

- [n/a] Includes UI changes?


src/ManifestTests/IntegrationTests.cs
```diff
@@ -0,0 +1,74 @@
...
+        }
+
+        [TestMethod]
+        public void EmbeddedManifestMatchesRawManifest()
+        {
+            using (Stream stream = new FileStream(SignedManifestFilePath, FileMode.Open, FileAccess.Read, FileShare.Read))
+            {
+                stream.Position = 0;
+
+                ChannelInfo signedInfo = ChannelInfoUtilities.GetChannelInfoFromSignedManifest(stream, (_) => true);
```
sfoslund, line 62: > It would be cool if this bypass were skipped in the SignedRelease build, but I wasn't sure how to make that work.

I wonder if there's a way to use preprocessor directives to do this? We could conditionally define a `SIGNED_BUILD` constant using [DefineConstants](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-options/language#defineconstants) and use it to have an if/else check here along [these lines](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/preprocessor-directives#conditional-compilation)?

sfoslund (approved):
LGTM! One suggestion but I don't see it as blocking :)

sfoslund (approved):

PR #1509, opened 2022-12-07
feat: Improve discovery of keyboard shortcuts

DaveTryon:
#### Details

New keyboard users may not be familiar with the metaphor of using F6 to move between panes. To make this a bit easier to discover, we're including this information and a link to the full keyboard shortcuts in the Live Inspect pane. The link goes through a fwlink and resolves to https://accessibilityinsights.io/docs/windows/reference/keyboard/.

##### Motivation

Improve the UX for new keyboard users.

##### Context

Like many other places in the code, this uses string concatenation and is not really localizable. I tried a version that used a RegEx, but the Hyperlink styling was wrong. The same problem exists in all 3 text blocks in this control. so I'm deferring solving that problem for a later time if/when we take on proper localization.

We discussed putting this into the Welcome screen, but decided against it for the following reasons:
1. That dialog is already at its limit at 200% zoom
2. Should the user ever uncheck the "Show this dialog at startup" checkbox, we don't provide an easy way to get it back

#### Screenshots

Before | After
--- | ---
image | image

- [x] Includes UI changes?

### DIFF
--- a/src/AccessibilityInsights/Properties/Resources.resx
+++ b/src/AccessibilityInsights/Properties/Resources.resx
@@ -587,7 +587,7 @@ Confidence: {1}</value>
     <comment>{0} is the text displayed in the tooltip</comment>
   </data>
   <data name="LiveModeControl_Keyboard" xml:space="preserve">
-    <value>Use Tab or Shift+Tab to move between fields in the same pane. Use F6 or Shift+F6 to move between panes</value>
+    <value>Use Tab or Shift+Tab to move between fields in the currently active pane. Use F6 or Shift+F6 to move between panes</value>
   </data>
   <data name="LiveModeControl_LearnMoreKeyboard" xml:space="preserve">
     <value>Learn more about keyboard shortcuts</value>
--- a/src/AccessibilityInsights/Properties/Resources.Designer.cs
+++ b/src/AccessibilityInsights/Properties/Resources.Designer.cs
@@ -710,7 +710,7 @@ namespace AccessibilityInsights.Properties {
         }
         
         /// <summary>
-        ///   Looks up a localized string similar to Use Tab or Shift+Tab to move between fields in the same pane. Use F6 or Shift+F6 to move between panes.
+        ///   Looks up a localized string similar to Use Tab or Shift+Tab to move between fields in the currently active pane. Use F6 or Shift+F6 to move between panes.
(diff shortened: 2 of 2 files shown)


src/AccessibilityInsights/Properties/Resources.resx
```diff
@@ -586,4 +586,10 @@ Confidence: {1}</value>
     <value>{0} 1 of 2</value>
     <comment>{0} is the text displayed in the tooltip</comment>
   </data>
+  <data name="LiveModeControl_Keyboard" xml:space="preserve">
+    <value>Use Tab or Shift+Tab to move between fields in the same pane. Use F6 or Shift+F6 to move between panes</value>
```
codeofdusk, line 590:
```suggestion
    <value>Use Tab or Shift+Tab to move between fields in the currently active pane. Use F6 or Shift+F6 to move between panes</value>
```

codeofdusk (changes requested):
I like the approach here! However, I'm a bit concerned that this isn't very discoverable – focus lands on the "to inspect a UI Automation element..." text, then I have to tab four times to reach the new hint. Could the tab/focus order be changed so that this new text is immediately focused on app launch to ease discoverability?

DaveTryon:
> Could the tab/focus order be changed so that this new text is immediately focused on app launch to ease discoverability?

I think this would be a tough sell with design, since the navigation isn't really an option until something is either inspected or tested. I would think that a brand new user would tend to explore a bit on the first run, so requiring those first tabs doesn't seem terribly unreasonable. 

In an ideal world, I'd love to have a real tutorial experience, but that's probably unrealistic given the other things that would be competing for engineering and design budget.

codeofdusk:
> I would think that a brand new user would tend to explore a bit on the first run, so requiring those first tabs doesn't seem terribly unreasonable.

Consider:

1. A new user launches the app and navigates out of the welcome screen.
2. The "to begin inspecting..." text is presented.
3. The user navigates away (as instructed) and presses <kbd>Shift</kbd>+<kbd>F9</kbd>.
4. The user is dropped into the UIA tree. Tabbing around is, of course, locked to the UIA tree pane. At no time can the keyboard shortcuts hint be discovered.

codeofdusk (approved):
Discussed offline, concerns to be addressed in a follow-up.

Issue #1512, opened 2022-12-07
[BUG] Improve immediate discoverability of keyboard shortcuts


codeofdusk:
Please check whether the bug has [already been filed](https://github.com/Microsoft/accessibility-insights-windows/issues).

This repo contains code for Accessibility Insights for Windows. We maintain code for its automated checks rule engine at [axe-windows](https://github.com/microsoft/axe-windows/issues/). Most false positive questions or rule clarifications belong in [axe-windows](https://github.com/microsoft/axe-windows/issues/new/choose), whereas questions about the Accessibility Insights tool UI should go here. If you aren't sure, it's definitely OK to use this repo. Thanks for contributing!

**Describe the bug**
#1509 added a hint (in the tab order) describing <kbd>Tab</kbd>/<kbd>Shift</kbd>+<kbd>Tab</kbd> and <kbd>F6</kbd>/<kbd>Shift</kbd>+<kbd>F6</kbd>, but this is easy for new users to miss as earlier text redirects them away from this part of the app.

**To Reproduce**
Steps to reproduce the behaviour:
1. Launch the app and navigate out of the welcome screen. The "to begin inspecting a UI Automation element..." text is presented.
2. The user navigates away (as instructed) to another running app, then presses <kbd>Shift</kbd>+<kbd>F9</kbd>.

**Expected behaviour**
There is some means by which the keyboard shortcuts hint can be discovered in this scenario, such as moving it to the beginning of the tab order (in place of the "to inspect a UI Automation element..." text) or screen reader only help text directing the user to tab through and read all instructions before starting.

**Actual behaviour**
The user is dropped into the UIA tree. Tabbing around is, of course, locked to the UIA tree pane. At no time can the keyboard shortcuts hint be discovered.

**Desktop (please complete the following information):**
 - OS: [e.g. Windows 10 1809] (Get the version by running `winver` from the command line): Windows SV2
 - Accessibility Insights for Windows Version: 31fa6b5
 - Target Application: [e.g. Chrome, Wildlife Manager]: n/a
 - Target Application Version: [e.g. 22]: n/a

PR #1515, opened 2022-12-10
chore: Use existing link instead of new one

DaveTryon:
#### Details

In #1512, we added a new way to discover keyboard shortcuts. In doing so, we created a new link instead of using the one that already existed and was used at https://github.com/microsoft/accessibility-insights-windows/blob/main/src/AccessibilityInsights.SharedUx/Dialogs/StartUpModeControl.xaml#L159. This converts both locations to using the existing link. I'll delete the newly created link after this deploys to the Canary release channel.

Verified by clicking link from Live Inspect page and confirming that it ends up where expected.

##### Motivation

Avoid redundancy

#### Pull request checklist

- [n/a] Includes UI changes?


codeofdusk (approved):
LGTM! Verified that the fwlink link works as expected.

PR #1525, opened 2023-01-12
feat(audio-feedback): Play sounds when toggling event recording

### LINKED ISSUE #1440: [Feature Request] Report to screen reader users when event recording starts and stops

codeofdusk:
#### Details
This PR adds functionality for playing sounds on starting/stopping event recording when sound feedback is enabled.

##### Motivation
Closes #1440.

##### Context
* When rapidly pressing <kbd>Shift</kbd>+<kbd>F7</kbd>, there may be some delay in sound playback as event recording takes a moment to toggle in some circumstances.

#### Pull request checklist

- [x] Does this address an existing issue? If yes, Issue# - #1440
- [n/a] Includes UI changes?

Note: After the PR has been created, certain checks will be kicked off. All of these checks must pass before a merge.


src/AccessibilityInsights.SharedUx/Controls/EventRecordControl.xaml.cs
```diff
@@ -95,8 +101,12 @@ public static ConfigurationModel AppConfiguration
         public EventRecordControl()
         {
             InitializeComponent();
-
             InitCommandBindings();
+
+            player = new SoundPlayer();
+            Assembly assembly = Assembly.GetExecutingAssembly();
+            startRecordingSoundStream = assembly.GetManifestResourceStream(assembly.GetName().Name + ".Resources.Sound.start_event_recording.wav");
```
codeofdusk, line 108: Should this filename be referred to by a constant or resource string somewhere?

src/AccessibilityInsights.SharedUx/SharedUx.csproj
```diff
@@ -29,6 +29,8 @@
 
   <ItemGroup>
     <None Remove="Resources\Icons\BrandIcon.ico" />
+    <None Remove="Resources\Sound\start_event_recording.wav" />
```
codeofdusk, line 32: Should two references to this new resource be added?

DaveTryon: Yes, the resource should have both Remove and Include. I'd suggest that we follow the pattern that the IDE uses to add resources, and create a new ItemGroup for each resource. That keeps everything about each file in one place, instead of scattering it around. It would look something like this:

```
<ItemGroup>
  <None Remove="Resources\Sound\start_event_recording.wav" />
  <EmbeddedResource Include="Resources\Sound\start_event_recording.wav" />
</ItemGroup>

<ItemGroup>
  <None Remove="Resources\Sound\stop_event_recording.wav" />
  <EmbeddedResource Include="Resources\Sound\stop_event_recording.wav" />
</ItemGroup>
```

src/AccessibilityInsights.SharedUx/Controls/EventRecordControl.xaml.cs
DaveTryon: I would say yes if we were referencing it in multiple places. In this case, though, it seems simpler to just have the string inline as you've done.

src/AccessibilityInsights.SharedUx/SharedUx.csproj
```diff
@@ -49,6 +51,8 @@
 
   <ItemGroup>
     <None Remove="Resources\video_thumb.png" />
+    <EmbeddedResource Include="Resources\Sound\start_event_recording.wav" />
+    <EmbeddedResource Include="Resources\Sound\stop_event_recording.wav" />
```
DaveTryon, line 55: Build is failing because the .wav files aren't included in the changelist

DaveTryon (changes requested):
Sound files are not included in the PR

DaveTryon (approved):

Issue #1528, opened 2023-01-12
[BUG] Hierarchy settings menu does not honor dark mode


sfoslund:
**Describe the bug**
Hierarchy settings menu does not honor dark mode

**To Reproduce**
1. Open AIWin in dark mode
2. Live inspect any app
3. Click Settings icon at top right of left panel
4. See hierarchy settings menu is in light mode

**Expected behavior**
Entire app adheres to dark mode setting

**Actual behavior**
Hierarchy settings menu remains in light mode

**Screenshots or .GIF**

[Description: Screenshot of AI win in dark mode, with hierarchy settings menu open and still in light mode]
<img width="667" alt="AIWinDarkModeSettings" src="https://user-images.githubusercontent.com/16010855/212184028-e6a26171-2989-44c2-b76f-d6c4cff1de5e.png">

**Desktop (please complete the following information):**
 - OS: Windows 11 22621
 - Accessibility Insights for Windows Version: Current prod (1.1.2089.1)
 - Target Application: Wildlife Manager
 - Target Application Version: 2.0

jlperkins:
Repro'd, thanks for finding this Sarah!

sfoslund:
It looks like we have the same issue with the event settings menu as well:

[Image description: screenshot of AI win event settings menu open in dark mode, menu is still light grey]
<img width="831" alt="EventSettingsNotDarkMode" src="https://user-images.githubusercontent.com/16010855/213752075-c6ba760c-7297-45fb-aa7e-5f8f8b101382.png">

DaveTryon:
This has been fixed in Canary since release 1.1.2249.01. Closing the issue as fixed.

Issue #1514, opened 2022-12-09
[BUG] Places where we synthesize names instead of using UIA constructs


DaveTryon:
**Describe the bug**
In the spirit of #1464, there are a few places in the code where we pack information into the `Name` or `HelpText` properties instead of using other UIA constructs. This does an OK job of simulating default screen reader settings, but does not respect some non-default settings. If we can implement these using proper UIA constructs, the code will be cleaner and the user experience will be more customizable.

This bug just points to some string resources that demonstrate this behavior, then we can work back to the specific places in the code. There may be cases that are not included on this list. Some of these implementations may be fine @codeofdusk is probably our resident expert on which candidates are most worth fixing:

*** Synthesizing set information instead of using `SizeOfSet` and `PositionInSet`
- https://github.com/microsoft/accessibility-insights-windows/blob/main/src/AccessibilityInsights/MainWindowHelpers/ControlHelper.cs#L209-L226
- https://github.com/microsoft/accessibility-insights-windows/blob/main/src/AccessibilityInsights/Properties/Resources.resx#L124
- https://github.com/microsoft/accessibility-insights-windows/blob/main/src/AccessibilityInsights/Properties/Resources.resx#L130
- https://github.com/microsoft/accessibility-insights-windows/blob/main/src/AccessibilityInsights/Properties/Resources.resx#L145
- https://github.com/microsoft/accessibility-insights-windows/blob/main/src/AccessibilityInsights/Properties/Resources.resx#L169
- https://github.com/microsoft/accessibility-insights-windows/blob/main/src/AccessibilityInsights/Properties/Resources.resx#L202
- https://github.com/microsoft/accessibility-insights-windows/blob/main/src/AccessibilityInsights/Properties/Resources.resx#L447
- https://github.com/microsoft/accessibility-insights-windows/blob/main/src/AccessibilityInsights/Properties/Resources.resx#L586
- https://github.com/microsoft/accessibility-insights-windows/blob/main/src/AccessibilityInsights.SharedUx/Properties/Resources.resx#L901
- https://github.com/microsoft/accessibility-insights-windows/blob/main/src/AccessibilityInsights.SharedUx/Properties/Resources.resx#L910

*** Synthesizing expand/collapse state
- https://github.com/microsoft/accessibility-insights-windows/blob/main/src/AccessibilityInsights.SharedUx/Properties/Resources.resx#L385
[...]

PR #1529, opened 2023-01-12
fix: replace hardcoded checked/ unchecked names with toggle pattern

sfoslund:
#### Details

Fixes the checked/ unchecked instances in https://github.com/microsoft/accessibility-insights-windows/issues/1514

##### Motivation

Addresses issue https://github.com/microsoft/accessibility-insights-windows/issues/1514

##### Context

While checking that this did not disrupt any existing styling, I noticed/ filed https://github.com/microsoft/accessibility-insights-windows/issues/1528 to be addressed separately. 

#### Pull request checklist

- [x] Does this address an existing issue? If yes, https://github.com/microsoft/accessibility-insights-windows/issues/1514
- [n/a] Includes UI changes?

### DIFF
--- a/src/AccessibilityInsights.SharedUx/Controls/CustomControls/ToggleMenuItemAutomationPeer.cs
+++ b/src/AccessibilityInsights.SharedUx/Controls/CustomControls/ToggleMenuItemAutomationPeer.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using AccessibilityInsights.SharedUx.Utilities;
+using System.IO;
 using System.Windows.Automation;
 using System.Windows.Automation.Peers;
 using System.Windows.Automation.Provider;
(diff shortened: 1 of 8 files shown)


src/AccessibilityInsights.SharedUx/Controls/CustomControls/ToggleMenuItemAutomationPeer.cs
```diff
@@ -0,0 +1,46 @@
...
+namespace AccessibilityInsights.SharedUx.Controls.CustomControls
+{
+    internal class ToggleMenuItemAutomationPeer : MenuItemAutomationPeer, IToggleProvider
+    {
+        private readonly RadioButton _radioButton;
+
+        public ToggleMenuItemAutomationPeer(MenuItem item)
+            : base(item)
+        {
+            _radioButton = item.Header as RadioButton;
```
DaveTryon, line 19: Could malformed XAML cause _radioButton to be null? If so, would it make sense to throw an Exception here? ArgumentException wouldn't make sense, so it would need to be some other type

DaveTryon (approved):
Very nice! Approved with one small question

PR #1547, opened 2023-01-31
chore: Workaround for broken thirdpartynotice generator

DaveTryon:
#### Details

The API that we use to generate the `thirdpartynotices.html` has recently developed problems that may or may not be fixed in the near future. The recommendation from the team who owns the API is to commit the file into the repo and maintain it manually. This is far from ideal, but our options are pretty limited. In order to get the best balance, we do the following:

1. The manually curated version of `thirdpartynotices.html` is from the last build that successfully called the API
2. Local builds continue to use the placeholder file
3. The signed build copies the manually curated version over the placeholder version before calling the API
4. If the API succeeds, we get the newly generated file. If it fails, we get the manually curated file
5. Errors from the API are surfaced in the build logs

Dependabot changes will need to update the manually curated file
When the API is reliably fixed, we delete the manually curated file and revert the changes in `signedbuild.yml`

Validation build is at https://dev.azure.com/mseng/1ES/_build/results?buildId=19301482&view=results. The API failed, but the `thirdpartynotices.html` file in the artifacts and in the MSI are both correct.

##### Motivation

Unblock our build pipeline

#### Pull request checklist

- [n/a] Does this address an existing issue? If yes, Issue:
- [n/a] Includes UI changes?

### DIFF
--- a/build/signedbuild.yml
+++ b/build/signedbuild.yml
@@ -246,8 +246,17 @@ jobs:
   - task: ms.vss-governance-buildtask.governance-build-task-component-detection.ComponentGovernanceComponentDetection@0
     displayName: 'Component Detection'
 
+  - task: CopyFiles@2
+    displayName: 'TEMPORARY - Copy Prebuilt ThirdPartyNotices to: $(Build.ArtifactStagingDirectory)'
+    inputs:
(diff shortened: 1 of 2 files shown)


sfoslund (approved):
Too bad we don't have an ETA on the fix :(

PR #1558, opened 2023-02-11
feat(upgrader-security): Improve manifest handling

DaveTryon:
#### Details

This improves upgrade manifest _handling_ (content and structure are unmodified) in 2 ways:
1. Rather than needing to manually edit the value of `MinimumProdVersion` in [version.props](https://github.com/microsoft/accessibility-insights-windows/blob/main/src/props/prod-version.props) after each production release, we now populate the manifest's `production_minimum_version` field automatically based on the value of `IsMandatoryProdUpdate`, which will only need to changed if a mandatory upgrade released (which has happened **exactly once since 2019**):
    - If `IsMandatoryProdUpdate` is `false` (the usual state), then `production_minimum_version` is the same as the latest non-prerelease release from GitHub. This allows production clients to remain 1 version behind the latest release when this is released to production, consistent with our existing policy.
    - If `IsMandatoryProdUpdate` is `true`, then `production_minimum_version` is the same as `current_version`. This forces all production clients running earlier versions to upgrade when this is released to production.
2. We will now include the unsigned manifest in the `Manifest` build artifact. The objective is to replace human-provided data (specified when creating the Production release) with data from the `Manifest` build artifact. This reduces the risk of human error at release creation time.

##### Motivation

Streamline the release process

##### Context

This creates a new potential failure point, since the release build will fail if the GitHub API is down. Most local builds are debug (which does _not_ build the manifest), so the risk here is very low. The build pipeline might break, but if the GitHub API is down, it will probably break much earlier in the process. Altogether, the risk seems acceptably low. If we encounter too many build breaks with this new dependency, we can consider options to provide a workaround, possibly as a pipeline variable. YAGNI until proven otherwise.

#### Pull request checklist

- [ ] Includes UI changes?

### DIFF
--- a/src/Manifest/Manifest.csproj
+++ b/src/Manifest/Manifest.csproj
@@ -16,8 +16,12 @@
     <EmbeddedResource Include="$(MSBuildProjectDirectory)\..\bld\ReleaseInfo.json" />
   </ItemGroup>
 
+  <ItemGroup>
+    <PackageReference Include="Octokit" Version="5.0.0" />
+  </ItemGroup>
+
   <Target Name="PreBuild" BeforeTargets="PreBuildEvent">
-    <Exec Command="powershell -f $(SolutionDir)..\tools\scripts\create-json-for-manifest.ps1 -Verbose -MsiBasePath $(SolutionDir)MSI\bin\Release\msi -MinimumProductionVersion $(MinimumProdVersion) -OutputPath $(MSBuildProjectDirectory)\..\bld -OutputFile ReleaseInfo.json" />
+    <Exec Command="powershell -f $(SolutionDir)..\tools\scripts\create-json-for-manifest.ps1 -Verbose -MsiBasePath $(SolutionDir)MSI\bin\Release\msi -IsMandatoryProdUpdate $(IsMandatoryProdUpdate) -OutputPath $(MSBuildProjectDirectory)\..\bld -OutputFile ReleaseInfo.json" />
   </Target>
 
 </Project>
--- a/src/Manifest/Manifest.csproj
+++ b/src/Manifest/Manifest.csproj
@@ -17,6 +17,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <!-- NOTE: If you update this version, please also update the version in tools\scripts\create-json-for-manifest.ps1 -->
     <PackageReference Include="Octokit" Version="5.0.0" />
   </ItemGroup>
 
--- a/src/Manifest/Manifest.csproj
+++ b/src/Manifest/Manifest.csproj
@@ -13,7 +13,7 @@
   <Import Project="..\..\build\NetFrameworkRelease.targets" />
 
   <ItemGroup>
-    <EmbeddedResource Include="$(MSBuildProjectDirectory)\..\bld\ReleaseInfo.json" />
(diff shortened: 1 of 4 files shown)


src/Manifest/Manifest.csproj
```diff
@@ -13,11 +13,19 @@
   <Import Project="..\..\build\NetFrameworkRelease.targets" />
 
   <ItemGroup>
-    <EmbeddedResource Include="$(MSBuildProjectDirectory)\..\bld\ReleaseInfo.json" />
+    <EmbeddedResource Include="$(TargetDir)ReleaseInfo.json" LogicalName="AccessibilityInsights.Manifest.ReleaseInfo.json" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Octokit" Version="5.0.0" />
```
codeofdusk, line 20: Is this only retrieved when doing a release build/building the manifest?

tools/scripts/create-json-for-manifest.ps1
```diff
@@ -43,43 +43,81 @@ function CreatePaddedVersion([string]$version) {
...
+    Write-Verbose "    isMandatoryProdUpdate = $isMandatoryProdUpdate"
+
+    if ($isMandatoryProdUpdate -eq "true") {
+        Write-Verbose "    Using currentVersion for minimumProductionVersion"
+        $minimumProductionVersion = $currentVersion
+    } else {
+        Write-Verbose "    Fetching version from 'https://github.com/microsoft/accessibility-insights-windows/releases/tag/latest'"
+
+        $client = Get-Client
+        $release = $client.Repository.Release.GetLatest("Microsoft", "accessibility-insights-windows").Result
```
codeofdusk, line 98: Assuming the release object also contains error state and such? What happens if you attempt to access the `.result` member if an error occurs/the release can't be fetched?

codeofdusk (approved):

tools/scripts/create-json-for-manifest.ps1
DaveTryon: If an error occurs (can't reach the endpoint or somehow there's no "latest" build, then an Exception will get thrown and the script will fail

src/Manifest/Manifest.csproj
DaveTryon: Even though this project only _builds_ in release, the `nuget restore` step will always fetch this, even in debug. This is visible at [this step](https://dev.azure.com/accessibility-insights/Accessibility%20Insights/_build/results?buildId=44571&view=logs&j=54c8d87f-fb30-55c8-e653-392071771ae4&t=c36c8f7e-19fe-59c9-6025-4f2e0cb2c38b&l=653) of the PR build. I'll add a comment about keeping the version in sync with the PowerShell script, though

src/Manifest/Manifest.csproj
```diff
@@ -13,11 +13,20 @@
   <Import Project="..\..\build\NetFrameworkRelease.targets" />
 
   <ItemGroup>
-    <EmbeddedResource Include="$(MSBuildProjectDirectory)\..\bld\ReleaseInfo.json" />
+    <EmbeddedResource Include="$(TargetDir)ReleaseInfo.json" LogicalName="AccessibilityInsights.Manifest.ReleaseInfo.json" />
```
DaveTryon, line 16: The `LogicalName` is needed because the VS build changed the default resource name when the path changed. The client is hardcoded to use the original name, so we can't allow the name to change.

PR #1559, opened 2023-02-14
feat(upgrader-security): Deprecate unsigned manifests

DaveTryon:
#### Details

Now that all channels are using signed manifests, we can deprecate client support for unsigned manifests. This PR removes the code and tests that supported unsigned manifests. It also adds a test to be sure that we can always read the sign Canary manifest. Strictly speaking, this test is an integration test instead of a unit test, but the unit test class is so small that it seemed reasonable to put both types of tests into the same class.

##### Motivation

Complete upgrader security migration.

#### Pull request checklist

- [ ] Includes UI changes?


codeofdusk (approved):
Changes seem reasonable to me!

PR #1580, opened 2023-03-30
chore: Update analyzer suggestions

DaveTryon:
#### Details

A recent compiler update brought some updated analyzer suggestions. Some of these we need to block since they're not applicable to .NET Framework. Some others are incompatibilities between the analyzer and the code (`partial` classes confound the analyzer), and a few others are changes that are useful. We made similar changes in https://github.com/microsoft/axe-windows/pull/886 and are just catching up in this repo.

##### Motivation

Avoid noise/error from IDE suggestions

#### Pull request checklist

- [ ] Includes UI changes?


src/AccessibilityInsights.SharedUx/Controls/CustomControls/AutomatedChecksCustomListControl.xaml.cs
```diff
@@ -347,7 +347,9 @@ private void lviResults_PreviewKeyDown(object sender, KeyEventArgs e)
             }
         }
 
+#pragma warning disable CA1822 // Analyzer incorectly thinks this can be static
```
dbjorge, line 350: There are a lot of false positives for this rule; I think the downside of wasting time considering a false positive probably outweighs the benefit provided by this rule, so I'd be inclined to just disable the rule until we have a fix for the false positives. They look like the issue tracked by https://github.com/dotnet/roslyn-analyzers/issues/4733 - I'm not sure when exactly we'd expect to receive a fix for it, though.

dbjorge (approved):
LGTM! Left one comment about the CA1822 false positives, but I don't feel super strongly about it, happy leaving it to your preference whether to merge the PR as-is or accept the comment

DaveTryon:
> Left one comment about the CA1822 false positives, but I don't feel super strongly about it, happy leaving it to your preference whether to merge the PR as-is or accept the comment

I added a local `.editorconfig` file in the SharedUx project, which is the assembly that uses the pattern that triggers the CA1822 false positives. The net result is that we avoi the noise but still use the rule in most of the other projects.

PR #1699, opened 2023-09-11
chore: Enable chromium content scanning (advanced feature, not for general use)

DaveTryon:
#### Details

Provide an option to include Chromium HTML content in test results. This is intentionally _not_ exposed in the UI, as this is intended uniquely for browser development teams who can reasonably edit their config file by hand. To enable this functionality, do the following:

1. Run Accessibility Insights for Windows at least once, then exit the program. This will create the config file.
2. Using your preferred text editor, edit `%localappdata%\AccessibilityInsights\V1\Configurations\Configuration.json`
3. Add the following line in the json--not at the bottom of the file or the comma will be off
```
  "ShouldTestAllChromiumContent": true,
```
4. Save the JSON file
5. Restart Accessibility Insights for Windows

Chromium HTML content will now be included in rule checks.

Since we aren't supporting this in the UI, the setting gets locked in when the app initializes. Changing the setting requires shutting down the app, manually modifying the JSON file, then restarting the app.

##### Motivation

Feature requested by the Edge team

##### Comparing old and new

These results were on the same instance of Edge, using the modified version of AIWin. The only difference was the state of the config file when AIWin was launched.

Error | Count Before | Count After | Are extra results HTML content
--- | --- | --- | ---
An element must not have the same Name and LocalizedControlType as its parent | 0 | 6 | Yes
An element of the given ControlType must support the Text pattern | 0 | 1 | Yes
An on-screen element must not have a null BoundingRectangle property | 5 | 5 | n/a
Chromium components should be scanned with the web-based scanner | 2 | 2 | n/a
The Name must not include the same text as the LocalizedControlType | 1 | 1 | n/a
The Name property must not include the element's control type | 1 | 2 | Yes
The Name property of a focusable element must not be null | 0 | 1 | Yes
**TOTAL** | 9 | 18 | Yes

For anyone who may want to delve more deeply, both a11ytest files are contained in [Comparison.zip](https://github.com/microsoft/accessibility-insights-windows/files/12580328/Comparison.zip)

#### Pull request checklist

- [n/a] Includes UI changes?

### DIFF
--- a/src/AccessibilityInsights/MainWindowHelpers/Initialize.cs
+++ b/src/AccessibilityInsights/MainWindowHelpers/Initialize.cs
@@ -201,7 +201,8 @@ namespace AccessibilityInsights
             sa.IsFocusSelectOn = ConfigurationManager.GetDefaultInstance().AppConfig.SelectionByFocus;
             sa.IsMouseSelectOn = ConfigurationManager.GetDefaultInstance().AppConfig.SelectionByMouse;
             sa.TreeViewMode = ConfigurationManager.GetDefaultInstance().AppConfig.TreeViewMode;
-
+            // TODO: Enable this once Axe.Windows is updated
+            //SelectAction.ShouldTestAllChromiumContent = ConfigurationManager.GetDefaultInstance().AppConfig.ShouldTestAllChromiumContent;
         }
 
         /// <summary>
--- a/src/AccessibilityInsights/MainWindowHelpers/Initialize.cs
+++ b/src/AccessibilityInsights/MainWindowHelpers/Initialize.cs
@@ -201,8 +201,7 @@ namespace AccessibilityInsights
             sa.IsFocusSelectOn = ConfigurationManager.GetDefaultInstance().AppConfig.SelectionByFocus;
             sa.IsMouseSelectOn = ConfigurationManager.GetDefaultInstance().AppConfig.SelectionByMouse;
             sa.TreeViewMode = ConfigurationManager.GetDefaultInstance().AppConfig.TreeViewMode;
-            // TODO: Enable this once Axe.Windows is updated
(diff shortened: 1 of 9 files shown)


src/AccessibilityInsights/MainWindowHelpers/Initialize.cs
```diff
@@ -201,7 +201,8 @@ private static void InitSelectActionMode()
             sa.IsFocusSelectOn = ConfigurationManager.GetDefaultInstance().AppConfig.SelectionByFocus;
             sa.IsMouseSelectOn = ConfigurationManager.GetDefaultInstance().AppConfig.SelectionByMouse;
             sa.TreeViewMode = ConfigurationManager.GetDefaultInstance().AppConfig.TreeViewMode;
-
+            // TODO: Enable this once Axe.Windows is updated
+            //SelectAction.ShouldTestAllChromiumContent = ConfigurationManager.GetDefaultInstance().AppConfig.ShouldTestAllChromiumContent;
```
codeofdusk, line 205: Axe.Windows has indeed been updated

codeofdusk (changes requested):
* Consider changing the config key name to mention that the setting is for browser developers. Something like `showAllChromiumContentForBrowserDevelopment`
* We should publish some kind of log/telemetry events when this flag is on.
* Otherwise, looks pretty good to me!

src/AccessibilityInsights/MainWindowHelpers/Initialize.cs
DaveTryon: Oops! Had that change locally but hadn't pushed it.

DaveTryon:
@codeofdusk, I've fixed the forgotten push and added telemetry

codeofdusk (approved):

PR #1704, opened 2023-09-21
Chore: Migrate from ADAL to MSAL

DaveTryon:
#### Details

ADAL (Active Directory Authentication Library) is being replaced by MSAL (Microsoft Authentication Library) across all products. Our auth is a transitive dependency of the Visual Studio client assemblies, so our upgrade path is to to upgrade the Visual Studio client packages. These packages are still in preview with no publicly announced ETA for the full release, but there are reasons to make this change sooner, rather than later. In conversations with the teams that own the Visual Studio packages, no significant changes are expected between now and the fully released issues, so we're making the change now.

Change was tested by building an MSI, installing it on a machine, and ensuring that ADO bug filing works as expected.

##### Motivation

Complete update requirement

#### Pull request checklist

- [n/a] Includes UI changes?

### DIFF
--- a/src/AccessibilityInsights.Extensions.AzureDevOps/ForcedDependencies.cs
+++ b/src/AccessibilityInsights.Extensions.AzureDevOps/ForcedDependencies.cs
@@ -33,7 +33,6 @@ namespace AccessibilityInsights.Extensions.AzureDevOps
         private static void ManuallyEnforcedDependencies()
         {
             // Types that we can't detect via unit tests. Namespace generally matches the package.
-            ForceDependency(typeof(Microsoft.IdentityModel.Clients.ActiveDirectory.AdalClaimChallengeException));
             ForceDependency(typeof(Microsoft.IdentityModel.JsonWebTokens.JsonClaimValueTypes));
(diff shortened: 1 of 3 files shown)


src/AccessibilityInsights.Extensions.AzureDevOps/ForcedDependencies.cs
```diff
@@ -33,7 +33,6 @@ private static void TestEnforcedDependencies()
         private static void ManuallyEnforcedDependencies()
         {
             // Types that we can't detect via unit tests. Namespace generally matches the package.
-            ForceDependency(typeof(Microsoft.IdentityModel.Clients.ActiveDirectory.AdalClaimChallengeException));
```
codeofdusk, line 36: Is there a replacement type that needs to be added here?

codeofdusk (approved):

src/AccessibilityInsights.Extensions.AzureDevOps/ForcedDependencies.cs
DaveTryon: Good question. I confirmed that all of the files in the ADO extension make it into the MSI file, so I think we're ok

Issue #1761, opened 2024-01-22
[BUG] *.yml and *.targets files should contain copyright headers


DaveTryon:
Please check whether the bug has [already been filed](https://github.com/Microsoft/axe-windows/issues).

## Describe the bug

Our `*.yml` and `*.targets` files currently lack copyright headers. Suggested change:
1. Add `*.yml` and `*.targets` to the script to check for headers
2. Run the script to update the files
3. Include the updated files in the PR

**To Reproduce**

Open any of the files under https://github.com/microsoft/accessibility-insights-windows/tree/main/build

## Expected behavior

These files should include copyright headers

## Actual behavior

No copyright headers

## Context
It probably makes sense to keep the script in sync with changes from https://github.com/microsoft/axe-windows/issues/991

PR #1767, opened 2024-01-29
fix: Add license headers to .yml and .targets files

DaveTryon:
#### Details

While updating to 1ES pipeline templates, we realized that several of our files were missing copyright headers. This adds them and updates the tooling to ensure that we don't miss them in these files types in the future. Copyright headers were all added by running the script.

The policy service file can't have comments without breaking the bot, so an exception is made for any files under `.github\policies`.

##### Motivation

Address #1761 

#### Pull request checklist

- [x] Does this address an existing issue? If yes, Issue #1761
- [n/a] Includes UI changes?

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,3 +1,5 @@
+﻿# Copyright (c) Microsoft. All rights reserved.
+# Licensed under the MIT license. See LICENSE file in the project root for full license information.
 version: 2
 updates:
 - package-ecosystem: nuget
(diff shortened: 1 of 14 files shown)


.github/dependabot.yml
```diff
@@ -1,3 +1,5 @@
+﻿# Copyright (c) Microsoft. All rights reserved.
```
DaveTryon, line 1: There's a potential that this will break the dependabot agent, but we don't have a way to check it without merging it. If it breaks, we'll back out the change and expand the exceptions.

codeofdusk (approved):

